PR #36, opened 2015-01-05
Going with the new `Iterator`s

akiss77:
Fixes erros reported by latest rustc because of the change in Iterator traits.


alexcrichton:
Thanks!

PR #39, opened 2015-01-10
Start making the library build with alpha compiler

zmbush:
Mostly fixing issues related to https://github.com/rust-lang/rust/pull/20507/

There is some memory corruption somewhere, but I'm not sure where to start with debugging it.

### DIFF
--- a/examples/add.rs
+++ b/examples/add.rs
@@ -13,6 +13,7 @@
  */
 
 #![deny(warnings)]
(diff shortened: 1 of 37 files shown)


alexcrichton:
Thanks! I started a bunch of this but then lost the computer I was working on and was hoping to get it back soon. Ah well, I'll merge in the meantime and see if I can't get it compiling again.

PR #45, opened 2015-02-03
Derive Debug for Delta.

Hoverbear:
Is there any harm to deriving Debug for Delta? It might be useful for development.


alexcrichton:
Sure!

PR #47, opened 2015-02-08
Mention pkg-config Dependency

killercup:
At least on OS X you need to explicitly install pkg-config (e.g. using brew).


alexcrichton:
Thanks!

PR #51, opened 2015-02-26
fixup for nightly

jmesmon:


jmesmon:
Depends on ssh2-rs nightly fixups

PR #55, opened 2015-04-08
Replace old static link flag syntax in build.rs

hugoduncan:
Replace the `-l git2:static` flag with `-l static=git2`.


alexcrichton:
Wow I thought I had removed all this ages ago!

PR #56, opened 2015-04-12
Output stdout and stderr if external command fails.

o01eg:


alexcrichton:
Hm for debugging I personally prefer to not capture this output and only print it on error as it's often quite nice to debug the output by checking script's output. Thanks for the PR though!

PR #59, opened 2015-04-28
Add function revparse_ext to Repository impl

nott:
This PR adds Repository::revparse_ext. This function is a thin proxy for [https://libgit2.github.com/libgit2/#HEAD/group/repository/git_revparse_ext](git_revparse_ext).

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -144,6 +144,32 @@ impl Repository {
         }
     }
 
+    /// Find a single object and intermediate reference by a revision string.
+    ///
+    /// See `man gitrevisions`, or
+    /// http://git-scm.com/docs/git-rev-parse.html#_specifying_revisions for
+    /// information on the syntax accepted.
+    ///
+    /// In some cases (`@{<-n>}` or `<branchname>@{upstream}`), the expression may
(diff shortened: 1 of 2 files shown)


src/repo.rs
```diff
@@ -144,6 +144,32 @@ impl Repository {
...
+    pub fn revparse_ext(&self, spec: &str) -> Result<(Object, Option<Reference>), Error> {
+        let spec = try!(CString::new(spec));
+        let mut git_obj = 0 as *mut raw::git_object;
+        let mut git_ref = 0 as *mut raw::git_reference;
+        unsafe {
+            try_call!(raw::git_revparse_ext(&mut git_obj, &mut git_ref,
+                                            self.raw, spec));
+            assert!(!git_obj.is_null());
+            let object = Binding::from_raw(git_obj);
+            let reference = match git_ref.is_null() {
```
alexcrichton: I think this could actually leverage `Binding::from_raw_opt` to cover the null/not null case as well

alexcrichton:
Awesome, thanks!

src/repo.rs
nott: Thanks for pointing that out.

PR #61, opened 2015-05-03
[RDY] Bind the blame API

askobara:
Looks like I'm finished.
- [x] bind `git_blame_hunk`
- [x] bind `git_blame_get_hunk_byindex`
- [x] bind `git_blame_get_hunk_byline`
- [x] think of how to get raw_data for hunks
- [x] write example of use

`git_blame_buffer` is too complicated for my understanding :)

### DIFF
--- a/examples/blame.rs
+++ b/examples/blame.rs
@@ -1,0 +1,105 @@
+/*
+ * libgit2 "blame" example - shows how to use the blame API
+ *
+ * Written by the libgit2 contributors
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright
+ * and related and neighboring rights to this software to the public domain
+ * worldwide. This software is distributed without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
+
+extern crate git2;
+extern crate docopt;
+extern crate rustc_serialize;
+
+use docopt::Docopt;
+use git2::{Repository, BlameOptions};
+use std::path::Path;
+use std::io::{BufReader, BufRead};
+
+#[derive(RustcDecodable)] #[allow(non_snake_case)]
+struct Args {
+    arg_path: String,
+    arg_spec: Option<String>,
+    flag_M: bool,
+    flag_C: bool,
+    flag_F: bool,
+}
+
+fn run(args: &Args) -> Result<(), git2::Error> {
+    let repo = try!(Repository::open("."));
+    let path = Path::new(&args.arg_path[..]);
+
+    // Prepare our blame options
+    let mut opts = BlameOptions::new();
+    opts.track_copies_same_commit_moves(args.flag_M)
+        .track_copies_same_commit_copies(args.flag_C)
+        .first_parent(args.flag_F);
+
+    let mut commit_id = "HEAD".to_string();
+
+    // Parse spec
+    if let Some(spec) = args.arg_spec.as_ref() {
+
+        let revspec = try!(repo.revparse(spec));
+
+        let (oldest, newest) = if revspec.mode().contains(git2::REVPARSE_SINGLE) {
+            (None, revspec.from())
+        } else if revspec.mode().contains(git2::REVPARSE_RANGE) {
+            (revspec.from(), revspec.to())
+        } else {
+            (None, None)
+        };
+
+        if let Some(commit) = oldest {
+            opts.oldest_commit(commit.id());
+        }
+
+        if let Some(commit) = newest {
+            opts.newest_commit(commit.id());
+            if !commit.id().is_zero() {
+                commit_id = format!("{}", commit.id())
+            }
+        }
+
+    }
+
+    let spec = format!("{}:{}", commit_id, path.display());
+    let blame = try!(repo.blame_file(path, Some(&mut opts)));
+    let object = try!(repo.revparse_single(&spec[..]));
+    let blob = try!(repo.find_blob(object.id()));
+    let reader = BufReader::new(blob.content());
+
+    for (i, line) in reader.lines().enumerate() {
+        if let (Ok(line), Some(hunk)) = (line, blame.get_line(i+1)) {
+            let sig = hunk.final_signature();
+            println!("{} {} <{}> {}", hunk.final_commit_id(),
+                     String::from_utf8_lossy(sig.name_bytes()),
+                     String::from_utf8_lossy(sig.email_bytes()), line);
+        }
+    }
+
+    Ok(())
+}
+
+fn main() {
+    const USAGE: &'static str = "
+usage: blame [options] [<spec>] <path>
+
+Options:
+    -M                  find line moves within and across files
+    -C                  find line copies within and across files
+    -F                  follow only the first parent commits
+";
+
+    let args = Docopt::new(USAGE).and_then(|d| d.decode())
+                                 .unwrap_or_else(|e| e.exit());
+    match run(&args) {
+        Ok(()) => {}
+        Err(e) => println!("error: {}", e),
+    }
+}
--- a/src/blame.rs
+++ b/src/blame.rs
@@ -1,0 +1,301 @@
+use std::marker;
+use {raw, Repository, Oid, signature, Signature};
+use util::{self, Binding};
+use std::path::Path;
+use std::ops::Range;
+use std::mem;
+
+/// Opaque structure to hold blame results.
+pub struct Blame<'repo> {
+    raw: *mut raw::git_blame,
+    _marker: marker::PhantomData<&'repo Repository>,
+}
+
+/// Structure that represents a blame hunk.
+pub struct BlameHunk<'blame> {
+    raw: *mut raw::git_blame_hunk,
+    _marker: marker::PhantomData<&'blame raw::git_blame>,
+}
+
+/// Blame options
+pub struct BlameOptions {
+    raw: raw::git_blame_options,
+}
+
+/// An iterator over the hunks in a blame.
+pub struct BlameIter<'blame> {
+    range: Range<usize>,
+    blame: &'blame Blame<'blame>,
+}
+
+impl<'repo> Blame<'repo> {
+
+    /// Gets the number of hunks that exist in the blame structure.
+    pub fn len(&self) -> usize {
+        unsafe { raw::git_blame_get_hunk_count(self.raw) as usize }
+    }
+
+    /// Gets the blame hunk at the given index.
+    pub fn get_index(&self, index: usize) -> Option<BlameHunk> {
+        unsafe {
+            let ptr = raw::git_blame_get_hunk_byindex(self.raw(), index as u32);
+            if ptr.is_null() {
+                None
+            } else {
+                Some(BlameHunk::from_raw_const(ptr))
+            }
+        }
+    }
+
+    /// Gets the hunk that relates to the given line number in the newest
+    /// commit.
+    pub fn get_line(&self, lineno: usize) -> Option<BlameHunk> {
+        unsafe {
+            let ptr = raw::git_blame_get_hunk_byline(self.raw(), lineno as u32);
+            if ptr.is_null() {
+                None
+            } else {
+                Some(BlameHunk::from_raw_const(ptr))
+            }
+        }
+    }
+
+    /// Returns an iterator over the hunks in this blame.
+    pub fn iter(&self) -> BlameIter {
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -81,6 +81,7 @@ use std::fmt;
 use std::str;
 use std::sync::{Once, ONCE_INIT};
 
+pub use blame::Blame;
```
alexcrichton: This may also want to reexport BlameHunk and BlameIter

src/repo.rs
```diff
@@ -1088,6 +1088,21 @@ impl Repository {
...
 
+    /// Get the blame for a single file.
+    pub fn blame_file<'a>(&self, path: &Path) -> Result<Blame<'a>, Error> {
+        let path = try!(path.into_c_string());
+        let mut raw = 0 as *mut raw::git_blame;
+
+        unsafe {
+            let mut raw_opts = mem::zeroed();
+            try_call!(raw::git_blame_init_options(&mut raw_opts,
+                                            raw::GIT_BLAME_OPTIONS_VERSION));
```
alexcrichton: I've been trying to expose these sorts of options as builders whenever possible, so could this have an argument of `Option<&mut BlameOptions>` or something like that? There should be some examples when dealing with diffs.

It's ok to not bind the full option API just yet, that can always come later!

src/repo.rs
```diff
@@ -1088,6 +1088,21 @@ impl Repository {
         }
     }
 
+    /// Get the blame for a single file.
+    pub fn blame_file<'a>(&self, path: &Path) -> Result<Blame<'a>, Error> {
```
alexcrichton: I think this lifetime parameter means that the returned `Blame` is connected to any lifetime? You should be fine just omitting this entirely (the lifetime parameter)

src/blame.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl<'repo> Blame<'repo> {
+
+    /// Gets the number of hunks that exist in the blame structure.
+    pub fn len(&self) -> usize {
+        unsafe { raw::git_blame_get_hunk_count(self.raw) as usize }
+    }
+
+    /// Gets the blame hunk at the given index.
+    pub fn get_index(&self, index: u32) -> Option<BlameHunk> {
```
alexcrichton: Although the underlying API takes `u32`, this may be best to take a `usize` to be consistent with `len`

src/blame.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Returns an iterator over the hunks in this blame.
+    pub fn iter(&self) -> BlameIter {
+        BlameIter { range: 0..self.len(), blame: self }
+    }
+
+}
+
+impl<'blame> BlameHunk<'blame> {
```
alexcrichton: Could this block also expose fields like `lines_in_hunk` and friends?

src/blame.rs
```diff
@@ -0,0 +1,167 @@
...
+    unsafe fn from_raw_const(raw: *const raw::git_blame_hunk)
+                                 -> BlameHunk<'blame> {
+        BlameHunk {
+            raw: raw as *mut raw::git_blame_hunk,
+            _marker: marker::PhantomData,
+        }
+    }
+
+    /// Returns OID of the commit where this line was last changed.
+    pub fn id(&self) -> Oid {
```
alexcrichton: Perhaps this should be called `final_commit_id` to leave from for `orig_commit_id` to be exposed?

alexcrichton:
Nice!

src/repo.rs
askobara: yep, good idea with `Option`, I will do it!

src/repo.rs
```diff
@@ -1088,6 +1088,21 @@ impl Repository {
         }
     }
 
+    /// Get the blame for a single file.
+    pub fn blame_file(&self, path: &Path, opts: Option<&mut BlameOptions>)
+                                                -> Result<Blame, Error> {
```
alexcrichton: My current style here is to align the `->` with the first character after the opening `(` of the arguments.

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+        self
+    }
+
+    /// Get the newest commit OID.
+    pub fn newest_commit(&self) -> Oid {
+        unsafe { Binding::from_raw(&self.raw.newest_commit as *const _) }
+    }
+
+    /// Get the oldest commit OID.
+    pub fn oldest_commit(&self) -> Oid {
```
alexcrichton: Currently the style for builders provided in git2-rs is to only have setters, and the stylistic guideline is for all setter methods to have the form `foo` instead of `set_foo`, so perhaps this could only have the `set_newest_commit` and `set_oldest_commit` for now?

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+        self.flag(raw::GIT_BLAME_TRACK_COPIES_ANY_COMMIT_COPIES, opt)
+    }
+
+    /// First parent
+    pub fn first_parent(&mut self, opt: bool) -> &mut BlameOptions {
+        self.flag(raw::GIT_BLAME_FIRST_PARENT, opt)
+    }
+
+    /// Raw pointer
+    pub unsafe fn raw(&mut self) -> *mut raw::git_blame_options {
```
alexcrichton: Could this be omitted from the public API for now? I've been trying to cut back on the raw pointers presented in hopes of a future more-organized system.

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub fn track_copies_same_commit_copies(&mut self, opt: bool) -> &mut BlameOptions {
+        self.flag(raw::GIT_BLAME_TRACK_COPIES_SAME_COMMIT_COPIES, opt)
+    }
+
+    /// Any commit copies
+    pub fn track_copies_any_commit_copies(&mut self, opt: bool) -> &mut BlameOptions {
+        self.flag(raw::GIT_BLAME_TRACK_COPIES_ANY_COMMIT_COPIES, opt)
+    }
+
+    /// First parent
```
alexcrichton: Could this include the docs that libgit2 provides? For example:

> Restrict the search of commits to those reachable following only the first parents.

alexcrichton: (same for the above methods as well)

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// Tests whether this hunk has been tracked to a boundary commit.
+    pub fn is_boundary(&self) -> bool {
+        unsafe {
+            (*self.raw).boundary == 1
+        }
+    }
+
+    /// Returns number of lines in this hunk.
+    pub fn len(&self) -> usize {
```
alexcrichton: This may be best named `lines_in_hunk` to retain clarity (otherwise the "length of a hunk" may be a little ambiguous)

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /// Returns path to the file where this hunk originated.
+    pub fn path(&self) -> &Path {
+        unsafe {
+            util::bytes2path(::opt_bytes(self, (*self.raw).orig_path).unwrap())
+        }
+    }
+
+    /// Tests whether this hunk has been tracked to a boundary commit.
```
alexcrichton: Could this mention the extra docs for the field that talk about the `oldest_commit` option as well?

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub fn orig_start_line(&self) -> usize {
+        unsafe { (*self.raw).orig_start_line_number as usize }
+    }
+
+    /// Returns path to the file where this hunk originated.
+    pub fn path(&self) -> &Path {
+        unsafe {
+            util::bytes2path(::opt_bytes(self, (*self.raw).orig_path).unwrap())
+        }
+    }
```
alexcrichton: Could this return an `Option<&Path>` for now and avoid the `.unwrap()`? Also, could this document that `None` could be returned for non-unicode paths on Widnows?

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+    ///
+    /// Note that the start line is counting from 1.
+    pub fn final_start_line(&self) -> usize {
+        unsafe { (*self.raw).final_start_line_number as usize }
+    }
+
+    /// Returns the OID of the commit where this hunk was found.
+    ///
+    /// This will usually be the same as `final_commit_id`,
+    /// except when `GIT_BLAME_TRACK_COPIES_ANY_COMMIT_COPIES`
```
alexcrichton: Could this mention of the option be altered to match the  method on the builder below?

src/blame.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+impl<'blame> Iterator for BlameIter<'blame> {
+    type Item = BlameHunk<'blame>;
+    fn next(&mut self) -> Option<BlameHunk<'blame>> {
+        self.range.next().and_then(|i| self.blame.get_index(i))
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) { self.range.size_hint() }
+}
```
alexcrichton: Could you implement `DoubleEndedIterator` and `ExactSizeIterator` here as well?

examples/blame.rs
```diff
@@ -0,0 +1,88 @@
+extern crate git2;
```
alexcrichton: Was this example drawn from libgit2's `examples/blame.c`? If so, could you be sure to include the license block at the top?

alexcrichton:
Awesome work @askobara! Also feel free to comment on this PR whenever you update, unfortunately github doesn't send out notifications otherwise :(

src/blame.rs
askobara: I'm not understand what you mean by the "builder" :(

src/blame.rs
askobara: hmm... but how to get access to the raw data from raw::git_blame_* functions?

src/blame.rs
askobara: raw::git_blame_file can modify git_blame_options struct that passed into it and this scenario used in example to determine revision of file to load it from blob.
So, I think it should have the getters and setters.
Of course I can rename setters, but how to name getters? :)

src/blame.rs
alexcrichton: Right now there's nothing in libgit2 that generates `git_blame_options`, this is the only creator of it, so it's in theory possible to track the values in Rust itself. This is something that's generally true of all builders in this library right now (you can't read the options), and we can add some accessors later on if necesary.

src/blame.rs
alexcrichton: The documentation of `git_blame_file` doesn't seem to mention this behavior, so I'm curious where you're seeing it?

src/blame.rs
alexcrichton: Ah by builder I mean `BlameOptions`

src/blame.rs
askobara: First I notice that last parameter of the `git_blame_file` is not `const` and than I see [example/blame.c](https://github.com/libgit2/libgit2/blob/f2e09b8aac1e107a3a02f2831b46d5b4367cd5ca/examples/blame.c#L86)

src/blame.rs
askobara: Oh, if I right understand you mean that this method should return structure by value like `RepositoryInitOptions::raw`?

alexcrichton: Gah that probably shouldn't be a public function! I'd like to perhaps expose the `Binding` trait one day to lower the representation of each type to the underlying libgit2 type, but I don't think that's quite necessary just yet.

src/blame.rs
alexcrichton: Looking at the [current source](https://github.com/libgit2/libgit2/blob/25f355cbedb2320ca5c4db5e02d1ab9ca7ac9cab/src/blame.c#L354-L382) while `const` is not located, the [only usage](https://github.com/libgit2/libgit2/blob/25f355cbedb2320ca5c4db5e02d1ab9ca7ac9cab/src/blame.c#L365) does indeed [have the `const` qualifier](https://github.com/libgit2/libgit2/blob/25f355cbedb2320ca5c4db5e02d1ab9ca7ac9cab/src/blame.c#L184).

I also think that the example in the repo is probably just using that variable as a proxy for holding other data (e.g. looking at the value set previously) rather than relying on `git_file_blame` to actually modify it.

askobara:
@alexcrichton looks like I fixed all errors in code, but I'm not sure that [is](https://github.com/alexcrichton/git2-rs/pull/61/files#diff-51851bcdff00b4967b4fb82de658c381R242) correct type casting

alexcrichton:
Looks all good to me, thanks @askobara!

PR #62, opened 2015-05-04
Display for Signature

askobara:
Little helpful thing :)

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -2,6 +2,7 @@ use std::ffi::CString;
 use std::marker;
 use std::mem;
 use std::str;
+use std::fmt;
 use libc;
 
 use {raw, Error, Time};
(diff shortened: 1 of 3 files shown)


src/signature.rs
```diff
@@ -139,6 +140,18 @@ impl<'a> Drop for Signature<'a> {
     }
 }
 
+impl<'a> fmt::Display for Signature<'a> {
+
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let _ = write!(f, "{} <{}>",
```
alexcrichton: This should proxy along the return value of `write!` instead of explicitly ignoring it, but other than that this looks great to me, thanks!

alexcrichton:
Thanks!

PR #65, opened 2015-05-18
adding bitrig dependency

dhuseby:


dhuseby:
@alexcrichton should I have bumped the patch number up one on these crates that I added bitrig dependency clauses to?  what's the right way to get cargo to pull the crates with the bitrig changes in them?

It seems to me I should have bumped the patch number up one and then changed the dependencies versions in the toml/lock file in the Cargo repo.  Is that right?

alexcrichton:
Ah it's ok, I can do the bump, I may have just forgotten to do it on a few of them...

dhuseby:
@alexcrichton I just graphed out the dependencies and it looks like you need to bump:
* libssh2-sys-0.1.22 -> libssh2-sys-0.1.23
* curl-sys-0.1.21 -> curl-sys-0.1.22
* libgit2-sys-0.2.13 -> libgit2-sys-0.2.14
* curl-0.2.9 -> curl-0.2.10

PR #66, opened 2015-05-20
Implement peeling for references

carlosmn:
When a command is interested in e.g. a tree-ish, calling the reference's
peel function is a lot more convenient than going through the objects by
hand.


alexcrichton:
Thanks!

PR #74, opened 2015-06-15
Add bindings for git_merge()

tksfz:


tksfz:
Hi I think this is at least ready to get some feedback.  I'm only trying to build enough to get git_merge() implemented which requires AnnotatedCommit and MergeOptions.  There are several other merge-related functions, as well as cherry-pick and rebase functions.  I'm hoping they'll get easier to implement once the basics are laid down.

Feel free to give any kind of feedback or let me know if this is not the right direction.

src/lib.rs
```diff
@@ -230,6 +231,19 @@ pub enum ConfigLevel {
     Highest,
 }
 
+/// An enumeration for favoring sides to resolve merge conflicts
+#[derive(PartialEq, Eq, Debug, Copy, Clone)]
+pub enum FileFavor {
+    /// Record conflicts in the index without favoring sides
```
alexcrichton: Could these docs copy over the longer form docs in the corresponding header file?

src/merge.rs
```diff
@@ -0,0 +1,138 @@
...
+use libc::c_uint;
+
+use {raw, Object, FileFavor};
+use util::Binding;
+use call::Convert;
+
+/// annotated commits, the input to merge and rebase
+pub struct AnnotatedCommit<'repo> {
+    raw: *mut raw::git_annotated_commit,
+    _marker: marker::PhantomData<Object<'repo>>,
```
alexcrichton: Looking at the definition in C, it looks like this actually contains a `Commit`, so could this phantom data be `Commit<'repo>` instead?

src/merge.rs
```diff
@@ -0,0 +1,138 @@
+use std::marker;
+use std::mem;
+use libc::c_uint;
+
+use {raw, Object, FileFavor};
+use util::Binding;
+use call::Convert;
+
+/// annotated commits, the input to merge and rebase
```
alexcrichton: Could this be expanded a bit (and use the same sentence-like style of other docs)?

src/merge.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Take extra time to find minimal diff
+    pub fn minimal(&mut self, find: bool) -> &mut MergeOptions {
+        self.flag(raw::GIT_MERGE_FILE_DIFF_MINIMAL, find)
+    }
+
+    /// Acquire a pointer to the underlying raw options.
+    ///
+    /// This function is unsafe as the pointer is only valid so long as this
+    /// structure is not moved, modified, or used elsewhere.
+    // modeled after DiffOptions.raw()
```
alexcrichton: It's ok to drop these kinds of comments

src/merge.rs
alexcrichton: Also, could some accessor functions of `AnnotatedCommit` also be exposed? It looks like there's a few simple-ish ones in the header files

src/repo.rs
```diff
@@ -785,6 +786,17 @@ impl Repository {
         }
     }
 
+    /// Creates a git_annotated_commit from the given reference.
+    pub fn reference_to_annotated_commit(&self, reference: &Reference) -> Result<AnnotatedCommit, Error> {
```
alexcrichton: Could you make sure this wraps at 80 chars?

src/repo.rs
```diff
@@ -989,6 +1001,36 @@ impl Repository {
         Ok(())
     }
 
+    /// Merges the given commit(s) into HEAD, writing the results into the working directory.
+    /// Any changes are staged for commit and any conflicts are written to the index. Callers
+    /// should inspect the repository's index after this completes, resolve any conflicts and
+    /// prepare a commit.
```
alexcrichton: Could this included the rest of the docs as well? (e.g. just copied from libgit2), also, can this be wrapped to 80 chars?

src/repo.rs
```diff
@@ -989,6 +1001,36 @@ impl Repository {
...
+    pub fn merge(&self,
+                 annotated_commits: &[&AnnotatedCommit],
+                 merge_opts: &MergeOptions,
```
alexcrichton: I think that this can technically be `Option<&MergeOptions>`

src/repo.rs
```diff
@@ -989,6 +1001,36 @@ impl Repository {
...
+            try_call!(raw::git_checkout_init_options(&mut raw_checkout_opts,
+                                raw::GIT_CHECKOUT_OPTIONS_VERSION));
+            match checkout_opts {
+                Some(c) => c.configure(&mut raw_checkout_opts),
+                None => {}
+            }
+
+            let commit_ptrs: Vec<*const raw::git_annotated_commit> =  annotated_commits.iter().map(|c| {
+                c.raw() as *const raw::git_annotated_commit
+            }).collect();
```
alexcrichton: Stylistically I'd recommend putting the type annotation at the end and doing something like `.collect::<Vec<_>>()`

alexcrichton:
Thanks @tksfz!

src/lib.rs
tksfz: Yes I'll change it to copy the docs from libgit2.  I was under the impression that other docs in this code base seemed to be paraphrasing the libgit2 docs but I like copying them.

src/merge.rs
tksfz: re phantom data: interesting, i'll take a look.  I vaguely recall that you can create annotated commits from things other than commits but I may well be mistaken.

src/merge.rs
tksfz: Ok, I noticed other raw() methods were doing the same thing.  I'll remove all of it except for the first line.

src/merge.rs
tksfz: done

src/merge.rs
tksfz: done

src/repo.rs
tksfz: done

tksfz:
Re accessor functions for annotated_commit - are you referring to the functions that create a new annotated_commit (get_annotated_commit_from_{fetchhead, revspec}, etc.) or the get_annotated_commit_id() function?

tksfz:
Re `Option<&MergeOptions>` - can you clarify?  How would I generate a raw_merge_options for that, in a clean way?  I think I'd need to call `git_merge_init_options` at least.  Currently, I'm supposing that callers will call `MergeOptions::new()` to create it.

src/merge.rs
tksfz: I changed it to `Commit<`repo>`

src/repo.rs
tksfz: Done - it's kind of long now though.

alexcrichton:
> Re accessor functions for annotated_commit - are you referring to the functions that create a new annotated_commit (get_annotated_commit_from_{fetchhead, revspec}, etc.) or the get_annotated_commit_id() function?

Ah I mostly meant the accessors, but I do think that some extra constructors would be nice as well (just so the methods of creating them are all bound).

>  How would I generate a raw_merge_options for that, in a clean way?

In the `None` case I believe you can just pass down a null pointer.

tksfz:
Re None for MergeOptions - ah ok great passing null seems to work great in that case.

Re accessors:  Ok, I added the only accessor which is for id().

I haven't added the constructors yet but I'll likely have a branch in the future to add that.

Btw, I already started another follow-on branch that adds Repository::cleanup_state(), which is kind of necessary for using merge, as well as a few other functions.  For the sake of keeping this branch small I figured I'd keep this branch as-is, and create another pull for that branch.

alexcrichton:
Thanks!

PR #76, opened 2015-07-06
Add a few more merge-related bindings

tksfz:
Adding repo.merge_commits() here, repo.cleanup_state(), as well as ref.set_target()

### DIFF
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -179,6 +179,19 @@ impl<'repo> Reference<'repo> {
         }
     }
 
+    /// Conditionally create a new reference with the same name as the given
+    /// reference but a different OID target. The reference must be a direct
+    /// reference, otherwise this will fail.
+    pub fn set_target(&mut self, id: Oid, msg: &str) -> Result<Reference<'repo>, Error> {
+        let mut raw = 0 as *mut raw::git_reference;
+        let msg = try!(CString::new(msg));
+        unsafe {
+            try_call!(raw::git_reference_set_target(&mut raw, self.raw,
+                                                    id.raw(), msg));
+            Ok(Binding::from_raw(raw))
+        }
+    }
+
 }
 
 impl<'repo> PartialOrd for Reference<'repo> {
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -182,6 +182,9 @@ impl<'repo> Reference<'repo> {
     /// Conditionally create a new reference with the same name as the given
     /// reference but a different OID target. The reference must be a direct
     /// reference, otherwise this will fail.
+    ///
+    /// The new reference will be written to disk, overwriting the given
+    /// reference.
     pub fn set_target(&mut self, id: Oid, msg: &str) -> Result<Reference<'repo>, Error> {
         let mut raw = 0 as *mut raw::git_reference;
         let msg = try!(CString::new(msg));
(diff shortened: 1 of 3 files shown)


src/reference.rs
```diff
@@ -179,6 +179,19 @@ impl<'repo> Reference<'repo> {
         }
     }
 
+    /// Conditionally create a new reference with the same name as the given
+    /// reference but a different OID target. The reference must be a direct
+    /// reference, otherwise this will fail.
```
alexcrichton: Can this also include the clause "The new reference will be written to disk, overwriting the given reference."

src/reference.rs
```diff
@@ -179,6 +179,19 @@ impl<'repo> Reference<'repo> {
...
+    pub fn set_target(&mut self, id: Oid, msg: &str) -> Result<Reference<'repo>, Error> {
```
alexcrichton: Could you try passing in a `None` here for `msg` and see what happens? It wasn't clear to me whether it was required or not from reading the code.

alexcrichton: er, reading the libgit2 code that is

src/repo.rs
```diff
@@ -1041,6 +1041,32 @@ impl Repository {
         Ok(())
     }
 
+    /// Merge two commits, producing a git_index that reflects the result of
+    /// the merge. The index may be written as-is to the working directory or
+    /// checked out. If the index is to be converted to a tree, the caller
+    /// should resolve any conflicts that arose as part of the merge.
```
alexcrichton: Can this also include the clauses that start with "The merge performed uses the first common ancestor ..."

src/repo.rs
```diff
@@ -1041,6 +1041,32 @@ impl Repository {
         Ok(())
     }
 
+    /// Merge two commits, producing a git_index that reflects the result of
```
alexcrichton: s/a git_index/an Index/

alexcrichton:
Overall looks great to me, thanks @tksfz!

src/repo.rs
tksfz: done

src/repo.rs
tksfz: done

src/reference.rs
tksfz: done

src/reference.rs
tksfz: So when I pass in `0 as *const libc::c_char`  for msg (to the raw call), everything seems to work (in at least one particular case).  The reflog records an empty message.

Do you  want  me to make `msg` an `Option`?  I could do that but then we might want  to  modify the  existing `reference.rename()` function, which also takes a  msg argument, to be the same.

tksfz: IMHO I'd be inclined to leave it as `msg: &str`.  If people want an empty log message they can/should pass in empty string.

tksfz:
Btw  check the conversation here for reference.rs (showing up as hidden for an outdated diff):

So when I pass in `0 as *const libc::c_char`  for msg (to the raw call), everything seems to work (in at least one particular case).  The reflog records an empty message.

Do you  want  me to make `msg` an `Option`?  I could do that but then we might want  to  modify the  existing `reference.rename()` function, which also takes a  msg argument, to be the same.

IMHO I'd be inclined to leave it as `msg: &str`.  If people want an empty log message they can/should pass in empty string.

alexcrichton:
Thanks!

alexcrichton:
(not having an `Option` is fine)

PR #79, opened 2015-07-29
Fix build on OS X

carlosmn:
The recent update to libgit2 v0.23 removed all dependencies from on OS
X. While there is no OpenSSL dependency anymore, the libz-sys crate is
still needed by libgit2-sys/lib.rs.

Even then, libgit2-sys/lib.rs tries to run an OpenSSL function to get
set up with the certs, which fails to compile as we don't have this on
OS X.

Change the conditional compilation flags to make OS X the same as
Windows, an completely OpenSSL-free zone.

### DIFF
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -20,6 +20,10 @@ libc = "0.1"
 [build-dependencies]
 pkg-config = "0.3"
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Gah I knew I forgot to test something! Thanks!

Issue #68, opened 2015-05-27
libgit2-sys provides no way to use a system provided libgit2


jmesmon:

alexcrichton:
The libgit2-sys library specifies [the `links` manifest key](https://github.com/alexcrichton/git2-rs/blob/master/libgit2-sys/Cargo.toml#L6) which [allows overrides](http://doc.crates.io/build-script.html#overriding-build-scripts), but the system library is almost never the right version so that's why the check isn't currently implemented.

jmesmon:
I still think this is an issue, perhaps more accurately stated as:
```
libgit2-sys provides no way to use a system provided libgit2 via pkg-config
```

Notably, this means that pkg-config would need to be driven by an external program and it's output translated into something like the overrides for links (only to be converted later by cargo back into the original form). That seems rather awkward and unnecessary.

alexcrichton:
Right, this was explicitly removed in b5334b057b555d77aed4a0031c7b83b4186aa4f7 because the depended-upon version of libgit2 is rarely found in the wild, and future versions of libgit2 frequently break ABI compatibility, so it really does need to use a precise version for now until the ABI stabilizes.

jmesmon:
I opened this because right now I am packaging a recent enough libgit2, and have to add this nasty patch to git2-rs to get it to be used: https://github.com/jmesmon/meta-rust/blob/master/recipes/cargo/files/git2-rs/0002-libgit2-sys-avoid-the-build-script-it-is-a-disaster.patch

I'd expect it might be a good idea to at least allow some type of override to get the expected pkg-config behavior so packagers can hack up packages a bit less.

alexcrichton:
Hm ok, how about something like if the `LIBGIT2_SYS_USE_PKG_CONFIG` environment variable is set, then `pkg-config` is queried first, but otherwise it's not used at all. Would that work?

jmesmon:
That would be fine.

alexcrichton:
Ok! I've added this in https://github.com/alexcrichton/git2-rs/commit/ed629d783f629838611f07c5940a2608b7f97959

PR #80, opened 2015-08-09
build: always check for libgit2 via pkg-config first

lucab:
I'd like to have libgit2-sys automatically pull the right lib via pkg-config, if a compatible version is available. I'm not sure why #68 introduced a new ENV key instead of explicitly checking for the available version, so this PR just does it.

### DIFF
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -20,6 +20,7 @@ libz-sys = "0.1.0"
 
 [build-dependencies]
 pkg-config = "0.3"
+semver = "0.1"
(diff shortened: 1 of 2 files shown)


alexcrichton:
The ABI of libgit2 is very unstable at the moment, and while this is currently pinned to 0.23.0 any bugfix being pulled in will likely pull in API breakage. This means that using a system library by default is almost always going to lead to breakage, which is why the environment variable exists.

If you want to force usage of a particular libgit2 then you can either use cargo overrides or get pkg-config to succeed and use the environment variable.

PR #81, opened 2015-09-18
implement match_diff for Pathspec

blaenk:
This implements a higher-level binding for [`git_pathspec_match_diff`](https://libgit2.github.com/libgit2/#HEAD/group/pathspec/git_pathspec_match_diff).

It was a very straightforward implementation thanks to the way you set it all up. I just copied the `match_tree` method and substituted a `Diff` instead. Hope I didn't miss anything. It's working great on my end.

This enables me to, for example, add many pathspecs to a `DiffOptions`, compute a diff once (which I imagine is expensive) and then for each of those pathspecs I can see if it's contained in the resulting `Diff` using `match_diff` (I do this because say I add 3 pathspecs "A", "B", and "C" to the `DiffOptions`, I can't just assume that the diff contains them all, it could be that it only contains "A" and "B" but not "C", at least, that's my understanding).

If there's a better way to do that please let me know. Although, doing this instead of what I was previously doing already gave me a huge performance boost.

### DIFF
--- a/src/pathspec.rs
+++ b/src/pathspec.rs
@@ -4,7 +4,7 @@ use std::ops::Range;
 use std::path::Path;
 use libc::size_t;
 
-use {raw, Error, Tree, PathspecFlags, Index, Repository, DiffDelta, IntoCString};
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #82, opened 2015-09-22
implement git_merge_bases

cmsd2:
tested, but I would appreciate someone checking what I've done is sensible, especially the unsafe parts.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -79,6 +79,16 @@ impl Clone for git_strarray {
 }
 
 #[repr(C)]
+#[derive(Copy)]
+pub struct git_oidarray {
+    pub ids: *mut git_oid,
+    pub count: size_t,
+}
+impl Clone for git_oidarray {
+    fn clone(&self) -> git_oidarray { *self }
+}
+
+#[repr(C)]
 pub struct git_signature {
     pub name: *mut c_char,
     pub email: *mut c_char,
@@ -1478,6 +1488,9 @@ extern {
     // strarray
     pub fn git_strarray_free(array: *mut git_strarray);
 
+    // oidarray
+    pub fn git_oidarray_free(array: *mut git_oidarray);
+
     // signature
     pub fn git_signature_default(out: *mut *mut git_signature,
                                  repo: *mut git_repository) -> c_int;
@@ -2127,6 +2140,11 @@ extern {
                           repo: *mut git_repository,
                           one: *const git_oid,
                           two: *const git_oid) -> c_int;
+
+    pub fn git_merge_bases(out: *mut git_oidarray,
+                           repo: *mut git_repository,
+                           one: *const git_oid,
+                           two: *const git_oid) -> c_int;
 
     // pathspec
     pub fn git_pathspec_free(ps: *mut git_pathspec);
--- a/src/oid_array.rs
+++ b/src/oid_array.rs
@@ -6,6 +6,7 @@ use oid::Oid;
 use raw;
 use util::Binding;
 use std::slice;
+use std::mem;
 
 /// An oid array structure used by libgit2
 ///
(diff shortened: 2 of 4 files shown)


libgit2-sys/lib.rs
```diff
@@ -79,6 +79,16 @@ impl Clone for git_strarray {
...
 
 #[repr(C)]
+#[derive(Copy)]
+pub struct git_oidarray {
+    pub ids: *mut git_oid,
+    pub count: size_t,
+}
+impl Clone for git_oidarray {
+    fn clone(&self) -> git_oidarray { *self }
+}
```
alexcrichton: Could this omit the `Clone + Copy` for now? Looking below it looks like it's not required at least?

src/oid_array.rs
```diff
@@ -0,0 +1,110 @@
...
+                let ptr = self.raw.ids.offset(i as isize);
+                Some(Oid::from_raw(ptr))
+            }
+        } else {
+            None
+        }
+    }
+
+    /// Returns None if `i` is out of bounds.
+    pub fn get_bytes(&self, i: usize) -> Option<&[u8]> {
```
alexcrichton: Perhaps this could just expose the `get` method for now?

alexcrichton: Er actually, better idea! Perhaps this could implement `Deref<Target=[Oid]>`?

I was gonna say this may want to implement `Index` but with a deref implementation you should get basically everything!

alexcrichton:
Oh wow nice, thanks!

libgit2-sys/lib.rs
cmsd2: either Copy or Clone seems to be required for the Binding::raw method on OidArray
the Clone methods is implemented in terms of Copy.
seems like it'd be easier to leave as is.

src/oid_array.rs
cmsd2: i'll have a go.

cmsd2:
good, so the ci passed those additional changes, and adding Deref definitely simplifies the code, though i still need your blessing that the unsafe bits are ok.

libgit2-sys/lib.rs
alexcrichton: Ah right!

src/oid_array.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    /// Returns an iterator over the Oids contained within this array.
+    pub fn iter(&self) -> Iter {
+        Iter { range: 0..self.len(), arr: self }
+    }
+
+    /// Returns the number of strings in this array.
+    pub fn len(&self) -> usize { self.raw.count as usize }
```
alexcrichton: You should be able to actually omit all these methods because they're picked up from the `Deref` implementation to a slice.

src/oid_array.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl Deref for OidArray {
+    type Target = [Oid];
+
+    fn deref(&self) -> &[Oid] {
+        unsafe {
+            slice::from_raw_parts(self.raw.ids as *const Oid, self.raw.count as usize)
```
alexcrichton: Could you add a `debug_assert_eq!` that the size of `Oid` is the same as pointee here? e.g.

```rust
debug_assert_eq!(mem::size_of::<Oid>(), mem::size_of_val(&*self.raw.ids))
```

alexcrichton:
Looks good to me! All the unsafe bits are fine by me as well

cmsd2:
you're right, that's a huge improvement using deref to get everything. the iterator now yields &Oid which i think is better too.

src/oid_array.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Some apis return arrays of oids which originate from libgit2. This
+/// wrapper type behaves a little like `Vec<&Oid>` but does so without copying
+/// the underlying Oids until necessary.
+pub struct OidArray {
+    raw: raw::git_oidarray,
+}
+
+impl OidArray {
+    /// Returns None if i is out of bounds.
+    pub fn get(&self, i: usize) -> Option<Oid> {
```
alexcrichton: I think this method is also present on slices, so perhaps it could be deferred in favor of that?

cmsd2:
I've removed the get method now too. there won't be much left at this rate :)

alexcrichton:
Awesome, thanks!

PR #83, opened 2015-09-27
Tree iter lifetime

cmsd2:
I've been trying to crawl over a repo's commits (hence my interest so far)... but now I'm finding the tree api difficult to use.

I'm trying to recurse or iterate over all the objects in a tree, but I'm having trouble with the lifetime restrictions imposed by the tree iterator.

I think my problem is that the TreeIter elides the tree lifetime with the repo lifetime. I don't think that's right.

I've modified tree.rs to show what I mean.

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -26,9 +26,9 @@ pub struct TreeEntry<'tree> {
 }
 
 /// An iterator over the entries in a tree.
-pub struct TreeIter<'tree> {
+pub struct TreeIter<'tree,'repo:'tree> {
     range: Range<usize>,
-    tree: &'tree Tree<'tree>,
+    tree: &'tree Tree<'repo>,
 }
 
 impl<'repo> Tree<'repo> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Unfortunately this may not actually be memory safe! Looking at the implementation of libgit2, it looks like the lifetime of the `TreeEntry` objects are tied to the lifetime of the tree that they originate from, so it's not safe to lift their lifetime up to the lifetime of the repository itself.

I'd love to help fix your problems, though! Perhaps you've got some code I could poke around? Or could you elaborate a bit on why the lifetimes aren't quite working out?

cmsd2:
The first commit in this pull-request shows a breadth-first search of the tree for a single commit, implemented in a struct called TestTreeIter, and without the second commit it demonstrates the compile errors I was getting.
The code is almost exactly the same as RecursiveTreeIter in https://github.com/cmsd2/codelauf/blob/master/src/repo.rs#L42
I should be able to use this to crawl all the files in a bare repo and then I want to send them to elasticsearch.
I figured iterators was a rusty way to do it, but perhaps I haven't got it quite right. I also tried a non-iterator approach using callbacks in both depth and breadth first order. Breadth first suffers the same problems I'm having here in that I try and save the result of temporaries for use in the next level of iteration.
Depth first with a callback I think could still work, but I couldn't get past a compiler problem matching the type signature of an FnMut with lifetimes, with a || closure.

alexcrichton:
Yeah unfortunately this is a case where using an iterator may not quite work out due to the lifetimes in play, it ends up just being a fundamentally hard problem to solve in a memory safe way! Luckily, however, I believe, however, that this is why libgit2 provides `git_tree_entry_dup`, a method to separate the lifetime of the `TreeEntry` from the tree itself. This is currently bound via a [`Clone` impl](https://github.com/alexcrichton/git2-rs/blob/ed0573ae4472b3d965b53398e158160442e21013/src/tree.rs#L187-L195) on `TreeEntry<'a>`, but I unfortunately forgot a matching [`to_owned` method as well](https://github.com/alexcrichton/git2-rs/blob/ed0573ae4472b3d965b53398e158160442e21013/src/signature.rs#L87-L94).

In other words, perhaps this could add a `to_owned` method on `TreeEntry`? That should allow you to store `TreeEntry<'static>` in the iterator, which should in turn allow the BFS to be written?

cmsd2:
gotcha
lets try adding a to_owned instead.

PR #84, opened 2015-09-28
add TreeEntry to_owned

cmsd2:
ok here's the suggested to_owned impl
and the test I added earlier passes using this impl, so I'm happy.


alexcrichton:
Looks good to me, thanks!

PR #86, opened 2015-11-05
Add bindings for git_treebuilder functions

wthrowe:


src/treebuilder.rs
```diff
@@ -0,0 +1,223 @@
...
+        unsafe { raw::git_treebuilder_clear(self.raw) }
+    }
+
+    /// Get the number of entries
+    pub fn len(&self) -> usize {
+        unsafe { raw::git_treebuilder_entrycount(self.raw) as usize }
+    }
+
+    /// Get en entry from the builder from its filename
+    pub fn get<P: AsRef<OsStr>>(&self, filename: P) -> Result<TreeEntry, Error> {
```
wthrowe: Thinking about this some more, `AsRef<[u8]>` is probably more appropriate than `AsRef<OsStr>` here since the query functions all return either `&str` or `&[u8]`.  I'll make that change tomorrow.

libgit2-sys/lib.rs
```diff
@@ -600,6 +603,7 @@ pub enum git_ref_t {
 }
 pub use git_ref_t::*;
 
+enum_from_primitive!{
```
alexcrichton: I'd prefer to not pick up this dependency for now, I think that it may just be best to remove this altogether and make it a raw `i32` as I think that's basically what it's later treated as anyway.

src/tree.rs
```diff
@@ -125,7 +125,7 @@ impl<'tree> TreeEntry<'tree> {
     ///
     /// The lifetime of the entry is tied to the tree provided and the function
     /// is unsafe because the validity of the pointer cannot be guaranteed.
-    unsafe fn from_raw_const(raw: *const raw::git_tree_entry)
+    pub unsafe fn from_raw_const(raw: *const raw::git_tree_entry)
```
alexcrichton: I'd prefer to leave this private for now (inaccessible to external users of the crate), could this be made a public free function in this module?

src/treebuilder.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// Constructor for in-memory trees
+pub struct TreeBuilder<'repo> {
+    raw: *mut raw::git_treebuilder,
+    _marker: marker::PhantomData<&'repo Repository>,
+}
+
+impl<'repo> TreeBuilder<'repo> {
+    /// Create a new TreeBuilder with no entries
+    pub fn new(repo: &'repo Repository) -> Result<TreeBuilder, Error> {
```
alexcrichton: Currently most construction methods for types like this are on the `Repository` type itself, could this move over there with a method name like `treebuilder`?

alexcrichton: Also, it could take an `Option<&Tree>` argument which could allow this and `from_tree` below to be collapsed into one method.

src/treebuilder.rs
alexcrichton: There's actually a local `IntoCString` trait which I've been using for bounds on functions like this, so perhaps that could be used?

src/treebuilder.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// Get the number of entries
+    pub fn len(&self) -> usize {
+        unsafe { raw::git_treebuilder_entrycount(self.raw) as usize }
+    }
+
+    /// Get en entry from the builder from its filename
+    pub fn get<P: AsRef<OsStr>>(&self, filename: P) -> Result<TreeEntry, Error> {
+        let filename = try!(filename.as_ref().into_c_string());
+        unsafe {
+            let ret = raw::git_treebuilder_get(self.raw, ::call::convert(&filename));
```
alexcrichton: Instead of calling `::call::convert` manually, could this just use `.as_ptr()`?

alexcrichton: Also, this'll need a check to see whether the returned pointer is NULL in which case an error should be returned.

src/treebuilder.rs
```diff
@@ -0,0 +1,223 @@
...
+        Ok(())
+    }
+
+    /// Selectively remove entries from the tree
+    pub fn filter<F>(&mut self, mut filter: F)
+                     where F: FnMut(&TreeEntry) -> FilterResult {
+        let mut cb: &mut FilterCb = &mut filter;
+        let ptr = &mut cb as *mut _;
+        unsafe {
+            raw::git_treebuilder_filter(self.raw, filter_cb, ptr as *mut _);
```
alexcrichton: Could this use `try_call!` to avoid the explicit `panic::check`?

src/treebuilder.rs
```diff
@@ -0,0 +1,223 @@
...
+    Discard,
+}
+
+wrap_env! {
+    fn filter_cb(entry: *const raw::git_tree_entry,
+                 payload: *mut c_void) -> c_int {
+        unsafe {
+            // There's no way to return early from git_treebuilder_filter, so
+            // just keep panicing until it stops calling us.
+            panic::check();
```
alexcrichton: I think this may need a new helper method to return early because otherwise this'll just keep taking the same panic result out of TLS and putting it back in (with another box).

alexcrichton:
Should this also expose [`git_treebuilder_write`](https://libgit2.github.com/libgit2/#HEAD/group/treebuilder/git_treebuilder_write)?

src/treebuilder.rs
wthrowe: Oh, yeah, that's even better.

libgit2-sys/lib.rs
wthrowe: Do we know that `git_filemode_t` is layout-compatible with an `i32`?  It looks like C could choose a representation using either `i32` or `u16`.  (Although it looks like anything `rustc` currently supports will use `i32`, so we're probably OK.)

src/treebuilder.rs
wthrowe: Good ideas.  Will do.

src/treebuilder.rs
wthrowe: Will fix.

src/treebuilder.rs
wthrowe: No, because `git_treebuilder_filter` returns `void`.

src/treebuilder.rs
wthrowe: I can change that.  I missed that we gained a box on each iteration.

wthrowe:
`fn write` is there.  It's the last method in the `impl` block.

wthrowe:
New commit added.  (I'm happy to squash them but figured this might make reviewing easier.)

The only change from my previous replies is that I decided to return a `Result<Option<TreeEntry>>` from `get` rather than erroring on nonexistence to be more consistent with other `get`-like methods.

wthrowe:
Oh, and I haven't changed the filemode thing because I want your opinion on enum representations first.

libgit2-sys/lib.rs
alexcrichton: Yeah there's a few other enums that are exposed as `i32` as they're just a pile of flags. Or, to be more concrete, they're exposed as `__enum_ty` as apparently enums change signedness on MSVC!

src/treebuilder.rs
alexcrichton: Oh, right!

alexcrichton:
Oh oops, now I see `write`!

src/treebuilder.rs
```diff
@@ -0,0 +1,207 @@
...
+        let filename = try!(filename.into_c_string());
+        unsafe {
+            try_call!(raw::git_treebuilder_remove(self.raw, filename));
+        }
+        Ok(())
+    }
+
+    /// Selectively remove entries from the tree
+    pub fn filter<F>(&mut self, mut filter: F)
+                     where F: FnMut(&TreeEntry) -> FilterResult {
```
alexcrichton: I'd be fine having this just return a `bool` as `filter`'s pretty common with iterators already so the enum may not be buying mcuch

libgit2-sys/lib.rs
wthrowe: Ah, excellent!  If we already have a type for that then there's no problem making the change.

wthrowe:
I made a mess of my git history so I just squashed everything now.  The new stuff is the `git_filemode_t` change and removing the `FilterResult` enum in exchange for a warning in the documentation.

alexcrichton:
Thanks!

PR #91, opened 2016-01-13
Add `FromStr` implementation for `Oid`

hauleth:
This will allow `Oid` parsing with `str::parse`.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -81,6 +81,18 @@ impl fmt::Display for Oid {
     }
 }
 
+impl str::FromStr for Oid {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh wow I'm not sure why this wasn't here in the first place... thanks!

hauleth:
I haven't removed `Oid::from_str` as it would need version bump. It should be marked as deprecated and removed ASAP.

alexcrichton:
Ah it's fine to stick around, there's no harm in both an inherent and a trait method

PR #93, opened 2016-01-13
[WIP] Remove unneeded FFI calls

hauleth:
Reduce FFI calls. Some of them can be made in safe, idiomatic Rust without FFI calls. It can be also step forward to make `git2-rs` independent from `libgit2-sys`.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 use std::cmp::Ordering;
 use std::hash::{Hasher, Hash};
+use std::str;
 use libc;
 
 use {raw, Error};
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -1,7 +1,6 @@
 use std::fmt;
 use std::cmp::Ordering;
 use std::hash::{Hasher, Hash};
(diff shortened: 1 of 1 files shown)


alexcrichton:
Out of curiosity, is there motivation for this change? I don't think it's really feasible to envision this all written in Rust in the near future, and I'd prefer to lean on libgit2 as much as possible to ensure we don't get anything wrong

hauleth:
For now I am only replacing functions that are simple enough that calling `libgit2

tkelman:
Did you see https://github.com/libgit2/libgit2/pull/3540?

alexcrichton:
@tkelman fascinating! Thanks for the link!

@hauleth ok in that case for now I think I'm gonna leave this as-is to binding to libgit2 wherever possible, but rewriting in Rust sounds really cool! It seems there's even upstream interest in doing so as well :)

hauleth:
@tkelman I haven't know that. Nice.

@alexcrichton ok, no problem. In that case I hope that you don't mind that I use this crate as a base for Rust-native libgit2.

PR #101, opened 2016-02-06
`Diff.foreach` support

Nemo157:
For some reason the ignored test causes segmentation faults. I've done a little debugging and as far as I can tell it's entirely inside `libgit2` code and has nothing to do with the `foreach` code I've added. It seems like the test repository isn't being fully setup correctly. I've used this new functionality on a project of mine running against real repositories and haven't encountered any crashes. I'll come back to the test at some point soon, any hints you could give me on why the test repository might be different would be useful.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -1188,12 +1188,13 @@ mod tests {
         }
 
         #[test]
-        #[ignore]
         fn file_and_hunk() {
             let path = Path::new("foo");
             let (td, repo) = ::test::repo_init();
             t!(t!(File::create(&td.path().join(path))).write_all(b"bar"));
-            let diff = t!(repo.diff_tree_to_workdir(None, Some(DiffOptions::new().include_untracked(true))));
+            let mut index = t!(repo.index());
+            t!(index.add_path(path));
+            let diff = t!(repo.diff_tree_to_index(None, Some(&index), Some(DiffOptions::new().include_untracked(true))));
             let mut new_lines = 0;
             t!(diff.foreach(
                 &mut |_file, _progress| { true },
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -1143,6 +1143,12 @@ impl DiffFindOptions {
 
 #[cfg(test)]
 mod tests {
+    use DiffOptions;
+    use std::fs::File;
+    use std::path::Path;
+    use std::borrow::Borrow;
+    use std::io::Write;
+
     #[test]
     fn smoke() {
         let (_td, repo) = ::test::repo_init();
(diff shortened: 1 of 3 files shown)


src/diff.rs
```diff
@@ -765,6 +901,80 @@ impl Drop for DiffStats {
...
+    unsafe fn from_raw(raw: raw::git_diff_binary_t) -> DiffBinaryKind {
+        match raw {
+            raw::GIT_DIFF_BINARY_NONE => DiffBinaryKind::None,
+            raw::GIT_DIFF_BINARY_LITERAL => DiffBinaryKind::Literal,
+            raw::GIT_DIFF_BINARY_DELTA => DiffBinaryKind::Delta,
+            _ => panic!("Unknown git diff binary kind"),
+        }
+    }
+    fn raw(&self) -> raw::git_diff_binary_t {
+        match self {
```
alexcrichton: stylistically this is typically done via `match *self` and then you can avoid the leading `&` in front of each arm

src/diff.rs
```diff
@@ -74,8 +74,46 @@ pub struct DiffStats {
     raw: *mut raw::git_diff_stats,
 }
 
+/// Structure describing the binary contents of a diff.
+pub struct DiffBinary<'diff> {
+    raw: *const raw::git_diff_binary,
+    _marker: marker::PhantomData<&'diff raw::git_diff_binary>,
```
alexcrichton: Currently this is largely being expressed as:

```rust
_marker: marker::PhantomData<&'diff Diff>,
```

which is to say the `PhantomData` types tend to hold the Rust types rather than the C ones

src/diff.rs
```diff
@@ -74,8 +74,46 @@ pub struct DiffStats {
...
+pub struct DiffBinaryFile<'diff> {
+    raw: *const raw::git_diff_binary_file,
+    _marker: marker::PhantomData<&'diff raw::git_diff_binary_file>,
+}
+
+/// When producing a binary diff, the binary data returned will be
+/// either the deflated full ("literal") contents of the file, or
+/// the deflated binary delta between the two sides (whichever is
+/// smaller).
+#[derive(Copy, Clone)]
```
alexcrichton: Perhaps `Debug` here as well?

src/diff.rs
```diff
@@ -765,6 +901,80 @@ impl Drop for DiffStats {
     }
 }
 
+impl<'a> DiffBinary<'a> {
```
alexcrichton: Could you change this `'a` to `'diff`?

src/diff.rs
```diff
@@ -939,4 +1149,59 @@ mod tests {
         assert_eq!(stats.deletions(), 0);
         assert_eq!(stats.files_changed(), 0);
     }
+
+    mod foreach {
```
alexcrichton: Ah it's ok to just shove these all in the same module!

alexcrichton:
Oh gah it looks like the initial `Diff` implementation messed up, it needs to have a lifetime connected back to the `Repository` (e.g. `Diff<'repo>`).

Unfortunately that's a bit of a breaking change, but as it's required for safety I'd be fine doing it on the main branch anyway. Would you be interested in adding that in?

Nemo157:
Sure, I can do that now, just as part of this PR ok or would you prefer a separate one for that?

alexcrichton:
Yeah part of this PR is fine by me

src/diff.rs
```diff
@@ -74,51 +75,89 @@ pub struct DiffStats {
...
+/// Structure describing the binary contents of a diff.
+pub struct DiffBinary<'diff> {
+    raw: *const raw::git_diff_binary,
+    _marker: marker::PhantomData<&'diff Diff<'diff>>,
+}
+
+/// The contents of one of the files in a binary diff.
+pub struct DiffBinaryFile<'diff> {
+    raw: *const raw::git_diff_binary_file,
+    _marker: marker::PhantomData<&'diff Diff<'diff>>,
```
Nemo157: Is this ok or do I need to do something like

```rust
pub struct DiffBinary<'diff, 'repo: 'diff> {
  _marker: PhantomData<&'diff Diff<'repo>>,
}
```

I tried a couple of changes to the test that I thought might break with this and they didn't so I assume there's some sort of greater than relationship automatically happening with the lifetimes here?

src/diff.rs
Nemo157: I did this as an alternative to mentioning the method name in each test, I like what it does for the `cargo test` output, giving lines like `test diff::tests::foreach::file_and_hunk`. I can pull them out and put a `foreach_` prefix on them if you think that would be better.

alexcrichton: ah yeah let's go with a prefix to avoid extra indentation

src/diff.rs
alexcrichton: I *think* this is ok as-is (not having two lifetimes) because there's no need to mention the `'repo` lifetime in the implementation

src/diff.rs
```diff
@@ -130,7 +169,7 @@ impl Diff {
     /// as if the old version was from the "onto" list and the new version
     /// is from the "from" list (with the exception that if the item has a
     /// pending DELETE in the middle, then it will show as deleted).
-    pub fn merge(&mut self, from: &Diff) -> Result<(), Error> {
+    pub fn merge(&mut self, from: &Diff<'repo>) -> Result<(), Error> {
```
alexcrichton: I think that this can probably be an anonymous lifetime

alexcrichton:
Thanks! I may touch up the style a bit when merging, but just a few last nits

Nemo157:
That's cool. I tried following where I could but some of my vim plugins make that a bit hard, e.g. lining up parameters is always a little off :disappointed:

```rust
 332 extern λ line_cb_c(delta: *const raw::git_diff_delta,
 333                     hunk: *const raw::git_diff_hunk,
 334                     line: *const raw::git_diff_line,
 335                     data: *∫ c_void) → c_int {
```

Thanks for figuring out why my test wasn't working :smile_cat:

alexcrichton:
Thanks!

PR #104, opened 2016-02-11
Fix lifetime names

Nemo157:
Just realised I screwed up the naming for these lifetimes, the binary diffs are only usable inside the callbacks, they don't live for the lifetime of the Diff. Luckily the lifetime inference is correct so it's just the naming that's off.

Found when I was trying to come up with some way to create a tree of iterators to use instead of this callback directly, really disappointed that libgit2 doesn't have a way to externally iterate over the diff lines as well as this callback system :cry:.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -76,15 +76,15 @@ pub struct DiffStats {
 }
 
 /// Structure describing the binary contents of a diff.
-pub struct DiffBinary<'diff> {
+pub struct DiffBinary<'a> {
     raw: *const raw::git_diff_binary,
-    _marker: marker::PhantomData<&'diff Diff<'diff>>,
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think that `'diff` here is actually correct. For example the `git_diff_binary_file` has a `const char*` pointer, but that can really only possibly be a pointer in to the `git_diff` it originated from, I believe, so these functions basically act as if they have a pointer over to a `Diff<'repo>` structure.

Nemo157:
https://github.com/libgit2/libgit2/blob/master/src/diff_patch.c#L310-L337

The binary objects are created on the stack and the data is temporarily created for the callback call.

alexcrichton:
Whelp! Nevermind then!

Thanks :)

PR #106, opened 2016-02-17
Add iOS targets to Cargo.toml

nanotech:
Depends on <https://github.com/alexcrichton/ssh2-rs/pull/35>.


alexcrichton:
I believe libgit2 attempts to use Security.framework wherever possible, so in theory the openssl dep shouldn't be necessary?

nanotech:
Oh, I think I see now. Then it's the `cfg`s in `libgit2-sys/lib.rs` that need adjusting instead. I'll open a new PR for that.

PR #109, opened 2016-02-18
Add `impl AsRef<[u8]> for Oid` (closes #108)

### LINKED ISSUE #108: Should `Oid` implement `AsRef<[u8]>`
Not entirely certain this makes sense, but it would allow avoiding a call to `as_bytes` when passing an `Oid` into functions taking in `&[u8]` generically.

Looking through `std` there's precedence in that `str.as_bytes()` and `AsRef::<[u8]>::as_ref(str)` are equivalent.

Nemo157:


alexcrichton:
Thanks!

PR #116, opened 2016-03-04
Add IntoIterator trait around Tree::iter() method

Parakleta:
Tree::iter() already returns a TreeIter which implements the Iterator trait, so this is a simple convenience.


alexcrichton:
Thanks!

PR #117, opened 2016-03-08
Add libgit2 stream filter binding

chyh1990:
Rust binding for libgit2 stream filter API, so that applications using git2-rs can provide smudge/clean filter support (e.g. Github LFS service).

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -1,0 +1,326 @@
+//! Interfaces for adding custom filter to libgit2
+
+use libc::{c_int, c_char, c_void, size_t};
+use std::{slice, ptr, marker};
+use {raw, call, util, Error, panic};
+use std::ffi::CString;
+use std::path::Path;
+use raw::{git_filter, git_writestream};
+
+#[repr(C)]
+struct Filter {
+    filter: git_filter,
+    attrs: Option<CString>,
+    instance: Box<StreamFilter>,
+}
+
+#[repr(C)]
+struct Stream {
+    parent: git_writestream,
+    handler: Box<WriteStream>,
+}
+
+/// Wrapper of the raw git_writestream object, representing the input
+/// of next filter of the chain
+struct RawWriteStream {
+    stream: *mut git_writestream,
+}
+
+/// A trait representing operations to handle filter data stream.
+///
+/// These callbacks are typically called by libgit2.
+pub trait WriteStream : Send + 'static {
+    /// New stream data, Implementations should process the buffer data
+    /// and then write the output to the next stream if needed.
+    fn write(&mut self, buf: &[u8]) -> Result<(), Error>;
+
+    /// Close stream. Implementations should do finalizing work and then
+    /// close the next stream.
+    fn close(&mut self) -> Result<(), Error>;
+}
+
+/// Information to construct a new stream filter.
+pub struct FilterSource<'a> {
+    raw: *const raw::git_filter_source,
+    _marker: marker::PhantomData<&'a str>,
+}
+
+impl<'a> FilterSource<'a> {
+    fn from_raw(raw: *const raw::git_filter_source) -> FilterSource<'a> {
+        FilterSource {
+            raw: raw,
+            _marker: marker::PhantomData,
+        }
+    }
+
+    /// Returns `true` if it is a smudge filter, else a clean filter.
+    pub fn to_worktree(&self) -> bool {
+        unsafe {
+            raw::git_filter_source_mode(self.raw) == raw::GIT_FILTER_TO_WORKTREE
+        }
+    }
+
+    /// Returns the path that the source data is coming from.
+    pub fn source_path(&self) -> &Path {
+        unsafe {
+            let ptr = raw::git_filter_source_path(self.raw);
+            util::bytes2path(::opt_bytes(self, ptr).unwrap())
+        }
+    }
+}
+
+/// A trait representing a customized filter to be registered into libgit2
+/// filter chain.
+///
+/// Modeled after git2/sys/filter.h in a streaming manner.
+pub trait StreamFilter : Send + Sync + 'static {
+	/// Called when the filter is first used for any file.
+    fn init(&self) -> Result<(), Error> { Ok(()) }
+
+    /// Create a new stream to handle new blob, should return a initialized
+    /// WriteStream.
+    ///
+    /// Custom filters can retrieve information of source blob from `source`,
+    /// and custom filter output should be written to `next` later.
+    fn open_stream(&self, source: &FilterSource, next: Box<WriteStream>)
+                   -> Result<Box<WriteStream>, Error>;
+}
+
+unsafe impl Send for RawWriteStream {}
+
+impl WriteStream for RawWriteStream {
+    fn write(&mut self, buf: &[u8])
+        -> Result<(), Error> {
+        let code = unsafe {
+            ((*self.stream).write)((*self).stream,
+                buf.as_ptr() as *const c_char, buf.len())
+        };
+        call::try(code).and(Ok(()))
+    }
+
+    fn close(&mut self) -> Result<(), Error> {
+        let code = unsafe { ((*self.stream).close)(self.stream) };
+        call::try(code).and(Ok(()))
+    }
+}
+
+extern fn stream_write(stream: *mut git_writestream,
+                       buf: *const c_char,
+                       len: size_t) -> c_int {
+    let ws = stream as *mut Stream;
+    panic::wrap(|| unsafe {
+        let input = slice::from_raw_parts(buf as *const u8, len);
+        match (*ws).handler.write(input) {
+            Ok(..) => 0,
+            Err(e) => e.raw_code() as c_int,
+        }
+    }).unwrap_or(-1)
+}
+
+extern fn stream_free(stream: *mut git_writestream) {
+    let _ = unsafe { Box::from_raw(stream as *mut Stream) };
+}
+
+extern fn stream_close(stream: *mut git_writestream) -> c_int {
+    let ws = stream as *mut Stream;
+    panic::wrap(|| unsafe {
+        match (*ws).handler.close() {
+            Ok(..) => 0,
+            Err(e) => e.raw_code() as c_int,
+        }
+    }).unwrap_or(-1)
+}
+
+extern fn filter_init(filter: *mut git_filter) -> c_int {
+    let filter = filter as *mut Filter;
+    panic::wrap(|| unsafe {
+        match (*filter).instance.init() {
+            Ok(..) => 0,
+            Err(e) => e.raw_code() as c_int,
+        }
+    }).unwrap_or(-1)
+}
+
+extern fn filter_shutdown(filter: *mut git_filter) {
+    let _ = unsafe { Box::from_raw(filter as *mut Filter) };
+}
+
+extern fn filter_stream_init(out: *mut *mut raw::git_writestream,
+                             filter: *mut git_filter,
+                             _payload: *mut *mut c_void,
+                             src: *const raw::git_filter_source,
+                             next: *mut raw::git_writestream) -> c_int
+{
+    let filter = filter as *mut Filter;
+    // XXX: should bound lifetime of `next` and `source` to some hidden
+    // target blob object
+    let next = RawWriteStream { stream: next };
+    let source = FilterSource::from_raw(src);
+    panic::wrap(|| unsafe {
+        match (*filter).instance.open_stream(&source, Box::new(next)) {
+            Ok(s) => {
+                let stream = Box::new(Stream {
+                    parent: git_writestream {
+                        write: stream_write,
+                        close: stream_close,
+                        free: stream_free,
+                    },
+                    handler: s,
+                });
+                let p = Box::into_raw(stream) as *mut git_writestream;
+                *out = p;
+                0
+            },
+            Err(e) => e.raw_code() as c_int,
+        }
+    }).unwrap_or(-1)
+}
+
+impl Filter {
+    /// create a command filter
+    fn new<T: StreamFilter>(user_filter: T, attrs: Option<&str>) -> Filter {
+        let mut f = Filter {
+            filter: git_filter {
+                version: raw::GIT_FILTER_VERSION,
+                attributes: ptr::null(),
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -116,6 +116,7 @@ pub use time::{Time, IndexTime};
 pub use tree::{Tree, TreeEntry, TreeIter};
 pub use treebuilder::TreeBuilder;
 pub use util::IntoCString;
+pub use filter::{StreamFilter, WriteStream, FilterChain, RawStream, FilterSource};
```
alexcrichton: API-wise, maybe this could work like transports where `filter` is just a public module with exports inside of it?

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+                cleanup: Some(filter_cleanup),
+            },
+            instance: f
+        };
+        f
+    }
+}
+
+/// The global filter chain.
+pub struct FilterChain;
```
alexcrichton: This may be best expressed as just two global functions rather than a public struct to hang methods on

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+use {raw, call, util, Error};
+use std::ffi::{CString, CStr};
+use std::path::PathBuf;
+use raw::{git_filter, git_writestream};
+
+
+#[repr(C)]
+struct Filter<T> {
+    filter: git_filter,
+    instance: T,
```
alexcrichton: To avoid monomorphization costs, perhaps this could store `Box<StreamFilter>`?

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub path: PathBuf,
+    /// A sumgde filter if true, otherwise a clean filter.
+    pub to_worktree: bool,
+}
+
+/// A trait representing a customized filter to be registered into libgit2
+/// filter chain.
+///
+/// Modeled after git2/sys/filter.h in a streaming manner.
+pub trait StreamFilter {
```
alexcrichton: This'll need to inherit from `Send + Sync + 'static` I believe to be safe.

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Customized WriteStream implementation.
+    type S: WriteStream;
```
alexcrichton: Instead of having an associated type this may want to just return `Box<WriteStream>` below to avoid extra monomorphization costs

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+/// Wrapper of the raw git_writestream object, representing the input
+/// of next filter of the chain
+pub struct RawStream {
+    stream: *mut git_writestream,
+}
+
+/// A trait representing operations to handle filter data stream.
+///
+/// These callbacks are typically called by libgit2.
+pub trait WriteStream {
```
alexcrichton: Let's have this inherit from `Send + 'static` for now, although we may be able to relax those in the future. I'd want to double check that libgit2 matches the ownership model we've got in the usage of a stream as well.

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+
+
+#[repr(C)]
+struct Filter<T> {
+    filter: git_filter,
+    instance: T,
+}
+
+#[repr(C)]
+struct Stream<T> {
```
alexcrichton: Perhaps this could be named `RawWriteStream` and the above type could be named `RawFilter`?

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+        *out = p;
+    }
+    mem::forget(stream);
+    raw::GIT_OK
+}
+
+extern fn filter_cleanup(_filter: *mut git_filter,
+                         _payload: *mut *mut c_void)
+{
+}
```
alexcrichton: A number of these look like they're not forwarded to the underlying trait, but perhaps they could and the trait could have default noop methods?

src/filter.rs
```diff
@@ -0,0 +1,300 @@
...
+{
+    let mode = unsafe { raw::git_filter_source_mode(src) };
+    let filter = filter as *mut Filter<T>;
+    let next = RawStream { stream: next };
+    let source = FilterSource {
+        to_worktree: (mode == raw::GIT_FILTER_TO_WORKTREE),
+        path: unsafe { util::bytes2path(CStr::from_ptr(raw::git_filter_source_path(src)).to_bytes())
+            .to_path_buf() }
+    };
+    let stream = unsafe { match (*filter).instance.open_stream(source, next) {
```
alexcrichton: A number of locations like this will need the panic protection around them to ensure that panics are handled.

src/filter.rs
chyh1990: Yeah, i'm struggling whether i should use trait object here, maybe there won't be so such different kinds of filters and monomorphization costs are not so high?

chyh1990:
Amend this patch according to the comments :-)

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+                cleanup: None,
+            },
+            instance: Box::new(user_filter)
+        };
+        f
+    }
+}
+
+/// register a command filter
+pub fn register<T: StreamFilter>(name: &str, attributes: Option<&str>,
```
alexcrichton: Stylistically in this library I tend to prefer smaller generic functions, so perhaps this could be defined as:

```
pub fn register<T: StreamFilter>(...) -> Result {
    _register(..., Box::new(filter))
}

fn _register(..., filter: Box<StreamFilter>) -> Result {
    ...
}
```

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+                // this callback is not needed, custom filter can implement
+                // passthrough easily
+                check: None,
+
+                // do not use legency filter model
+                apply: None,
+                stream: Some(filter_stream_init),
+
+                // stream filter do not need this
+                cleanup: None,
```
alexcrichton: This, apply, and check above seem like they may want to be bound? How come they aren't part of the `StreamFilter` trait?

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+    raw::GIT_OK
+}
+
+extern fn filter_shutdown(filter: *mut git_filter) {
+    let _ = unsafe { Box::from_raw(filter as *mut Filter) };
+    // dropped
+}
+
+extern fn filter_stream_init(out: *mut *mut raw::git_writestream,
+                        filter: *mut git_filter,
```
alexcrichton: The alignment here may be a bit off

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+                        _payload: *mut *mut c_void,
+                        src: *const raw::git_filter_source,
+                        next: *mut raw::git_writestream) -> c_int
+{
+    let filter = filter as *mut Filter;
+    let next = RawWriteStream { stream: next };
+    let source = unsafe { FilterSource::from_raw(src) };
+    let stream = unsafe { match (*filter).instance.open_stream(source, next) {
```
alexcrichton: Stylistically if there ends up being multiple unsafe blocks like this in a function I tend to just wrap them all up into one unsafe block.

alexcrichton: This'll also need to be wrapped up in a call to catch the panic if one happens.

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+    unsafe {
+        match (*ws).handler.close() {
+            Ok(..) => raw::GIT_OK,
+            Err(_) => raw::GIT_ERROR,
+        }
+    }
+}
+
+extern fn filter_init(_filter: *mut git_filter) -> c_int {
+    raw::GIT_OK
```
alexcrichton: Perhaps this could be bound as an entry point from libgit2? It could be an empty default function for now returning `Ok(())` as it's not necessarily required.

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+extern fn stream_free(stream: *mut git_writestream) {
+    let _ = unsafe { Box::from_raw(stream as *mut Stream) };
+}
+
+extern fn stream_close(stream: *mut git_writestream) -> c_int {
+    let ws = stream as *mut Stream;
+    unsafe {
+        match (*ws).handler.close() {
```
alexcrichton: This'll need to also be wrapped up in a block that catches panics (same with calls above as well)

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/// A trait representing a customized filter to be registered into libgit2
+/// filter chain.
+///
+/// Modeled after git2/sys/filter.h in a streaming manner.
+pub trait StreamFilter : Send + Sync + 'static {
+    /// Create a new stream to handle new blob, should return a initialized
+    /// WriteStream.
+    fn open_stream(&self, source: FilterSource, next: RawWriteStream)
```
alexcrichton: Could you also describe what the `source` and `next` arguments here are?

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// close the next stream.
+    fn close(&mut self) -> Result<(), Error>;
+}
+
+/// Information to construct a new stream filter.
+pub struct FilterSource {
+    /// Path to the source file to be filter.
+    pub path: PathBuf,
+    /// A sumgde filter if true, otherwise a clean filter.
+    pub to_worktree: bool,
```
alexcrichton: I tend to prefer to use private fields with method accessors to allow for expansion in the future.

alexcrichton: Also, to mirror the design of the rest of this library, perhaps this should just contain the raw value? It looks like it may also need a lifetime parameter to ensure it's not sent to other threads.

src/filter.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/// A trait representing operations to handle filter data stream.
+///
+/// These callbacks are typically called by libgit2.
+pub trait WriteStream : Send + 'static {
+    /// New stream data, Implementations should process the buffer data
+    /// and then write the output to the next stream if needed.
+    fn write(&mut self, buf: &[u8])
+        -> Result<(), Error>;
```
alexcrichton: This probably doesn't need to be on a new line

src/filter.rs
chyh1990: `apply` must be `None` for a stream filter (otherwise `stream` won't be called).
The `stream` API is the new API for filter, which do not read the whole blob into memory.

The `check` API is not exposed, because I feel it difficult to express it in safe way
(e.g. `payload: *mut *mut c_void`, `attri_vars: *mut *const c_char`). According to
libgit2 document, it is mainly used for attributes like `filter=*`.

See https://github.com/libgit2/libgit2/blob/master/tests/filter/wildcard.c

Maybe we can't add it to the trait (with default implementation) later?

src/filter.rs
chyh1990: fixed

chyh1990:
Besides the problem regarding `check` callback, I remove `unregister()` in this patch (`Transport` also does not support unregister).

The root cause is that, libgit2 do not call `shutdown` if the filter is unregistered but not used. However, 
`shutdown` will always be called in `git_libgit2_shutdown` for registered filters. Strange behavious...
And I think there a potential memory leak in libgit2's example, see https://github.com/libgit2/libgit2/blob/master/tests/filter/custom_helpers.c

Also see libgit2's implementation:

https://github.com/libgit2/libgit2/blob/master/src/filter.c#L228
https://github.com/libgit2/libgit2/blob/master/src/filter.c#L316

src/filter.rs
```diff
@@ -0,0 +1,322 @@
...
+pub trait StreamFilter : Send + Sync + 'static {
+	/// Called when the filter is first used for any file.
+    fn init(&self) -> Result<(), Error> { Ok(()) }
+
+    /// Create a new stream to handle new blob, should return a initialized
+    /// WriteStream.
+    ///
+    /// Custom filters can retrieve information of source blob from `source`,
+    /// and custom filter output should be written to `next` later.
+    fn open_stream(&self, source: FilterSource, next: RawWriteStream)
```
alexcrichton: Perhaps this could pass in a `Box<WriteStream>` rather than a `RawWriteStream`? May give us some more flexibility in the future to extend this

src/filter.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn write(&mut self, buf: &[u8]) -> Result<(), Error>;
+
+    /// Close stream. Implementations should do finalizing work and then
+    /// close the next stream.
+    fn close(&mut self) -> Result<(), Error>;
+}
+
+/// Information to construct a new stream filter.
+pub struct FilterSource {
+    raw: *const raw::git_filter_source
```
alexcrichton: I don't think that this owns the pointer so this'll want a `<'a>` lifetime parameter on it somewhere I believe.

src/filter.rs
alexcrichton: Also maybe this should pass in `&FilterSource` as it's not passing ownership down?

src/filter.rs
```diff
@@ -0,0 +1,322 @@
...
+                    parent: git_writestream {
+                        write: stream_write,
+                        close: stream_close,
+                        free: stream_free,
+                    },
+                    handler: s,
+                });
+                let p = (&mut *stream) as *mut Stream as *mut git_writestream;
+                *out = p;
+                mem::forget(stream);
```
alexcrichton: Maybe `Box::into_raw` could be used here instead of `mem::forget`?

src/filter.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+        f
+    }
+}
+
+/// register a command filter.
+pub fn register<T: StreamFilter>(name: &str, attributes: Option<&str>,
+                                 priority: i32, filter: T)
+                                 -> Result<(), Error> {
+    let f = Filter::new(filter, attributes);
```
alexcrichton: Could this call be pushed into `_register`? The `filter` is just boxed up in there anyway, right?

src/filter.rs
alexcrichton: Hm is `check` important as an optimization to avoid unnecessarily creating streams?

Thanks for the info about `apply`! Could you expand the comment here to say that it's an essentially deprecated way of doing things in libgit2?

Also, should `cleanup` get bound here?

alexcrichton:

PR #120, opened 2016-03-22
Packbuilder

Nemo157:
Obviously this still needs the documentation added and some more tests. I just wanted to get some feedback on using multiple structs for safety. The two things that I've done this for are getting the hash of the pack and thread safety with the callbacks.

Getting the hash is only valid after the pack has been written somewhere, and once you insert a new item to the pack the hash is invalid until you re-write it. I think this could be done as a runtime check against `object_count` and returning an `Option<Oid>`, but by instead returning a struct holding a mutable borrow against the builder that is the only way to get the hash you are guaranteed that you can't insert any new objects until the struct is dropped. In the common case of not actually caring about the hash the struct will be dropped straight away (and hopefully the construction and drop could even be optimised out) and you can go on using the builder as you want.

Thread safety is much more important. I spent a while looking at the `libgit2` code around this and verified what guarantees it gives around the callbacks. This seems like the sort of thing that `libgit2` should actually document to ensure they aren't violated in the future, but anyway, the guarantees as I see them:

 * If `libgit2` was built with threading disabled or the packbuilder has its thread count set to 1 (the default):
   * All operations and callbacks will run on the thread that calls one of the the `write` methods.
   * Returning false from the callback will stop the build and the callback will not be invoked again.
 * If the packbuilder has its thread count set to anything other than 1:
   * The progress callback will be called from whichever thread next has some progress to report when the timeout between reports expires.
   * Access to the progress callback is locked by a mutex so the callback will not be called concurrently on multiple threads.
   * Returning false from the callback will _eventually_ stop the build, the callback may (by my reading of the code "will almost certainly") be invoked again from other threads until the build has stopped.
 * The `foreach` callback will only ever be called from within the main thread that `foreach` was invoked on.

So the unthreaded case is basically the same as all the other `git2` callbacks, the threaded case needs some special handling. The two changes I could think of to make the threaded case safe are:

 * The callback must be `Send` as it may be invoked from other threads spawned by `libgit2`.
 * If the callback panics we must ensure it doesn't get invoked again. Normally returning false handles this for us, but as all the threads aren't guaranteed to stop together the C callback may be re-invoked so needs to keep track of panicking and avoid re-invoking the rust callback.

The second change is an internal implementation detail. The first however affects the public API for the pack builder. To minimize issues with this I created two "subclasses" of `PackBuilder`, one threaded on non-threaded. Calling the `set_threads` method consumes the `PackBuilder` and returns a `ThreadedPackBuilder` that only accepts `Send` callbacks for its progress callback. If you set the progress callback first then that method consumes the `PackBuilder` and returns a `NonThreadedPackBuilder` that doesn't have the `set_threads` method.

I've temporarily pushed the documentation to https://nemo157.com/git2-rs/packbuilder/doc/git2/struct.Repository.html#method.packbuilder if you want to check how it looks with all the `Deref` implementations.


Nemo157:
Also, I've used this and successfully cloned a repo from a pack built with it :smile: now just need https://github.com/rust-lang/cargo/pull/2385 so I can actually build and deploy those changes to my server while I finish this off :wink:.

alexcrichton:
Thanks for the PR @Nemo157! Always nice to expand the API bindings :)

I wonder if perhaps the state-based transition between packbuilder types may be a bit unergonomic for now? It's a lot of structs to juggle and it seems the primary benefit is that the closure's bounds are either `Send` or not. Maybe we could go with just `Send` for now, and if necessary we can add a separate one-off builder for adding a non-`Send` closure which hardcodes threads to 0?

src/packbuilder.rs
```diff
@@ -0,0 +1,377 @@
...
+            } else {
+                false
+            }
+        });
+        if r.is_none() {
+            // git can attempt to call the progress callback
+            // again even if it returns false.
+            // If the callback panicked then we probably want
+            // to ensure it doesn't get called again.
+            (*data).callback.take();
```
Nemo157: I probably should have looked more into how that was implemented, just assumed it couldn't be doing anything like that by the method signature :disappointed:.

How will the `thread_local` interact with threads spawned by `libgit2`? If it does remain thread local then that implementation won't work for this case; the callback could panic on one thread then get called again on another thread which won't have its `LAST_ERROR` set.

src/packbuilder.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// TODO(Nemo157): Documentation
+    pub fn insert_walk(&mut self, walk: &mut Revwalk) -> Result<(), Error> {
+        unsafe {
+            try_call!(raw::git_packbuilder_insert_walk(self.raw, walk.raw()));
+        }
+        Ok(())
+    }
+
+    /// TODO(Nemo157): Documentation
+    pub fn insert_recur(&mut self, id: Oid, name: Option<&str>) -> Result<(), Error> {
```
Nemo157: Sounds good.

Nemo157:
> I wonder if perhaps the state-based transition between packbuilder types may be a bit unergonomic for now?

Yep, that's the main reason I thought a WIP PR would be good for this :smile:. It was an interesting experiment, but you're probably right that it's overkill for this case. I'm tempted to try and write a couple of tests actually using all the types to see how ergonomic it is in use, then I'll fallback to just having a `Send` bound.

src/packbuilder.rs
alexcrichton: Hm that's actually a very good point, this doesn't catch the case where a panic happens on one thread and then it gets propagated to other threads.

In fact, that actually means that the thread local used won't propagate the panic back to the original thread I think? Maybe this should use a channel here?

Nemo157:
> Maybe we could go with just Send for now, and if necessary we can add a separate one-off builder for adding a non-Send closure which hardcodes threads to 0?

Thinking about what would actually be needed to propagate the panic correctly from any of the callbacks through the packbuilder makes me think going the other way initially might be better. Don't require `Send` and don't expose the `set_threads` method. If someone wants to put in the effort to write a multi-threaded, multi-callback panic recovery system then they can add it, I'm going to be sticking to single-threaded for my use-case anyway.

src/packbuilder.rs
```diff
@@ -0,0 +1,263 @@
...
+        unsafe {
+            try_call!(raw::git_packbuilder_foreach(self.raw, foreach_c, ptr as *mut _));
+        }
+        Ok(())
+    }
+
+    /// TODO(Nemo157): Documentation
+    pub fn set_progress_callback<'a, F>(&'a mut self, mut cb: &'a mut F) -> Result<(), Error>
+                         where F: FnMut(Stage, u32, u32) -> bool {
+        let ptr = &mut cb as *mut _;
```
alexcrichton: Hm I think this may not be sound? This passes the data payload as a stack pointer to libgit2, but that data is invalid after this function returns?

This may want to store `Box<FnMut(..)>` internally and then pass that boxed pointer down to libgit2

alexcrichton:
Looking good! I think with that bug fixed and some docs this is good to go!

Nemo157:
Sorry I took so long to get back to this. Changed the progress callback to be passed in to methods that actually will cause it to get called to fix lifetime issues with it and added documentation.

src/lib.rs
```diff
@@ -98,6 +98,7 @@ pub use message::{message_prettify, DEFAULT_COMMENT_CHAR};
 pub use note::{Note, Notes};
 pub use object::Object;
 pub use oid::Oid;
+pub use packbuilder::PackBuilder;
```
alexcrichton: This may also want to reexport `Stage` as I believe that's exported as part of the public API

Nemo157: Would it make sense to rename it to something like `PackBuilderStage` or `PackingStage` in then?

src/packbuilder.rs
```diff
@@ -0,0 +1,358 @@
...
+        unsafe {
+            try_call!(raw::git_packbuilder_write_buf(buf.raw(), self.raw));
+        }
+        if has_progress {
+            unsafe {
+                try_call!(raw::git_packbuilder_set_callbacks(self.raw,
+                                                             None,
+                                                             ptr::null_mut()));
+            }
+        }
```
alexcrichton: Unfortunately this may not be memory safe because if `git_packbuilder_write_buf` fails then the progress callback isn't unset, which means that a future call to this with `None` wouldn't reset the callbacks and could possibly use an unsafe pointer.

I wonder if perhaps an API could be added like:

```
pub fn set_progress_callback<'a: 'repo>(&mut self, progress: &'a mut ProgressCb) { 
    // ...
}
```

That way you can only set a callback which lives longer than the object itself? It'd also get the optional parameter out of these APIs and make them a little more lean

src/lib.rs
alexcrichton: Yeah PackBuilderStage sounds good to me.

src/packbuilder.rs
Nemo157: That would mean the callback has to live longer than the repository though, right? That would be annoying if say the repository was being passed by reference in to a method that wanted to create a packfile and keep track of progress information.

alexcrichton: Hm right, another alternative may be:

```rust
pub fn foo<'a, 'b: 'a>(&'a mut self, b: &'b mut Callback)
```

I'd want to verify that it's actually sound, however

Nemo157: Yeah, I don't think that's sound as it's only constraining the callback to outlive the call to `foo`, at least I was able to get code like:

```rust
let builder = ...;
{
  let cb = ...;
  builder.set_progress_callback(cb);
}
builder.write_buf(...);
```

to compile using those lifetime constraints.

Either of these are also quite unergonomic to use because the callback needs 2 layers of references to make it a slim pointer, which breaks type inference requiring a typed temporary:

```rust
pub fn foo<'a, 'b: 'a>(&'a mut self, b: &'b mut &mut Callback);

let mut cb = || { println!("called"); };
let mut cb: &mut Callback = &mut cb;
foo(&mut cb)
```

---

Alternative would be to keep the current API but explicitly null out the callback if `None` is passed, that should ensure the only time we call the callback is when it's currently owned by a parent stack frame, it's just slightly less nice as any change to this class would have to ensure that the callback is updated before calling any libgit2 function that could invoke it.

alexcrichton: Oh right, in the past the bounds I've used are:

```rust
pub fn foo<F>(&mut self, b: &'a mut F) where F: FnMut(...)
```

That way you call it as:

```rust
let mut cb = || { ... };
let obj = ...;
obj.foo(&mut cb);
```

The part about `'a` and `'b` is also kinda unfortunate, and I think there may not be a great way around that. For now, though, maybe we could remove this argument? We could always add another argument in another function in the future, or if perhaps we figure out how to pass the progress callback as a separate function.

Or is it basically necessary to expose the progress callback to land?

> Alternative would be to keep the current API but explicitly null out the callback if None is passed

I'd be somewhat worried as we'd still be storing dangling pointers, and some other function in the future may eventually start calling the progress callback.

I think instead a good solution here is to store the result of the underling raw call, unset the callbacks, and then return the original result (basically don't use `try!`

Nemo157: It's not a hard requirement for my use case, but I would much prefer to have it if possible. I just thought of something that may solve both issues; by binding the lifetime `'a` into a returned guard that removes the callback when dropped I think it should be safe, it could also allow boxing the callback which might be needed to fix the double reference type inference problem. I have to prototype it and think about all the ways it could fail, especially if the guard is forgotten instead of dropped, but I hope for a signature something like:

```rust
pub fn<'a, 'b: 'a, F>(&'a mut self, cb: &'b F) -> CallbackGuard<'a> where F: Callback<'b>
```

alexcrichton: Yeah unfortunately we can't rely on `Drop` running (as it could be leaked). Although... How about this:

1. A separate API to set the progress callback, returning an RAII guard that *owns the closure*.
2. A raw pointer in `PackBuilder` back to the RAII guard (which internally has a `Box` for a stable address)
3. A destructor in the RAII guard which unsets this raw pointer.

That way if the RAII guard is dropped, it unsets the callback, and if the RAII leaks the closure itself leaks so it's still safe to call. Thoughts?

src/packbuilder.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub fn write_buf(&mut self, buf: &mut Buf) -> Result<(), Error> {
+        unsafe {
+            try_call!(raw::git_packbuilder_write_buf(buf.raw(), self.raw));
+        }
+        Ok(())
+    }
+
+    /// Create the new pack and pass each object to the callback.
+    pub fn foreach(&mut self,
+                   mut cb: &mut ForEachCb)
```
alexcrichton: Could this be a generic function (as they're easier to call) and then switch to a monomorphic version like this privately?

src/packbuilder.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// operations, so performance may be affected.
+    pub fn set_progress_callback<F>(&mut self, progress: F) -> Result<(), Error>
+        where F: FnMut(PackBuilderStage, u32, u32) -> bool + 'repo {
+        let ptr = Box::into_raw(Box::new(Box::new(progress) as Box<ProgressCb>));
+        let progress_c = Some(progress_c as raw::git_packbuilder_progress);
+        unsafe {
+            try_call!(raw::git_packbuilder_set_callbacks(self.raw,
+                                                         progress_c,
+                                                         ptr as *mut _));
+            self.progress = Some(Box::from_raw(ptr));
```
alexcrichton: This dance of `into_raw` and `from_raw` shouldn't be necessary, right?

src/packbuilder.rs
Nemo157: I thought an RAII guard might allow limiting the lifetime of the the closure a bit, but since it can leak it doesn't really allow that as the lifetime would be a lie; since it can't do that I don't see any advantage to returning one. The only extra thing it would add is support for unsetting the progress callback, I think changing `set_progress_callback` to take in an `Option` would probably be a better way to do that.

src/packbuilder.rs
Nemo157: Definitely, I think this was leftover from an old attempt at making the progress callback work.

src/packbuilder.rs
Nemo157: I think it is, we need to pass in the raw pointer of the outer box as the callback argument and simultaneously keep a reference to it for cleanup when the `Packbuilder` is dropped. That reference in `self.progress` could be changed to a raw pointer as well and the `Box::from_raw` be called in `Packbuilder::drop` instead, this way just seemed cleaner to me.

Although, I just realised this may be breaking some aliasing rules, during a call to `extern fn progress_c` the inner box containing the callback is both simultaneously owned by the unborrowed box in the Packbuilder and referenced by a `*mut` in `progress_c`. Might be safer to store the raw pointer just to make sure there's no way this could be unsafe.

src/packbuilder.rs
Nemo157: Passing in an `Option` isn't very nice to use without being able to default the type parameter when passing `None`. I went with a separate `unset_progress_callback` method instead.

src/packbuilder.rs
```diff
@@ -0,0 +1,398 @@
...
+                    raw::git_packbuilder_set_callbacks(self.raw,
+                                                       progress_c,
+                                                       ptr as *mut _)));
+            // Current implementation will never return an error, but since it
+            // may change in the future...
+            // If an error is returned we don't know if it has kept the old
+            // pointers or the new ones. So to avoid dropping a callback that
+            // may actually get called in the future we keep the old callback
+            // stored and just leak the new one by not recreating a `Box` from
+            // the raw pointer.
```
alexcrichton: This seems like it'd be a bug in libgit2 to continue using the pointer after the call returned an error, I'd expect some form of transactional semantics where nothing happens unless it succeeds entirely. Along those lines that's why I was thinking that the extra `Box::into_raw` dance may be unnecessary?

Nemo157: That's the semantics I'd expect as well, but they don't say what the error semantics are in the header. Currently they only return an error if the packbuilder pointer is null, which we ensure never happens; but if it changes in the future for some reason I'd prefer to have the current wrapper memory safe with a tiny performance/leak penalty.

Is there any other way to get the raw pointer of a box without using `into_raw`? The only alternative I can think of would be something along the lines of

```rust
let progress = Box::new(progress) as Box<ProgressCb>;
let ptr = &mut progress as *mut _;
unsafe {
    raw::git_packbuilder_set_callbacks(self.raw, progress_c, ptr as *mut _);
    self.progress = progress;
}
```

but that would be unsafe as the pointer would be to the stack allocation of the box before it got moved into the `Packbuilder`.

alexcrichton: Ah yeah you can do something like:

```rust
let ptr = &*boxed as *const _ as *mut _;
```

I think for now let's stick to simplicity and avoid the dance of `into_raw` + `from_raw`, we can always deal with futures versions of libgit2 later if need be.

alexcrichton:
Awesome, thanks @Nemo157!

PR #124, opened 2016-03-28
Expose an API for reading raw object bytes

cgwalters:
I'm just learning Rust by adding an implementation in Rust to
https://github.com/cgwalters/git-evtag

In order to compute the checksum I need access to the raw data.
Adding the libgit2-sys API should hopefully be uncontroversial, but I
had a choice of either adding this odd high level method to
`git2::Object`, or exposing safe bindings for the ODB.  Let me know if
you prefer the latter.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -50,6 +50,8 @@ pub enum git_config {}
 pub enum git_config_iterator {}
 pub enum git_index {}
 pub enum git_object {}
+pub enum git_odb {}
+pub enum git_odb_object {}
 pub enum git_reference {}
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I thought that this was already exposed, but I guess that's just for blobs, not for arbitrary objects.

Right now most APIs in git2-rs are pretty thin shims around mostly just one API in libgit2, so perhaps this could be broken apart into a few bindings? That'd help lay the groundwork for future expansion as well!

cgwalters:
Yeah, I'll rework things to expose the ODB then.

alexcrichton:

PR #126, opened 2016-03-29
Add binding for git_repository_open_ext

joshtriplett:
This addresses part of #123.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -1,11 +1,12 @@
-use std::ffi::{CStr, CString};
+use std::env;
+use std::ffi::{CStr, CString, OsStr};
 use std::iter::IntoIterator;
 use std::mem;
 use std::path::Path;
 use std::str;
 use libc::{c_int, c_char, size_t, c_void, c_uint};
 
-use {raw, Revspec, Error, init, Object, RepositoryState, Remote, Buf};
+use {raw, Revspec, Error, init, Object, RepositoryOpenFlags, RepositoryState, Remote, Buf};
 use {ResetType, Signature, Reference, References, Submodule, Blame, BlameOptions};
 use {Branches, BranchType, Index, Config, Oid, Blob, Branch, Commit, Tree};
 use {AnnotatedCommit, MergeOptions, SubmoduleIgnore, SubmoduleStatus};
@@ -55,6 +56,43 @@ impl Repository {
         let mut ret = 0 as *mut raw::git_repository;
         unsafe {
             try_call!(raw::git_repository_open(&mut ret, path));
+            Ok(Binding::from_raw(ret))
+        }
+    }
+
+    /// Find and open an existing repository, with additional options.
+    ///
+    /// If flags contains REPOSITORY_OPEN_NO_SEARCH, the path must point
+    /// directly to a repository; otherwise, this may point to a subdirectory
+    /// of a repository, and `open_ext` will search up through parent
+    /// directories.
+    ///
+    /// If flags contains REPOSITORY_OPEN_CROSS_FS, the search through parent
+    /// directories will not cross a filesystem boundary (detected when the
+    /// stat st_dev field changes).
+    ///
+    /// If flags contains REPOSITORY_OPEN_BARE, force opening the repository as
+    /// bare even if it isn't, ignoring any working directory, and defer
+    /// loading the repository configuration for performance.
+    ///
+    /// ceiling_dirs specifies a list of paths that the search through parent
+    /// directories will stop before entering.  Use the functions in std::env
+    /// to construct or manipulate such a path list.
+    pub fn open_ext<P, O, I>(path: P,
+                             flags: RepositoryOpenFlags,
+                             ceiling_dirs: I)
+                            -> Result<Repository, Error>
+            where P: AsRef<Path>, O: AsRef<OsStr>, I: IntoIterator<Item=O> {
+        init();
+        let path = try!(path.as_ref().into_c_string());
+        let ceiling_dirs_os = try!(env::join_paths(ceiling_dirs));
+        let ceiling_dirs = try!(ceiling_dirs_os.into_c_string());
+        let mut ret = 0 as *mut raw::git_repository;
+        unsafe {
+            try_call!(raw::git_repository_open_ext(&mut ret,
+                                                   path,
+                                                   flags.bits() as c_uint,
+                                                   ceiling_dirs));
             Ok(Binding::from_raw(ret))
         }
     }
@@ -1663,6 +1701,7 @@ impl RepositoryInitOptions {
 
 #[cfg(test)]
 mod tests {
+    use std::ffi::OsStr;
     use std::fs;
     use std::path::Path;
     use tempdir::TempDir;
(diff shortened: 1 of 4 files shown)


joshtriplett:
I intentionally didn't include the version bumps for libgit2-sys or git2 in this pull request, because I think it makes sense to wait until https://github.com/libgit2/libgit2/pull/3711 and then the rest of #123 get merged, and then bump both to 0.5.0 given the new APIs.

src/repo.rs
```diff
@@ -59,6 +59,24 @@ impl Repository {
         }
     }
 
+    /// Find and open a repository, with additional options.
+    pub fn open_ext<P: AsRef<Path>>(path: P,
+                                    flags: RepositoryOpenFlags,
+                                    ceiling_dirs: Option<&str>)
```
alexcrichton: I believe that this is actually a naturally-separated list of directories, so perhaps this could be an iterator of `T: AsRef<OsStr>`? I think there's some other APIs for consuming an iterator like that elsewhere as well.

src/repo.rs
```diff
@@ -59,6 +59,24 @@ impl Repository {
         }
     }
 
+    /// Find and open a repository, with additional options.
```
alexcrichton: Because flags aren't exactly namespaced just yet, could you add some examples here of what values `flags` can take on?

alexcrichton:
That all sounds good to me, thanks @joshtriplett!

src/repo.rs
joshtriplett: There's `iter2cstrs`, but that just constructs a `git_strarray`.  `git_repository_open_ext` takes a single string of paths separated by `GIT_PATH_LIST_SEPARATOR`.

Also, if the caller already has this parameter in the form libgit2 expects, it seems unfortunate to have to split it just to have git2-rs re-join it.

I think it makes more sense for the caller to call `std::env::join_paths` themselves, rather than having `open_ext` do it for them; that way, if you already have a pre-joined path in the form you'd get it from `$GIT_CEILING_DIRECTORIES`, you can just pass that.

src/repo.rs
joshtriplett: Sure, I can add more documentation here.

(It's unfortunate that I can't just copy the documentation from the comments in libgit2, because the license isn't the same.)

joshtriplett:
I expanded the documentation for `git::Repository::open_ext`, including documentation for all the flags, documentation for `ceiling_dirs`, and a reference to `std::env` for function to construct a path list for `ceiling_dirs`.

src/repo.rs
```diff
@@ -59,6 +59,42 @@ impl Repository {
...
+    ///
+    /// If flags contains REPOSITORY_OPEN_BARE, force opening the repository as
+    /// bare even if it isn't, ignoring any working directory, and defer
+    /// loading the repository configuration for performance.
+    ///
+    /// ceiling_dirs specifies a list of paths (separated by the platform path
+    /// list separator) that the search through parent directories will stop
+    /// before entering.  Use the functions in std::env to construct or
+    /// manipulate such a path list.
+    pub fn open_ext<P: AsRef<Path>>(path: P,
```
alexcrichton: I actually tend to use the bound `IntoCString` in this repo nad that should help elide the `.as_ref()` below as well

src/repo.rs
```diff
@@ -59,6 +59,42 @@ impl Repository {
...
+                                    flags: RepositoryOpenFlags,
+                                    ceiling_dirs: Option<&str>)
```
alexcrichton: You mentioned that this may be more convenient if the caller already has this format expected by libgit2, but my counter point to that would be:

* No one actually ever has that format expected by libgit2
* If you did, it'd likely be the return value of `env::join_paths` which is an `OsString`

For the most ergonomic usage this should probably take a `I: IntoIterator<T: IntoCString>`, but I'd be fine just taking a slice for now.

src/repo.rs
joshtriplett: For the path?  Sure, I can do that.  I'd modeled this after other functions in `git::Repository`, which also took `AsRef<Path>`, but `IntoCString` seems easy enough to switch to.

src/repo.rs
joshtriplett: The other likely source of this value would be from `std::env::var`, which would be a `String`.

If you'd like, I could try to make this take a trait `IntoPathList`, and implement that trait for both `IntoCString` and iterators of `IntoCString`; that would let you pass either.  Would that work?

alexcrichton: It's not necessarily that hard to do something like `env::split_path(&val)`, though, right? That returned iterator would satisfy the requirement if we have the right generic bounds.

We can perhaps add a new trait later, but I'd prefer to stick on the side of fewer traits for now (having `IntoCString` is already unfortunate)

src/repo.rs
alexcrichton: Oh weird, I should probably change those to `IntoCString` at some point. In that case feel free to leave this :)

src/repo.rs
joshtriplett: > It's not necessarily that hard to do something like env::split_path(&val), though, right? That returned iterator would satisfy the requirement if we have the right generic bounds.

And then `open_ext` would immediately call `env::join_paths` on it; that seems like a waste. :)

Since we have functions in `std::env` to do the conversion in either direction, and since libgit2 wants the joined form, it seems like we should either accept the joined form and have callers call `env::join_paths` if they have the split form, or accept both and call `env::join_paths` ourselves if given the split form.  Personally, I think the former seems simplest: accept an IntoCString and have the caller use `env::join_paths` if needed.

src/repo.rs
joshtriplett: OK, will do.

src/repo.rs
joshtriplett: I experimented with making `ceiling_dirs` use `Option<T>` where `T: IntoCString`.  However, I found that that created an annoying usability problem for callers: `None` doesn't disambiguate the type inside the `Option`, so callers would have to specify the type of the `None`.

Because of this ambiguity, I don't think IntoCString will work well here for a type that needs to live inside an Option.  Given that, I think it makes sense to stick with accepting a slice, as the current code does.

Any other suggestions to make this interface more usable?

alexcrichton: I don't think we really need to worry about runtime work here, you're opening a repository which is surely doing I/O and is far more expensive than any allocation here. The most ergonomic interface I can think of is `T: IntoIterator<Item: IntoCString>`

joshtriplett:
OK, the version I just pushed uses an `IntoIterator`, where the `Item` is an `AsRef<OsStr>`.

I didn't manage to get an `IntoIterator` over `IntoCString` to work (specifically, the conversion to the types `join_paths` wanted).  Please feel free to generalize if you'd like.

I also find it unfortunate that, in the common case of not passing `ceiling_dirs`, I still had to say `&[] as &[&OsStr]`, because `&[]` didn't provide enough information for type inference to work.

alexcrichton:
Oh yeah `AsRef<OsStr>` is fine if we're calling `join_paths`. That is pretty unfortunate about calling this with an empty slice, but in theory that won't happen most of the time as this is just a "power user" function over one of the more ergonomic variants?

Gonna go ahead and merge for now, thanks @joshtriplett!

alexcrichton:
oops

PR #132, opened 2016-04-28
Update to rust-url 1.0

SimonSapin:
The code change is not strictly necessary but it saves a `String` allocation.

This is not a breaking change since `Url` is not part of the public API.


alexcrichton:
Neat!

SimonSapin:
Please publish a git2-curl version, to help Cargo ship only one copy of rust-url :)

alexcrichton:
Done!

PR #136, opened 2016-05-06
Please provide bindings for git_reference_ensure_log and git_reference_has_log

joshtriplett:
`git_reference_ensure_log`, in particular, looks like the only way to force the existence of a reflog for a ref outside the normal branch namespace (similar to "HEAD").


alexcrichton:
Thanks!

PR #128, opened 2016-05-06
Please support git_repository_set_index

joshtriplett:
libgit2 provides a function `git_repository_set_index` to set the index of a repository, useful for doing index-based operations without disrupting the current state in `.git/index`.  Please consider providing a `git::Repository::set_index` function to access that.


joshtriplett:
I'm not entirely sure if this is the correct type signature.  The underlying function definitely needs to mutate the index (setting the owner, specifically).  But I see many functions in git2-rs that seem to hide mutation behind a plain `&` rather than a `&mut`, including all the Repository functions.

joshtriplett:
The Travis failure here looks like some spurious issue with systest, unrelated to this commit.

alexcrichton:
Thanks!

PR #137, opened 2016-05-07
Add index flags and index extended flags

joshtriplett:
Needed for the `flags` and `extended_flags` fields of IndexEntry

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -687,6 +687,34 @@ pub struct git_blame_hunk {
 pub type git_index_matched_path_cb = extern fn(*const c_char, *const c_char,
                                                *mut c_void) -> c_int;
 
+git_enum! {
+    pub enum git_idxentry_extended_flag_t: u16 {
+        GIT_IDXENTRY_INTENT_TO_ADD     = 1 << 13,
+        GIT_IDXENTRY_SKIP_WORKTREE     = 1 << 14,
+        GIT_IDXENTRY_EXTENDED2         = 1 << 15,
+
+        GIT_IDXENTRY_UPDATE            = 1 << 0,
+        GIT_IDXENTRY_REMOVE            = 1 << 1,
+        GIT_IDXENTRY_UPTODATE          = 1 << 2,
+        GIT_IDXENTRY_ADDED             = 1 << 3,
+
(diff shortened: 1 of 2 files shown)


alexcrichton:
Hm I wonder if in light of https://github.com/alexcrichton/git2-rs/issues/138 if the entire `IndexEntry` API should get overhauled? Perhaps a bunch of boolean accessors/setters instead of a flag-based API, along with validation and whitelists to ensure we don't allow the same sort of memory corruption. What do you think?

Eventually I'd kinda like to move toward a world where we don't use `bitflags!` in git2-rs at all...

joshtriplett:
I'm not sure you'll be able to do without `bitflags!` entirely, given that some APIs accept flags directly. And I don't know if it makes sense for every data structure that contains a flags field to expand it out into accessors, if there's more than one structure containing the same set of flags (that'd take a bit of an audit).  But I do agree that `IndexEntry` needs an overhaul, and it might make sense to automatically handle things like "valid" and "extended" (the latter being set if any of `flags_extended` are set).  So in the particular case of `IndexEntry`, I think accessors would make sense, though I'd still suggest having a way to get and set the entire `flags_extended` field at once, too.

alexcrichton:
Hm could you expand a bit on why you think we'd still want the ability to set `flags_extended` all at once? One major downside I've seen with flags is that they aren't documented well as it's difficult to see what the set of flags are. They're also not necessarily as idiomatic in Rust as they are in C, so they typically feel a little awkward to use.

I'd at least prefer when reading code to see the Rust-style accessors and setters as it'd fit in well (in theory at least).

joshtriplett:
@alexcrichton I had been thinking about the case of repeatedly using the same `flags_extended` for different index entries, but it occurs to me that cloning the whole IndexEntry and changing the name/oid would probably work better for that.  So nevermind; putting all the accessors directly on the IndexEntry seems fine.

alexcrichton:
Ok, I may try to get around to just a wholly redesigned `IndexEntry` tonight and see what happens

alexcrichton:
Ok I think I'm gonna go ahead and merge this for now, fixup the memory safety issues, and we can perhaps push for a different design later. Unfortunately I'm running short on time nowadays :(

alexcrichton:
Thanks though!

PR #142, opened 2016-05-18
add support for checkout notifications

blaenk:
This introduces a `.notify()` method on `CheckoutBuilder` which makes it
possible to register a checkout notification callback.

This also introduces the `CheckoutNotificationType` type which
represents the type of checkout notification.

There are a variety of methods such as `.notify_on_conflict()` added to
`CheckoutBuilder` which allow granular toggling of which kinds of
notifications to invoke the callback on.

---

I was going through the libgit2 documentation and noticed that there was a way [[0]](https://libgit2.github.com/libgit2/#HEAD/type/git_checkout_notify_t) [[1]](https://libgit2.github.com/libgit2/#HEAD/group/callback/git_checkout_notify_cb) to get notifications about different events while doing a checkout, so I wanted to try out what that was like and expanded the bindings for it. As usual, most of this stuff already existed in libgit2-sys, so I just wired things up!

I'd love to hear any feedback. Perhaps I'm going about it entirely the wrong way. I tried it in my downstream project and it seems great :smile: 

With respect to the various methods I added to `CheckoutBuilder` for granular toggling of the notification types, I'm not sure what you would prefer. Naively I was thinking to just have something like `.notify_on(notifications: CheckoutNotificationType)` but I saw other methods on `CheckoutBuilder` that are specific, so I added one for each, e.g. `.notify_on_conflict(on_conflict: bool)`. Perhaps I misunderstood, so if you prefer a single method like I was originally going to do, let me know and I'll set that up!

With respect to registering the callback, I used the same strategy you already use with the progress callback.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -6,13 +6,6 @@ use std::path::Path;
 use libc::{c_char, size_t, c_void, c_uint, c_int};
 
 use {raw, panic, Error, Repository, FetchOptions, IntoCString, CheckoutNotificationType, DiffFile};
-use {
-    CHECKOUT_NOTIFICATION_CONFLICT,
-    CHECKOUT_NOTIFICATION_DIRTY,
-    CHECKOUT_NOTIFICATION_UPDATED,
-    CHECKOUT_NOTIFICATION_UNTRACKED,
-    CHECKOUT_NOTIFICATION_IGNORED,
-};
 use util::{self, Binding};
 
 /// A builder struct which is used to build configuration for cloning a new git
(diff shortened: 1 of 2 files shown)


blaenk:
Woops didn't notice there was a new commit since I created this. Rebased.

src/build.rs
```diff
@@ -429,6 +528,33 @@ extern fn progress_cb(path: *const c_char,
...
+        } else {
+            Some(util::bytes2path(CStr::from_ptr(path).to_bytes()))
+        };
+
+        callback(CheckoutNotificationType::from_bits_truncate(why),
+                 path,
+                 DiffFile::from_raw(baseline),
+                 DiffFile::from_raw(target),
+                 DiffFile::from_raw(workdir))
+    }).unwrap()
```
blaenk: Actually I think we shouldn't panic here, but I'm not sure. I think `panic::wrap()` only yields `None` if the contents panicked? A notification callback cancels the checkout if the callback returns a non-zero value. So I'm not sure if the contents panicking would be reason to merely cancel the checkout or if it's much more serious and grounds to panic outright.

So is the fact that `panic::wrap()`'s contents panics serious enough to propagate that panic, or should it only cancel the checkout?

If simply cancelling the checkout is fine, I'll replace this with `.unwrap_or(1)`, unless there's something else that would be preferable.

blaenk:
Just a heads up, it seems like d5d91799e0 which switches to using curl-rust's rewrite branch is causing the tests to fail. It seems like curl-rust's rewrite branch is trying to import unstable names.

src/build.rs
alexcrichton: Yeah let's stick to `.unwrap_or(1)` to cancel the checkout. It also has the added bonus of avoiding undefined behavior (unwinding into C)

src/build.rs
```diff
@@ -295,6 +312,70 @@ impl<'cb> CheckoutBuilder<'cb> {
...
+        self.notify_flag(CHECKOUT_NOTIFICATION_UNTRACKED, on_untracked)
+    }
+
+    /// Indicate whether a notification callback should be invoked for
+    /// notifications of ignored files.
+    ///
+    /// Defaults to false.
+    pub fn notify_on_ignored(&mut self, on_ignored: bool) -> &mut CheckoutBuilder<'cb> {
+        self.notify_flag(CHECKOUT_NOTIFICATION_IGNORED, on_ignored)
+    }
```
alexcrichton: Perhaps these functions could be replaced with one which just accepts `CheckoutNotificationType`?

src/build.rs
```diff
@@ -40,6 +49,12 @@ pub struct CheckoutBuilder<'cb> {
 /// of completed steps so far, and the final is the total number of steps.
 pub type Progress<'a> = FnMut(Option<&Path>, usize, usize) + 'a;
 
+/// Checkout notifications callback.
+///
+/// The first argument is the notification type, the next is the path for the
+/// the notification, followed by the baseline diff, target diff, and workdir diff.
+pub type Notify<'a> = FnMut(CheckoutNotificationType, Option<&Path>, DiffFile, DiffFile, DiffFile) -> c_int + 'a;
```
alexcrichton: Could this return a `bool` to be a bit more "rust-y"? We could translate that to zero/nonzero under the hood

alexcrichton:
Thanks!

src/build.rs
blaenk: Sure will do! Oh and yeah I completely forgot that that was the entire point of `panic::wrap()` :sweat_smile:

src/build.rs
blaenk: Yeah! That was what I was going to do, as I mentioned in my post, but I thought perhaps since there were other functions for specific flags that you wanted to do the same here. I'm going to create one called `.notify_on(t: CheckoutNotificationType)`.

src/build.rs
blaenk: Oh yeah good catch, I completely missed that.

blaenk: It definitely shouldn't have been `c_int`, but I had left it as an integer before because the the docs say:

> Returning a non-zero value from this callback will cancel the checkout. The non-zero return value will be propagated back and returned by the git_checkout_... call.

So the number they returned could be retrieved from the result of the checkout function. Getting information out like that seems like a C-ism to me, as far as libgit2 is concerned all it cares about is whether it's non-zero or not, if it is then it cancels the checkout. Since the callback is FnMut, users could get information out through the environment captures, so I agree that it'll be simpler to return a bool.

So instead of returning 0 the callback will return false, which will read as "don't cancel the checkout", whereas true would have the opposite effect.

blaenk:
Alright I made those changes. When you get a chance feel free to let me know if there's anything else you'd like changed.

alexcrichton:
Ah oops yeah should have read more closely! So in general I definitely prefer lots of flags over bitflags (I'd love to drop the dep eventually...), but in this case we're already handing all the bitflags back anyway in the callback so I figured if we have to have an abstraction for that we may as well just take it as part of the builder anyway.

alexcrichton:
Thanks!

alexcrichton:
Ah and I'm actually gonna invert the return value of the callback so `true` indicates "keep going" as I think that's being used in a few other places in this lib as well

blaenk:
Yeah I agree with that change.

blaenk:
Note that I specified what the return value means in the documentation for the `Notify` type as well as the `notify()` method, so that should be changed as well. Would you like me to make those changes?

alexcrichton:
Ah no worries I'm already working on this as-is so I went ahead and made the change

PR #143, opened 2016-05-18
use regular doc comments

blaenk:
So I just learned from the previous PR that you actually want to eventually drop the bitflags dependency, so in light of that I split the previously-single commit into two commits in case you'd like to at least merge one of them. However, feel free to close and disregard this if neither one is desired!

The first switches bitflags documentation over to regular doc comments `///`, because the previous `#[doc="something"]` doc trick for macros is thankfully no longer necessary. However, I wonder if you kept them despite knowing this on purpose, perhaps to support older rustc versions or something. If so, disregard!

The second commit bumps the bitflags requirement to "0.7". Newer bitflags requires an explicit `pub` specifier to expose the constants, so that work is also done.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,9 +344,7 @@ pub enum FileFavor {
 }
 
 bitflags! {
-    #[doc = "
-Orderings that may be specified for Revwalk iteration.
-"]
+    /// Orderings that may be specified for Revwalk iteration.
     flags Sort: u32 {
         /// Sort the repository contents in no particular ordering.
         ///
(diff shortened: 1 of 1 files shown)


blaenk:
Hmmm... I don't think the doc format commit 0855fc9 came out correctly, I'll fix it.

alexcrichton:
I'm a little wary to upgrade bitflags just yet as it's likely to be a breaking change (as it tweaks the API here and there), but once a new major release of git2-rs is made I can do the bump. Could the docs be fixed up with 0.1 though?

blaenk:
Ah okay no problem!

Yeah I believe it'll work with 0.1, as far as I know it's a rustc change that removed the need for `#[doc]`, so all macros benefit.

However, I'm in the process of fixing up the docs change commit, since I didn't split it up correctly. I'll let you know when it's ready.

blaenk:
Alright the branch (now a misnomer) only includes the change to regular doc comments.

I did generate the docs and made sure that they're working fine. Feel free to close this though or postpone it if you like, it's no problem!

blaenk:
Just now rebased over master, hadn't noticed that there were new commits.

alexcrichton:
Thanks!

PR #145, opened 2016-05-21
allow passing RemoteCallbacks to Remote.connect()

blaenk:
Otherwise it's impossible to connect to remotes that require authentication.

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -96,11 +96,14 @@ impl<'repo> Remote<'repo> {
     }
 
     /// Open a connection to a remote.
-    pub fn connect(&mut self, dir: Direction) -> Result<(), Error> {
+    pub fn connect(&mut self, dir: Direction,
+                   callbacks: Option<&mut RemoteCallbacks>)
+                   -> Result<(), Error> {
         // TODO: can callbacks be exposed safely?
+        let cbs = callbacks.map(|cb| cb.raw());
         unsafe {
             try_call!(raw::git_remote_connect(self.raw, dir,
-                                              0 as *const _,
+                                              cbs.as_ref(),
                                               0 as *const _));
         }
         Ok(())


alexcrichton:
I thought I remembered trying this awhile ago and concluding that it wasn't safe just yet, but I could be wrong! It seemed though that libgit2 was storing the callbacks across this function call and calling them later, which means we'd have to have some sort of lifetime in effect to capture that I believe

blaenk:
Ah that's horrible. Without this it's not possible to push to a even connect to a github remote via git/ssh protocol for example, but indeed I hadn't noticed that libgit2 is saving these callbacks behind the scenes. It's possible

here's `git_remote_connect` in remote.c:

``` c
	if ((error = set_transport_callbacks(t, callbacks)) < 0 ||
	    (error = t->connect(t, url, credentials, payload, direction, flags)) != 0)
		goto on_error;
```

and `set_callbacks()`:

``` c
int set_transport_callbacks(git_transport *t, const git_remote_callbacks *cbs)
{
	if (!t->set_callbacks || !cbs)
		return 0;

	return t->set_callbacks(t, cbs->sideband_progress, NULL,
				cbs->certificate_check, cbs->payload);
}
```

Finally the `set_callbacks` function for the smart transport:

``` c
static int git_smart__set_callbacks(
	git_transport *transport,
	git_transport_message_cb progress_cb,
	git_transport_message_cb error_cb,
	git_transport_certificate_check_cb certificate_check_cb,
	void *message_cb_payload)
{
	transport_smart *t = (transport_smart *)transport;

	t->progress_cb = progress_cb;
	t->error_cb = error_cb;
	t->certificate_check_cb = certificate_check_cb;
	t->message_cb_payload = message_cb_payload;

	return 0;
}
```

So it looks like it's saving the progress, error, and certificate check callbacks, as well as the payload.

I guess it makes sense that it saves things so that they can be used when doing other operations on the repo.

Could this possibly be resolved using lifetime bounds, like saying something like:

``` rust
pub fn connect<'cb: 'a>(&'a mut self, dir: Direction, callbacks: Option<&mut RemoteCallbacks<'cb>>)
```

Or maybe I'm doing it wrong or it's not even possible, but some way to specify that the `Remote` must not outlive the `RemoteCallbacks`, because then that would mean that the pointers libgit2 saved are now dangling? Actually I'm getting myself mixed up I think. I'm just wondering if lifetime bounds may help in any way here. Or perhaps they would be too restrictive.

What are some alternatives? Are there any?

blaenk:
Ohh!! I just realized that the `push` methods etc. which do take (indirectly) a `RemoteCallbacks` already connects if it wasn't already connected. I think it's not too much of a problem then if this isn't handled for `connect()` yet.

I'll report back!

blaenk:
Yeah, I'll close this. The `fetch()` and `push()` methods work well!

alexcrichton:
Ah yeah and I *think* it's safe there as they shouldn't persist beyond that call, but I could be wrong...

I'm not actually 100% sure what the use case is for `connect` vs fetch/push, but maybe there's an opportunity for more optimization or perhaps getting certain errors early?

blaenk:
Yeah I imagine it's that as well as keeping an open connection to the remote. I haven't investigated but I imagine the push/fetch calls do everything in one go, i.e. they open a new connection, download/upload, update tips, etc. So if one were to do many operations it may be beneficial to connect once to avoid incurring the setup/tear down cost each time. That's just a guess.

In my case just using fetch and push suffices so it's not a problem for me.

I was originally using `connect()` because that's what the [git2-rs/examples/fetch.rs](https://github.com/alexcrichton/git2-rs/blob/master/examples/fetch.rs#L77-L86) uses. It connects manually and then performs a manual `download()`. The problem I was having was that it's currently not possible to authenticate when connecting manually since it doesn't take a `RemoteCallbacks`.

alexcrichton:
Yeah that is kinda unfortunate... I wonder if it's possible to finagle this to get it to work either way though...

PR #148, opened 2016-05-24
Updated build instructions for OSX

### LINKED ISSUE #147: Build instructions didn't work for me
The instructions for building on OSX didn't work for me. I'm on OSX 10.11.5 if that helps anyone reproduce:

```
% brew --version
Homebrew 0.9.9 (git revision d345; last commit 2016-05-22)
Homebrew/homebrew-core (git revision 994a3; last commit 2016-05-23)
```

The fix is simple enough, just using

``` sh
export OPENSSL_ROOT_DIR=`brew --prefix openssl`
```

worked well in place of the two exports in the current instructions. The upstream source of your instructions still has the other commands so maybe it's just my system.

xaocon:
The instructions given for building from homebrew on OSX didn't work for me. I found that the OPENSSL_ROOT_DIR env variable was needed instead. I left the old commands in case they are needed on other systems as they didn't affect the fix in my testing.

Resolves #147

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,13 @@ been removed, but if you're using Homebrew you can install them via:
 brew install openssl
 ```
 
-To get this library to pick them up the [standard `rust-openssl`
-instructions][instr] can be used to transitively inform libssh2-sys about where
-the header files are:
-
-[instr]: https://github.com/sfackler/rust-openssl#osx
+To get this library to pick them up the brew prefix command can be used to
+transitively inform libssh2-sys about where the header files are:
 
 ```sh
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm so testing out locally, is this required for you? I don't have this env var set and it seems to work?

xaocon:
I'm not able to reproduce the problem on my home system. It works fine without the change. It may be some oddity of my dev system.

alexcrichton:
Hm ok, wanna poke around there when you get a chance and maybe we can dig into what's going on?

xaocon:
I don't seem to be having the problems anymore. I don't know why but I'm closing.

PR #149, opened 2016-06-08
Please provide binding for git_config_parse_bool

joshtriplett:
To make it easier to handle git config values that can take either a bool or a string, I'd like to have a binding for git_config_parse_bool.  That would make it easier to get a value as a string, check if it's a valid bool, and handle it as a string otherwise.

I would suggest the type `fn config_parse_bool<S: IntoCString>(s: S) -> Result<bool, Error>`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -349,6 +349,43 @@ impl Config {
             Ok(Binding::from_raw(ret))
         }
     }
+
+    /// Parse a string as a bool.
+    /// Interprets "true", "yes", "on", 1, or any non-zero number as true.
+    /// Interprets "false", "no", "off", 0, or an empty string as false.
+    pub fn parse_bool<S: IntoCString>(s: S) -> Result<bool, Error> {
+        let s = try!(s.into_c_string());
+        let mut out = 0;
+        ::init();
+        unsafe {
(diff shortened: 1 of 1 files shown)


joshtriplett:
Attached a pull request adding `config_parse_bool`, along with `config_parse_i32`, `config_parse_i64`, and tests.

alexcrichton:
Thanks! I wonder if these may perhaps be better placed as `Config::parse_bool` or something like that?

joshtriplett:
As static methods that don't take self or return Self?  I can do that if you prefer.  They don't have anything to do with creating or manipulating a Config, but this change wouldn't impact usability at all, so I don't care what color the bikeshed is painted. :)

Let me know your preference and I'll make the change if needed.

joshtriplett:
I've pushed a version of this as static functions to [my config_parse_static](https://github.com/joshtriplett/git2-rs/tree/config_parse_static) branch; let me know and I can force-push that to this pull request instead.

alexcrichton:
Ah yeah that's what I was thinking, just generally one less thing to import hopefully if you're already working with `Config` :)

alexcrichton:
Thanks!

PR #150, opened 2016-06-08
Implement PartialEq for git2::Error (and thus Result<T, git2::Error>)

joshtriplett:
This allows comparing Error values using == and !=, which in turn also
allows comparing Result<T, Error> for any T that implements PartialEq.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,7 +8,7 @@ use libc::c_int;
 use {raw, ErrorClass, ErrorCode};
 
 /// A structure to represent errors coming out of libgit2.
-#[derive(Debug)]
+#[derive(Debug,PartialEq)]
 pub struct Error {
     code: c_int,
     klass: c_int,


alexcrichton:
I wonder if `#[derive]` is perhaps the wrong thing to do here? The same code/class are probably enough to identify an error that may have different descriptions perhaps?

joshtriplett:
Makes sense; given that no programmatic usage should rely on the content of the description strings, they shouldn't have semantic significance for comparison.  I'll update the branch.

joshtriplett:
Looking into this further, I think deriving PartialEq actually does make sense.  `Error::from_str` always produces errors with the same code and klass, so only comparing code and klass would make all errors produced with `Error::from_str` compare equal.

Since you can't actually construct an Error object with an arbitrary code and klass (without using the unsafe `from_raw` method), I don't think people will expect to use equality to match a git2::Error by just code and klass.  If Rust had a concept of "object identity" (similar to Python `is`), that would suffice here.

For simplicity, I'd propose continuing to use `derive`.  If `git2::Error` moved to using an enum or similar to track errors produced by git2-rs itself (or chained from some other Rust error) separately from errors produced by libgit2, then it might make sense to have the equality for the errors from libgit2 compare only code and klass.

I tried a local hack like this:
```rust
impl PartialEq<Error> for Error {
    fn eq(&self, other: &Error) -> bool {
        // For errors produced using from_str, compare the strings; otherwise
        // only compare the code and klass.
        (self.code == other.code && self.klass == other.klass
         && !(self.code == raw::GIT_ERROR && self.klass == (raw::GITERR_NONE as c_int) && self.message != other.message))
    }
}
```

But that doesn't feel very satisfying at all.

alexcrichton:
Ok, that sounds reasonable to me, thanks!

PR #151, opened 2016-06-13
Fix compilation of libgit2-sys on Illumos/Solaris

nbaksalyar:
This patch fixes the following compilation error on Illumos/Solaris OSes:
```
git2-rs/libgit2-sys/lib.rs:8:1: 8:37 error: can't find crate for `openssl` [E0463]
git2-rs/libgit2-sys/lib.rs:8 extern crate openssl_sys as openssl;
```


alexcrichton:
Thanks!

PR #123, opened 2016-07-02
Please support discovering repositories using the standard git environment variables

joshtriplett:
`git::Repository::discover` passes specific default arguments to the underlying `git_repository_discover` function, which prevents respecting `$GIT_DISCOVERY_ACROSS_FILESYSTEM` or `$GIT_CEILING_DIRECTORIES`.  Please consider providing a binding that allows passing these values explicitly (as a bool and a &str), and please consider making the default binding respect the default values.

In addition, please also consider providing a function `discover_default` or similar, which doesn't take a path parameter, and which defaults to `$GIT_DIR` (or "." if that doesn't exist).

Also see https://github.com/libgit2/libgit2/issues/3711 in libgit2, which requests a new underlying C API for that case.  However, it should be easy for git2-rs to support this with or without the underlying function.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -75,6 +75,13 @@ impl Repository {
     /// bare even if it isn't, ignoring any working directory, and defer
     /// loading the repository configuration for performance.
     ///
+    /// If flags contains REPOSITORY_OPEN_NO_DOTGIT, don't try appending
+    /// `/.git` to `path`.
+    ///
+    /// If flags contains REPOSITORY_OPEN_FROM_ENV, `open_ext` will ignore
+    /// other flags and `ceiling_dirs`, and respect the same environment
+    /// variables git does. Note, however, that `path` overrides `$GIT_DIR`.
+    ///
     /// ceiling_dirs specifies a list of paths that the search through parent
     /// directories will stop before entering.  Use the functions in std::env
     /// to construct or manipulate such a path list.
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -60,6 +60,23 @@ impl Repository {
         }
     }
 
+    /// Find and open an existing repository, respecting git environment
+    /// variables.  This acts like `open_ext` with the
+    /// `REPOSITORY_OPEN_FROM_ENV` flag, but additionally respects `$GIT_DIR`.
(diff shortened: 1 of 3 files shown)


joshtriplett:
I just turned this into a pull request, adding the two new flags and the new `Repository::open_from_env` function.

joshtriplett:
It looks like updating libgit2-sys to use current libgit2 from Git breaks several tests in the test suite, unrelated to this change.  That'll need fixing before this can go in.

src/repo.rs
```diff
@@ -60,6 +60,23 @@ impl Repository {
         }
     }
 
+    /// Find and open an existing repository, respecting git environment
+    /// variables.  This acts like `open_ext` with the
+    /// `REPOSITORY_OPEN_FROM_ENV` flag, but additionally respects `$GIT_DIR`.
+    /// With `$GIT_DIR` unset, this will search for a repository starting in
+    /// the current directory.
```
alexcrichton: Could this elaborate a bit on what "respecting git environment variables" means in the documentation as well? It's not entirely obvious to me at first glance.

alexcrichton:
Looks good to me, thanks!

This may have to do with the fact that some of the FFI definitions have broken as evidenced by the systest suite I believe (bottom of the travis logs). If it gets too hard trying to fix I can also help out as well to get this landed.

src/repo.rs
joshtriplett: Sure, I can elaborate a bit.  I'd rather not duplicate the entire libgit2 documentation for this, though.

(Not least of which because git2-rs uses a different license than libgit2, so I can't just copy/paste from libgit2.)

alexcrichton: Note that quite a few docs are copy/pasted from libgit2 (I think it's good to be exhaustive here as not everyone will jump to libgit2 docs), but if there's license shenanigans that should likely be mentioned in the README or something like that.

joshtriplett:
Updated for current libgit2.

alexcrichton:
Oh dear sorry I forgot about this!

joshtriplett:
Not a problem; this PR needed updating before merging anyway. Thanks for doing the work to update the version of libgit2!

PR #160, opened 2016-10-03
Update libgit2

alexcrichton:
This'll probably require a major version bump of both the -sys and wrapper crate


alexcrichton:
Currently motivated by https://github.com/rust-lang/cargo/issues/3042#issuecomment-251159702 as this may help Cargo segfaulting on FreeBSD

PR #165, opened 2016-10-31
allow empty git credential_helper

nrxus:
previously, if .gitconfig had a credential_helper field that was empty
the add_command function would crash due to trying to slice the empty
command string from [1..].

Signed-off-by: Alpha Chen <achen@pivotal.io>
Signed-off-by: Andres Medina <amedina@pivotal.io>


alexcrichton:
Thanks!

PR #166, opened 2016-11-01
Add bindings for the git_patch type and API

joshtriplett:
Add a new Patch type wrapping `git_patch`, along with bindings for all the `git_patch_*` functions.

In particular, this now exposes the functionality to diff blobs and buffers.


libgit2-sys/lib.rs
```diff
@@ -2488,6 +2489,58 @@ extern {
...
+                                opts: *const git_diff_options) -> c_int;
+    pub fn git_patch_from_blob_and_buffer(out: *mut *mut git_patch,
+                                          old_blob: *const git_blob,
+                                          old_as_path: *const c_char,
+                                          buffer: *const c_char,
+                                          buffer_len: size_t,
+                                          buffer_as_path: *const c_char,
+                                          opts: *const git_diff_options) -> c_int;
+    pub fn git_patch_from_buffers(out: *mut *mut git_patch,
+                                  old_buffer: *const c_char,
```
alexcrichton: The systest checkers are failing and it looks like this argument is `c_void` I think?

src/patch.rs
```diff
@@ -0,0 +1,203 @@
...
+        let mut ret = 0 as *mut raw::git_patch;
+        unsafe {
+            try_call!(raw::git_patch_from_diff(&mut ret, diff.raw(), idx));
+            Ok(Binding::from_raw_opt(ret))
+        }
+    }
+
+    /// Generate a Patch by diffing two blobs.
+    pub fn from_blobs<S1, S2>(old_blob: &Blob,
+                              old_path: Option<S1>,
```
alexcrichton: Perhaps these could be `Option<&Path>` to help with type inference when `None` is passed?

src/patch.rs
```diff
@@ -0,0 +1,203 @@
...
+        let mut ret = 0 as *const raw::git_diff_hunk;
+        let mut lines = 0;
+        unsafe {
+            try_call!(raw::git_patch_get_hunk(&mut ret, &mut lines, self.raw, hunk_idx));
+            Ok((Binding::from_raw(ret), lines))
+        }
+    }
+
+    /// Get the number of lines in a hunk.
+    pub fn num_lines_in_hunk(&self, hunk_idx: usize) -> Result<i32, Error> {
```
alexcrichton: Perhaps we should return a `u32` here? A negative number of lines seems like it'd be interesting...

libgit2-sys/lib.rs
joshtriplett: Apparently; it looks like for that function one buffer uses `void *` and the other uses `char *`.  Strange.

Reported as https://github.com/libgit2/libgit2/issues/3986

src/patch.rs
joshtriplett: Done.

src/patch.rs
joshtriplett: I'd done it this way because the underlying function would break for values more than int.  (It should have taken a `size_t` out parameter and returned an int as an error code only.)  But I'll switch it to `usize`.

PR #171, opened 2016-11-22
Switch from whitelist to blacklist in detecting OpenSSL path

tbu-:
This enables support for OpenBSD and NetBSD (and possibly
DragonFlyBSD?).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -548,8 +548,6 @@ fn openssl_init() {
 
     openssl_sys::init();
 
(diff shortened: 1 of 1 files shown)


tbu-:
I'm not sure why there was an additional whitelist in there. What OS was it supposed to exclude? Some obscure OS is probably bound to be forgotten on that list.

Should I re-add it and put NetBSD and FreeBSD on that list instead?

alexcrichton:
Hm to be honest I'm not sure why it's there either. I should really update this logic and only do it if OpenSSL is linked statically...

In any case seems fine to me!

PR #173, opened 2016-11-25
fix wrong name break_rewries to break_rewrites

silvio:
Signed-off-by: Silvio Fricke <silvio.fricke@gmail.com>


alexcrichton:
Thanks!

PR #177, opened 2016-12-03
Bindings for Stash

s-panferov:
Hi. Thanks for the amazing library. This is my first attempt to write something using Rust FFI (and I've never written anything in pure C), so I'll appreciate any help and guidance.

This PR is not ready yet. I created it just to make everyone know that I'm working on it.

1. [x] Impl everything from https://github.com/libgit2/libgit2/blob/master/include/git2/stash.h and make it compile.
2. [x] Write tests and make it work.
3. [x] Format and lint the code.
4. [x] Make CI tests pass.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -6,7 +6,7 @@ use std::path::Path;
 use std::str;
 use libc::{c_int, c_char, size_t, c_void, c_uint};
 
-use {raw, Revspec, Error, init, Object, RepositoryOpenFlags, RepositoryState, Remote, Buf};
+use {raw, Revspec, Error, init, Object, RepositoryOpenFlags, RepositoryState, Remote, Buf, StashFlags};
 use {ResetType, Signature, Reference, References, Submodule, Blame, BlameOptions};
 use {Branches, BranchType, Index, Config, Oid, Blob, Branch, Commit, Tree};
 use {AnnotatedCommit, MergeOptions, SubmoduleIgnore, SubmoduleStatus};
@@ -14,6 +14,7 @@ use {ObjectType, Tag, Note, Notes, StatusOptions, Statuses, Status, Revwalk};
 use {RevparseMode, RepositoryInitMode, Reflog, IntoCString, Describe};
 use {DescribeOptions, TreeBuilder, Diff, DiffOptions, PackBuilder};
 use build::{RepoBuilder, CheckoutBuilder};
+use stash::{StashApplyOptions, StashCb, StashCbData, stash_cb};
 use string_array::StringArray;
 use oid_array::OidArray;
 use util::{self, Binding};
@@ -1712,6 +1713,88 @@ impl Repository {
         }
     }
 
+    /// Save the local modifications to a new stash.
+    pub fn stash_save(&mut self,
+                stasher: &Signature,
+                message: &str,
+                flags: Option<StashFlags>)
+                -> Result<Oid, Error> {
+        unsafe {
+            let mut raw_oid = raw::git_oid { id: [0; raw::GIT_OID_RAWSZ] };
+            let message = try!(CString::new(message));
+            let flags = flags.unwrap_or(StashFlags::empty());
+            try_call!(raw::git_stash_save(
+                &mut raw_oid,
+                self.raw(),
+                stasher.raw(),
+                message,
+                flags.bits() as c_uint
+            ));
+
+            Ok(Binding::from_raw(&raw_oid as *const _))
+        }
+    }
+
+    /// Apply a single stashed state from the stash list.
+    pub fn stash_apply(&mut self,
+                 index: usize,
+                 opts: Option<&mut StashApplyOptions>)
+                 -> Result<(), Error> {
+        unsafe {
+            let opts = opts.map(|opts| opts.raw());
+            try_call!(raw::git_stash_apply(
+                self.raw(),
+                index,
+                opts
+            ));
+
+            Ok(())
+        }
+    }
+
+    /// Loop over all the stashed states and issue a callback for each one.
+    ///
+    /// Return `true` to continue iterating or `false` to stop.
+    pub fn stash_foreach<C>(&mut self, callback: C) -> Result<(), Error>
+        where C: FnMut(usize, &str, &Oid) -> bool
+    {
+        unsafe {
+            let mut data = Box::new(StashCbData { callback: Box::new(callback) as Box<StashCb> });
+            try_call!(raw::git_stash_foreach(
+                self.raw(),
+                stash_cb,
+                &mut *data as *mut _ as *mut _
+            ));
+            Ok(())
+        }
+    }
+
+    /// Remove a single stashed state from the stash list.
+    pub fn stash_drop(&mut self, index: usize) -> Result<(), Error> {
+        unsafe {
+            try_call!(raw::git_stash_drop(
+                self.raw(),
+                index
+            ));
+            Ok(())
+        }
+    }
+
+    /// Apply a single stashed state from the stash list and remove it from the list if successful.
+    pub fn stash_pop(&mut self,
+               index: usize,
+               opts: Option<&mut StashApplyOptions>)
+               -> Result<(), Error> {
+        unsafe {
+            let opts = opts.map(|opts| opts.raw());
+            try_call!(raw::git_stash_pop(
+                self.raw(),
+                index,
+                opts
+            ));
+            Ok(())
+        }
+    }
 }
 
 impl Binding for Repository {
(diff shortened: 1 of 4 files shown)


s-panferov:
@alexcrichton this PR is ready for review.

src/stash.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /// Set stash application flag
+    pub fn flags(&mut self, flags: raw::git_stash_apply_flags) -> &mut StashApplyOptions<'cb> {
+        self.raw_opts.flags = flags;
+        self
+    }
+
+    /// Options to use when writing files to the working directory
+    pub fn checkout_options(&mut self, mut opts: CheckoutBuilder) -> &mut StashApplyOptions<'cb> {
```
alexcrichton: I think this may need to be stored internally to persist the callbacks stored inside of `opts`

src/stash.rs
```diff
@@ -0,0 +1,287 @@
...
+            raw_opts: unsafe { mem::zeroed() },
+        };
+        assert_eq!(unsafe {
+            raw::git_stash_apply_init_options(&mut opts.raw_opts, 1)
+        }, 0);
+        opts
+    }
+
+    /// Set stash application flag
+    pub fn flags(&mut self, flags: raw::git_stash_apply_flags) -> &mut StashApplyOptions<'cb> {
```
alexcrichton: Could we add a method-per-flag to avoid exposing libgit2-sys types?

src/stash.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+/// A set of stash methods
+pub struct Stash {
+
+}
+
+impl Stash {
+    /// Save the local modifications to a new stash.
+    pub fn save(repo: &mut Repository,
```
alexcrichton: I've typically preferred to add these sorts of methods onto `Repository` itself, so we'd have something like:

```rust
repo.stash_save(...);
repo.stash_apply(...);
// etc
```

src/stash.rs
```diff
@@ -0,0 +1,287 @@
...
+            ));
+
+            Ok(())
+        }
+    }
+
+    /// Loop over all the stashed states and issue a callback for each one.
+    ///
+    /// Return `true` to continue iterating or `false` to stop.
+    pub fn foreach(repo: &mut Repository, callback: Box<StashCb>) -> Result<(), Error> {
```
alexcrichton: Could this be made generic as well? (like the one above)

src/stash.rs
s-panferov: Not quite understand what do you mean here. You want to introduce another Rust enum with conversion to `git_stash_apply_flags`?

src/stash.rs
```diff
@@ -0,0 +1,204 @@
...
+            let mut callback = options.progress.as_mut().unwrap();
+            callback(convert_progress(progress))
+        };
+
+        if res { 0 } else { -1 }
+    }).unwrap()
+}
+
+#[cfg(test)]
+mod tests {
```
s-panferov: @alexcrichton should I move tests to `repo.rs` as well? Or it's better to leave them here?

s-panferov:
@alexcrichton done.

src/stash.rs
alexcrichton: Ah yeah this is fine to leave here.

src/repo.rs
```diff
@@ -1712,6 +1713,88 @@ impl Repository {
...
+    }
+
+    /// Loop over all the stashed states and issue a callback for each one.
+    ///
+    /// Return `true` to continue iterating or `false` to stop.
+    pub fn stash_foreach<C>(&mut self, callback: C) -> Result<(), Error>
+        where C: FnMut(usize, &str, &Oid) -> bool
+    {
+        unsafe {
+            let mut data = Box::new(StashCbData { callback: Box::new(callback) as Box<StashCb> });
```
alexcrichton: I think the `Box` can be elided here, right? We could just use a `&mut FnMut(...)`?

src/repo.rs
```diff
@@ -1712,6 +1713,88 @@ impl Repository {
...
+                flags.bits() as c_uint
+            ));
+
+            Ok(Binding::from_raw(&raw_oid as *const _))
+        }
+    }
+
+    /// Apply a single stashed state from the stash list.
+    pub fn stash_apply(&mut self,
+                 index: usize,
```
alexcrichton: For now could this follow the style of the other functions here in terms of alignment of arguments?

src/repo.rs
```diff
@@ -1712,6 +1713,88 @@ impl Repository {
...
+                 opts: Option<&mut StashApplyOptions>)
+                 -> Result<(), Error> {
+        unsafe {
+            let opts = opts.map(|opts| opts.raw());
+            try_call!(raw::git_stash_apply(
```
alexcrichton: Similarly with the style above, could these function calls follow the same style found elsewhere in this file?

src/stash.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub fn reinstantiate_index(&mut self) -> &mut StashApplyOptions<'cb> {
+        self.raw_opts.flags = raw::GIT_STASH_APPLY_REINSTATE_INDEX;
+        self
+    }
+
+    /// Options to use when writing files to the working directory
+    pub fn checkout_options(&mut self, mut opts: CheckoutBuilder<'cb>) -> &mut StashApplyOptions<'cb> {
+        unsafe {
+            opts.configure(&mut self.raw_opts.checkout_options);
+        }
```
alexcrichton: I think we'll want to defer this until `raw` is called because otherwise movement of the structs may mess up some internal pointers and such.

src/stash.rs
```diff
@@ -0,0 +1,204 @@
...
+            let mut callback = &mut data.callback;
+            callback(
+                index,
+                CStr::from_ptr(message).to_str().unwrap(),
+                &Binding::from_raw(stash_id)
+            )
+        };
+
+        if res { 0 } else { 1 }
+    }).unwrap()
```
alexcrichton: I think this `unwrap` will likely want to be `unwrap_or(1)`

src/repo.rs
s-panferov: @alexcrichton I need some help with this. How can I unwrap `c_void` into &mut FnMut(...)`?

```
pub extern fn stash_cb(index: size_t,
                       message: *const c_char,
                       stash_id: *const raw::git_oid,
                       payload: *mut c_void)
                       -> c_int
{
    panic::wrap(|| unsafe {
        let mut callback = &mut *(payload as *mut StashCb);
```

```
error[E0277]: the trait bound `for<'r, 'r> libc::c_void: std::ops::FnMut<(usize, &'r str, &'r oid::Oid)>` is not satisfied
  --> src/stash.rs:84:35
   |
84 |         let mut callback = &mut *(payload as *mut StashCb);
   |                                   ^^^^^^^ trait `for<'r, 'r> libc::c_void: std::ops::FnMut<(usize, &'r str, &'r oid::Oid)>` not satisfied
   |
   = note: required for the cast to the object type `for<'r, 'r> std::ops::FnMut(usize, &'r str, &'r oid::Oid) -> bool`
```

s-panferov:
@alexcrichton done

src/stash.rs
```diff
@@ -0,0 +1,204 @@
...
+{
+    panic::wrap(|| unsafe {
+        let mut options = &mut *(payload as *mut StashApplyOptions);
+        let res = {
+            let mut callback = options.progress.as_mut().unwrap();
+            callback(convert_progress(progress))
+        };
+
+        if res { 0 } else { -1 }
+    }).unwrap()
```
alexcrichton: Oh like above this'll want to be `unwrap_or`

src/repo.rs
```diff
@@ -1712,6 +1713,71 @@ impl Repository {
...
+    }
+
+    /// Loop over all the stashed states and issue a callback for each one.
+    ///
+    /// Return `true` to continue iterating or `false` to stop.
+    pub fn stash_foreach<C>(&mut self, mut callback: C) -> Result<(), Error>
+        where C: FnMut(usize, &str, &Oid) -> bool
+    {
+        unsafe {
+            let mut data = Box::new(StashCbData { callback: &mut callback });
```
alexcrichton: I believe the `Box` can be elided here and you can just pass a stack pointer, right?

aaronschif:
Thanks to both of you for this library. I was building a little project with this branch and was surprised to find `stash_foreach` took a mutable reference. First time writing with this language, but perhaps this is a bug?

s-panferov:
@aaronschif this is not a const pointer. https://github.com/libgit2/libgit2/blob/master/include/git2/stash.h#L219

So potentially `libgit2` may change something internally.

s-panferov:
@alexcrichton Thanks for the advises! I've learned a lot from this PR. The latest notes are fixed.

alexcrichton:
Looks great to me, thanks again for the PR @s-panferov!

PR #183, opened 2017-01-24
feat: add cherrypick bindings

### LINKED ISSUE #181: Cherrypick bindings?
I'm interested in using `git2` to make a CLI tool for my team to automate some repetitive Git work we have to do. One of the main things we do a lot is cherrypicking a given revision to a bunch of different branches. I noticed that the `libgit2-sys` crate has some constants related to cherrypicking, but no bindings for the cherrypick functions themselves.

I'd love to add these, but I'm not 100% sure how to integrate them into the library. Is there a good module that I could look at for examples, or can someone provide some guidance? Thanks!

aergonaut:
Fixes #181

Adds bindings to `git_cherrypick` and `git_cherrypick_init_options`.

Allows cherrypicking a given `Commit` through a `Repository` with the new function `cherrypick`.

### DIFF
--- a/src/cherrypick.rs
+++ b/src/cherrypick.rs
@@ -1,0 +1,26 @@
+use std::mem;
+
+use raw;
+
+/// Options to specify when cherry picking
+pub struct CherrypickOptions {
+    raw_opts: raw::git_cherrypick_options
+}
+
+impl CherrypickOptions {
+    /// Creates a default set of cherrypick options
+    pub fn new() -> CherrypickOptions {
+        let mut opts = CherrypickOptions {
+            raw_opts: unsafe { mem::zeroed() }
+        };
+        assert_eq!(unsafe {
+            raw::git_cherrypick_init_options(&mut opts.raw_opts, 1)
+        }, 0);
+        opts
+    }
+
+    /// Acquire a pointer to the underlying raw options
+    pub unsafe fn raw(&self) -> *const raw::git_cherrypick_options {
+        &self.raw_opts as *const _
+    }
+}
--- a/src/cherrypick.rs
+++ b/src/cherrypick.rs
@@ -1,26 +1,59 @@
 use std::mem;
 
+use build::CheckoutBuilder;
+use merge::MergeOptions;
 use raw;
+use std::ptr;
 
 /// Options to specify when cherry picking
-pub struct CherrypickOptions {
-    raw_opts: raw::git_cherrypick_options
+pub struct CherrypickOptions<'cb> {
+    mainline: u32,
+    checkout_builder: Option<CheckoutBuilder<'cb>>,
+    merge_opts: Option<MergeOptions>
 }
 
-impl CherrypickOptions {
+impl<'cb> CherrypickOptions<'cb> {
     /// Creates a default set of cherrypick options
-    pub fn new() -> CherrypickOptions {
-        let mut opts = CherrypickOptions {
-            raw_opts: unsafe { mem::zeroed() }
+    pub fn new() -> CherrypickOptions<'cb> {
+        let opts = CherrypickOptions {
+            mainline: 1,
+            checkout_builder: None,
+            merge_opts: None
         };
-        assert_eq!(unsafe {
-            raw::git_cherrypick_init_options(&mut opts.raw_opts, 1)
-        }, 0);
         opts
     }
 
-    /// Acquire a pointer to the underlying raw options
-    pub unsafe fn raw(&self) -> *const raw::git_cherrypick_options {
-        &self.raw_opts as *const _
+    /// Set the checkout builder
+    pub fn checkout_builder(&mut self, cb: CheckoutBuilder<'cb>) -> &mut Self {
+        self.checkout_builder = Some(cb);
+        self
+    }
+
+    /// Set the merge options
+    pub fn merge_opts(&mut self, merge_opts: MergeOptions) -> &mut Self {
+        self.merge_opts = Some(merge_opts);
+        self
+    }
+
+    /// Obtain the raw pointer
+    pub fn raw(&mut self) -> *const raw::git_cherrypick_options {
+        unsafe {
+            let mut checkout_opts: raw::git_checkout_options = mem::zeroed();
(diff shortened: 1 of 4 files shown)


src/repo.rs
```diff
@@ -1778,6 +1779,18 @@ impl Repository {
             Ok(())
         }
     }
+
+    /// Perform a cherrypick
+    pub fn cherrypick(&self, commit: &Commit, options: Option<CherrypickOptions>) -> Result<(), Error> {
```
alexcrichton: Typically for these sorts of options the library takes `Option<&mut Options>` to allow re-using them over time.

alexcrichton:
Thank! I wonder if the options could contain the other sets of options as well so they can all get configured?

aergonaut:
@alexcrichton I updated the `cherrypick` function to use `Option<&mut CherrpickOptions>` and did the defaulting like in the `statuses` function.

I based on the options included on the options struct from the libgit2 docs, which only mention the merge and checkout opts. Which others do you think should be included?

alexcrichton:
`cherrypick` looks good, thanks! For the options I figured you could have something like [`PushOptions`](https://docs.rs/git2/0.6.3/git2/struct.PushOptions.html) where it has the ability to wholesale set merge/checkout options through [`MergeOptions`](https://docs.rs/git2/0.6.3/git2/struct.MergeOptions.html) and [`CheckoutBuilder`](https://docs.rs/git2/0.6.3/git2/build/struct.CheckoutBuilder.html).

aergonaut:
Sorry I kind of let this PR drop off. I'm still interested in getting this working but I am a little stuck!

I'm trying to work out how to let CherrypickOptions take MergeOptions and CheckoutBuilder like you suggested. The trouble I'm having is with converting the CherrypickOptions into the raw git_cherrypick_options, because MergeOptions' [`raw`](https://docs.rs/git2/0.6.3/git2/struct.MergeOptions.html#method.raw) returns a pointer rather than the raw git_merge_options directly. So I'm getting mismatched types when I try to construct the raw git_cherrypick_options.

Here's the implementation of `raw` I have currently:

```rust
pub fn raw(&mut self) -> *const raw::git_cherrypick_options {
    unsafe {
        let mut checkout_opts: raw::git_checkout_options = mem::zeroed();
        raw::git_checkout_init_options(&mut checkout_opts, raw::GIT_CHECKOUT_OPTIONS_VERSION);
        if let Some(ref mut cb) = self.checkout_builder {
            cb.configure(&mut checkout_opts);
        }
        let opts = raw::git_cherrypick_options {
            version: 1,
            mainline: self.mainline,
            checkout_opts: checkout_opts,
            merge_opts: self.merge_opts.as_ref().map(|m| m.raw()).unwrap_or(MergeOptions::new().raw())
        };
        &opts as * const _
    }
}
```

I'm having trouble with this line:

```rust
self.merge_opts.as_ref().map(|m| m.raw()).unwrap_or(MergeOptions::new().raw())
```

If I leave it like it is, the types are mismatched because I'm getting a pointer instead of a struct. But if I try to dereference the pointer, the compiler complains that I cannot move out of dereference of a raw pointer.

So I'm a little stuck on how to make this work. I took a look at how PushOptions works, but I think my problem is that MergeOptions' `raw` gives a pointer instead of the struct, whereas the elements in PushOptions all return structs when calling `raw`.

alexcrichton:
Ah yeah you may need to add a new method to configure an existing libgit2 merge options or perhaps return a struct instead of returning a pointer. Would that help here?

rylio:
@aergonaut This is what I did for the RebaseOptions bindings. I use `ptr::copy` to put it in a new struct.

```rust
pub unsafe fn raw(&mut self) -> Result<raw::git_rebase_options, Error> {
        let mut checkout_options: raw::git_checkout_options = mem::zeroed();
        try_call!(raw::git_checkout_init_options(&mut checkout_options,
                                                 raw::GIT_CHECKOUT_OPTIONS_VERSION));
        if let Some(ref mut opts) = self.checkout_options {
            opts.configure(&mut checkout_options);
        }
        let mut merge_options: raw::git_merge_options = mem::zeroed();
        if let Some(ref opts) = self.merge_options {
            ptr::copy(opts.raw(), &mut merge_options, 1);
        }

        Ok(raw::git_rebase_options {
               version: self.version as c_uint,
               quiet: self.quiet as c_int,
               inmemory: self.in_memory as c_int,
               rewrite_notes_ref: ::call::convert(&self.rewrite_notes_ref),
               merge_options: merge_options,
               checkout_options: checkout_options,
           })
    }
```

aergonaut:
@rylio Thanks! I'll try something like that!

src/cherrypick.rs
```diff
@@ -0,0 +1,59 @@
...
+            if let Some(ref opts) = self.merge_opts {
+                ptr::copy(opts.raw(), &mut merge_opts, 1);
+            }
+            let opts = raw::git_cherrypick_options {
+                version: 1,
+                mainline: self.mainline,
+                checkout_opts: checkout_opts,
+                merge_opts: merge_opts
+            };
+            &opts as * const _
```
alexcrichton: Ah unfortunately I think this'll be use-after-free because it's returning a pointer to the stack :(

alexcrichton: Could this struct be returned by value though?

src/cherrypick.rs
```diff
@@ -0,0 +1,59 @@
...
+    mainline: u32,
+    checkout_builder: Option<CheckoutBuilder<'cb>>,
+    merge_opts: Option<MergeOptions>
+}
+
+impl<'cb> CherrypickOptions<'cb> {
+    /// Creates a default set of cherrypick options
+    pub fn new() -> CherrypickOptions<'cb> {
+        let opts = CherrypickOptions {
+            mainline: 1,
```
alexcrichton: Could you add a builder method to configure this as well?

src/cherrypick.rs
aergonaut: Yeah, I'll take a look

src/cherrypick.rs
aergonaut: Sure thing

alexcrichton:

vthib:
I have resubmitted this PR with the review comments taken into account and an additional test: #432 . Could you take a look?

PR #186, opened 2017-02-01
Add RemoteConnection struct for Remote::connect

lawliet89:
This is a continuation of #182.

I had some difficulty implementing this without making breaking changes to `Remote`.

I had to store the raw `*mut raw::git_remote` pointer instead of `&mut Remote` because you won't be able to do anything with `Remote` as long as the connection is live otherwise. Ideally, this connection struct should be inside `Remote` like you mentioned.

For the test, for some reason, the `sideband_progress` is not actually run during the `fetch` but it's actually used somewhere. So it will cause a segfault if the callback is freed.  If you run `cargo test` in 	40bc55f it will segfault.

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -57,8 +57,7 @@ pub struct RemoteConnection<'repo, 'connection, 'cb> where 'repo: 'connection {
     callbacks_raw: raw::git_remote_callbacks,
     proxy: ProxyOptions<'cb>,
     proxy_raw: raw::git_proxy_options,
-    remote_raw: *mut raw::git_remote,
-    _marker: marker::PhantomData<&'connection Remote<'repo>>,
+    remote: &'connection mut Remote<'repo>,
 }
 
 impl<'repo> Remote<'repo> {
@@ -111,7 +110,7 @@ impl<'repo> Remote<'repo> {
     /// Open a connection to a remote.
     ///
     /// Returns a `RemoteConnection` that will disconnect once dropped
-    pub fn connect<'connection, 'cb>(&mut self,
+    pub fn connect<'connection, 'cb>(&'connection mut self,
                                      dir: Direction,
                                      cb: Option<RemoteCallbacks<'cb>>,
                                      proxy_options: Option<ProxyOptions<'cb>>)
@@ -133,8 +132,7 @@ impl<'repo> Remote<'repo> {
             callbacks_raw: cb_raw,
             proxy: proxy_options,
             proxy_raw: proxy_raw,
-            remote_raw: self.raw.clone(),
-            _marker: marker::PhantomData
+            remote: self,
         })
     }
 
@@ -145,11 +143,7 @@ impl<'repo> Remote<'repo> {
 
     /// Disconnect from the remote
     pub fn disconnect(&mut self) {
-        Remote::disconnect_raw(self.raw)
-    }
-
-    fn disconnect_raw(remote_raw: *mut raw::git_remote) {
-        unsafe { raw::git_remote_disconnect(remote_raw) }
+        unsafe { raw::git_remote_disconnect(self.raw) }
     }
 
     /// Download and index the packfile
@@ -469,9 +463,24 @@ impl<'cb> Binding for PushOptions<'cb> {
     }
 }
 
+impl<'repo, 'connection, 'cb> RemoteConnection<'repo, 'connection, 'cb> {
+    /// Check whether the remote is (still) connected
+    pub fn connected(&mut self) -> bool {
+        self.remote.connected()
+    }
+
+    /// Get the remote repository's reference advertisement list.
+    ///
+    /// This list is available as soon as the connection to
+    /// the remote is initiated and it remains available after disconnecting.
+    pub fn list(&self) -> Result<&[RemoteHead], Error> {
+        self.remote.list()
+    }
+}
+
 impl<'repo, 'connection, 'cb> Drop for RemoteConnection<'repo, 'connection, 'cb> {
     fn drop(&mut self) {
-        Remote::disconnect_raw(self.remote_raw)
+        self.remote.disconnect()
     }
 }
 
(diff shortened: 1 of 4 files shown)


lawliet89:
Hang on -- there's a segfault even with this. Trying to figure it out.

alexcrichton:
Looking good to me, thanks! It'd be great of the returned value could store a literal `Remote`, but did you find that the lifetimes didn't work out?

Also, could you describe a bit more what the segfault was you saw? From the "fix" commit it's not clear to me how it fixed it...

lawliet89:
> Looking good to me, thanks! It'd be great of the returned value could store a literal Remote, but did you find that the lifetimes didn't work out?

It wasn't really the lifetime that was the issue. I would have to store a mutable reference to the `Remote` to disconnect from it, and once I do that, I can't do things like `list` until `RemoteConnection` is dropped.

> Also, could you describe a bit more what the segfault was you saw? From the "fix" commit it's not clear to me how it fixed it...

The error message was literally `SIGSEGV: invalid memory reference` with no further information.

The fix was simple. The structs from calling `raw()` on the callbacks and proxy options didn't live long enough. They were dropped at the end of `connect`.  So there was a free after use. I just stored them in the `RemoteConnection` struct and returned it.

alexcrichton:
Is libgit2 storing the callback structures themselves? Or pointers into them? 

I think we'll want to thread through the `&mut Remote` to prevent usage of it anyway (to prevent accidental corruption) and then allow limited access to the `Remote` through the `RemoteConnection` structure.

lawliet89:
> Is libgit2 storing the callback structures themselves? Or pointers into them?

Pointers.

> I think we'll want to thread through the &mut Remote to prevent usage of it anyway (to prevent accidental corruption) and then allow limited access to the Remote through the RemoteConnection structure.

OK. Only `list` requires a manual connection. So the only function `RemoteConnection` would need to provide is `list`.

alexcrichton:
Oh so if libgit2 is storing pointers into the callbacks we're passing in, shouldn't they be moved to the heap before we call libgit2? That way we'd ensure a stable address for them?

lawliet89:
I could `Box` them up, but this is not currently done for all the other callbacks.

I'm not too familiar with how the compiler handles objects on the stack, but as far as I can tell, the address was stable when I printed out the raw pointer address.

alexcrichton:
Yeah stack addresses change all the time, so if libgit2 is storing pointers we need to box up before we pass down. Can you gist the location of where libgit2 is storing the pointers as well?

lawliet89:
I'll just do a quick trace of the code to show you where some of the pointers are saved. The code is quite a pain to read.

From [`git_remote_connect`](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L651), it [passes](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L696) the remote callbacks to [`set_transport_callback`](https://github.com/libgit2/libgit2/blob/master/src/remote.c#L634) which calls `t -> set_callbacks` where `t` is a pointer to some `git_transport`.

`git_transport` is either some user provided custom transport, or the built in [_smart_transport_](https://github.com/libgit2/libgit2/blob/master/src/transports/smart.c) or [_local_transport_](https://github.com/libgit2/libgit2/blob/master/src/transports/smart.c).

If we look at `smart_transport`, `t-> set_callbacks` maps to [`git_smart__set_callbacks`](https://github.com/libgit2/libgit2/blob/master/src/transports/smart.c#L53). This is where it assigns the pointers to the callbacks to itself.

src/remote.rs
```diff
@@ -50,6 +50,14 @@ pub struct PushOptions<'cb> {
     pb_parallelism: u32,
 }
 
+/// Holds callbacks for a connection to a `Remote`. Disconnects when dropped
+#[allow(dead_code)]
```
alexcrichton: How come this annotation is needed?

alexcrichton:
Oh those aren't storing pointers into `git_remote_callbacks` itself, though. I think that's just a temporary structure. That's storing pointers stored inside the remote callbacks, though, which should be fine as they're all boxed up in what we store.

alexcrichton:
Hm so the boxes here shouldn't be necessary, and may be hiding a segfault if the ended up fixing something...

alexcrichton:
Oh ok so it's `RemoteCallbacks` that needs a stable address because of [this line](https://github.com/alexcrichton/git2-rs/blob/afb3a4e67a42e9fe42479a98e702f8693bb72ef6/src/remote_callbacks.rs#L187), but `ProxyOptions` shouldn't need a box.

src/remote.rs
lawliet89: The `callbacks` and `proxy` fields are assigned to hold the callbacks but never used.

```
warning: field is never used: `callbacks`, #[warn(dead_code)] on by default
  --> src/remote.rs:55:5
   |
55 |     callbacks: Box<RemoteCallbacks<'cb>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: field is never used: `proxy`, #[warn(dead_code)] on by default
  --> src/remote.rs:56:5
   |
56 |     proxy: Box<ProxyOptions<'cb>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

alexcrichton: Oh you should be able to solve that by prefixing the field names with an underscore (e.g. `_proxy` and `_callbacks`)

lawliet89:
Right I think I understand it better now. [`proxy_options`](https://github.com/alexcrichton/git2-rs/blob/master/src/proxy_options.rs#L55) has no payload.

src/remote.rs
lawliet89: Good point -- I forgot about that.

lawliet89:
@alexcrichton The test passes though. Not sure why one of the builds on Travis [timed out](https://travis-ci.org/alexcrichton/git2-rs/jobs/199458306).

alexcrichton:
Ok looks great, thanks! As one final change, could the name of this method change to `connect_auth` perhaps to avoid a breaking change? That way we can delay the major version bump until a future date.

lawliet89:
Done. I have also restored some of the old test lines that exercised `connect`.

alexcrichton:
Thanks!

PR #200, opened 2017-04-27
libgit2: update to ee89941fa2b861332defb5e3a8ce12c23c496ed7

### LINKED ISSUE #178: libgit2-sys: Fix compatibility with libressl
Updating libgit2 to a tree where https://github.com/libgit2/libgit2/commit/567b83de5434abcc40eeeabb8167979240f27a8f is available should do the trick (backporting just this commit fixes the issue here).

Maybe issue a release afterwards?

Keruspe:
Fixes build with libressl
Fixes #178


alexcrichton:
Hm is the submodule commit one that isn't upstream?

Keruspe:
Heh, I actually did a cherry-pick as I thought you suggested, but as it actually creates a copy of the commit it's likely it doesn't work indeed.
The merge point of the two branches seems to be ee89941fa2b861332defb5e3a8ce12c23c496ed7
Will update to this one instead.

alexcrichton:
Thanks!

PR #210, opened 2017-05-23
Added fetch_refspecs and push_refspecs bindings to remote

rylio:
Not sure if `StringArray` is the best way to return the `refspecs`.


alexcrichton:
Thanks!

PR #215, opened 2017-05-29
Added rebase bindings

### LINKED ISSUE #146: Please provide bindings for the rebase API
See https://libgit2.github.com/libgit2/#HEAD/search/rebase .

Since the rebase API uses annotated commits, it would also help to have bindings to the other functions that create annotated commits, including [`git_annotated_commit_lookup`](https://libgit2.github.com/libgit2/#HEAD/group/annotated/git_annotated_commit_lookup) and [`git_annotated_commit_from_revspec`](https://libgit2.github.com/libgit2/#HEAD/group/annotated/git_annotated_commit_from_revspec).

rylio:
Closes #146. Probably could use some tests.


src/rebase.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub fn inmemory_index(&self) -> Result<Index, Error> {
+        let mut ret = 0 as *mut raw::git_index;
+        unsafe {
+            try_call!(raw::git_rebase_inmemory_index(&mut ret, self.raw));
+            Ok(Binding::from_raw(ret))
+        }
+    }
+
+    /// Performs the next rebase operation and returns the `RebaseOperation` about it.
+    pub fn next(&self) -> Result<RebaseOperation, Error> {
```
alexcrichton: I wonder if this may be best exposed as an `Iterator` structure perhaps? Maybe a manual iterator over a rebase entry?

src/rebase.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /// Options to control how files are written during a `rebase`.
+    pub fn checkout_builder(&mut self, checkout_builder: Option<CheckoutBuilder<'a>>) -> &mut Self {
+        self.checkout_builder = checkout_builder;
+        self
+    }
+
+    /// raw value of options
+    pub unsafe fn raw(&mut self) -> raw::git_rebase_options {
```
alexcrichton: I tend to prefer to keep these methods private if we can, could `pub` be elided here?

src/rebase.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /// The commit ID being cherry-picked.  This will be populated for
+    /// all operations except those of type `Exec`
+    pub fn id(&self) -> Oid {
+        unsafe { Binding::from_raw(&(*self.raw).id as *const raw::git_oid) }
+    }
+    /// The executable the user has requested be run.  This will only
+    /// be populated for operations of type `Exec`
+    pub fn exec(&self) -> Option<CString> {
```
alexcrichton: Typically this crate tries to expose `exec_bytes` and `exec` as non-allocating views into the bytes contained here, could that be done here as well?

src/rebase.rs
rylio: I don't think it can, it needs to be exposed to the repo module. `MergeOptions` also exposes the raw method. `#[doc(hidden)]` could be added, then when `pub(crate)` becomes stable that could be used.

alexcrichton:
Looking great! Could the fields be made private as well? Otherwise just a CI error left and I think this is good to go!

kamalmarhubi:
@alexcrichton any objections to me rebasing this and getting it landed as a new PR?

I'm basing this on the assumption that @rylio has gotten busy, which is plausible based on recent github activity. @rylio feel free to correct me! :-)

alexcrichton:
@kamalmarhubi nah feel free!

alexcrichton:

PR #218, opened 2017-06-07
libgit2-sys: update libgit2 to fix build with libressl

### LINKED ISSUE #212: `libgit2-sys` Version `0.6.11` fails to build with LibreSSL
This seems to be caused by https://github.com/libgit2/libgit2/issues/4220 - upstream is preparing a new release: https://github.com/libgit2/libgit2/issues/4227. Version `0.6.10` is known to work. I'm not sure we can do anything about it *here*, but I think it should be at least pointed out.

Keruspe:
Fixes #212

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -1348,7 +1348,7 @@ impl Repository {
                                               self.raw,
                                               their_heads.as_mut_ptr() as *mut _,
                                               their_heads.len()));
-            Ok((MergeAnalysis::from_bits_truncate(raw_merge_analysis), MergePreference::from_bits_truncate(raw_merge_preference)))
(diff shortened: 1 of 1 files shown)


Keruspe:
Second commit is there for a build failure with MSVC on appveyor, with a conflict between u32 and i32 for those types

alexcrichton:
Thanks! I personally prefer to keep the `raw` types out of the public API if possible, could the bitflags stay with a `u32` behind them?

Keruspe:
Yes, but then we hit this failure: https://ci.appveyor.com/project/alexcrichton/git2-rs/build/1.0.903

alexcrichton:
That just requires an extra `as u32` though?

Keruspe:
Yeah, right, thoguht of it backwards, fixing

Keruspe:
Note that this is needed to be able to build nightly (and probably the new beta) of rustc

alexcrichton:
Looks good to me, thanks!

Issue #220, opened 2017-06-17
Expose git_hash_*


Cyclic3:
As git is going to move to a different hash function some time soonish, would it be possible for you to expose the libgit2 hash.h functionality with libgit2_sys? (https://github.com/libgit2/libgit2/blob/master/src/hash.h)

This would allow libraries that need to hash things manually to be able to migrate with a simple recompile, rather than having to go through their code and change all of the third party references. It would also save a bit of space, as devs would no longer need to add an extra dependency to do the hashing.

Cyclic3:
How's that?

Cyclic3:
Closed until libgit2 changes

PR #221, opened 2017-06-17
Added git_hash_* declarations

Cyclic3:
Except for git_hash_vec, which requires git_buf_vec, which is not defined or used by the rest of the crate

#220


alexcrichton:
Thanks! Did you want to add high level bindings to this as well?

Cyclic3:
How's that?

alexcrichton:
Oh so I just realized, this isn't actually part of the public API of libgit2? The CI is failing because the function is undefined, and that's because this function is only defined in an internal header. I don't think it's expected that downstream consumers use this function?

Cyclic3:
The low level functions do not work (the ones that use git_hash_ctx), but git_wrap_buf does. I have compiled and used this, and I hate to be the guy who says `this works for me`, but I have seen no problem with it.

alexcrichton:
Unfortunately this crate only exposes the public API of libgit2, so if we want to expose this then you'd want to first convince libgit2 maintainers to expose it upstream.

Cyclic3:
Ok, I will talk to them. I'll close this issue for now, and reopen it if the libgit2 devs accept it.

BTW, why can I interface with it on my machine and Travis can't?

PR #224, opened 2017-06-24
Added ability to update a submodule

stephaneyfx:
This wraps `git_submodule_update`.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -22,6 +22,7 @@ pub const GIT_REMOTE_CALLBACKS_VERSION: c_uint = 1;
 pub const GIT_STATUS_OPTIONS_VERSION: c_uint = 1;
 pub const GIT_BLAME_OPTIONS_VERSION: c_uint = 1;
 pub const GIT_PROXY_OPTIONS_VERSION: c_uint = 1;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks great, thanks!

stephaneyfx:
That was fast, thank you!

Do you think implementing `Repository::clone_recurse` to clone and update all submodules would make sense (basically `git clone --recursive`)?

alexcrichton:
Yeah seems reasonable to me!

stephaneyfx:
Thank you. I will look into it.

stephaneyfx:
After looking into it more, I think adding an equivalent of `git clone --recursive` requires more thought and design, especially to do it in parallel like the `--jobs` option. This would need new callback types that are `Send`. Even without making it parallel, some questions don't have an obvious answer. For example, `CheckoutBuilder` has the ability to specify the directory to check out to and if `clone_recurse` takes a single instance of it, it is not clear to me how this instance can be used to check out multiple submodules.

In summary, this may need more thought to come up with a versatile and elegant design and I have other priorities at the moment. Hopefully, having `git_submodule_update` already helps, combined with the existing submodule functionality.

alexcrichton:
Ok no worries!

PR #226, opened 2017-06-30
Added function to clone a repository recursively

stephaneyfx:
It is similar to `git clone --recursive`. The current implementation is very
simple and limited as it does not allow to customize the fetch and checkout
tasks and submodules are processed sequentially (i.e. no option like `--jobs`
to process them concurrently).

Addressing the limitations described above requires some thought:

- `FetchOptions` and `CheckoutBuilder` are not `Send`.
- Should different submodules have different options?
- Are new options and callbacks needed?

I ended up creating repositories with recursive submodules and thought a simple implementation was better than none for now, especially for build scripts. `RepoBuilder` can be extended in the future to allow customizing the recursive cloning operation.

**Note**: One improvement I may be able to do now to this PR is updating submodules concurrently. I would need to add a dependency on rayon or crossbeam to git2. Just let me know if you want me to do it.

Thank you.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -183,6 +183,39 @@ impl Repository {
         RepoBuilder::new().clone(url, into.as_ref())
     }
 
+    /// Clone a remote repository, initialize and update its submodules
+    /// recursively.
+    ///
+    /// This is similar to `git clone --recursive`.
+    pub fn clone_recurse<P: AsRef<Path>>(url: &str, into: P)
+                                         -> Result<Repository, Error> {
+        let repo = Repository::clone(url, into)?;
+        repo.update_submodules()?;
+        Ok(repo)
+    }
+
+    /// Update submodules recursively.
+    ///
+    /// Uninitialized submodules will be initialized.
+    pub fn update_submodules(&self) -> Result<(), Error> {
+
+        fn add_subrepos(repo: &Repository, list: &mut Vec<Repository>)
+                        -> Result<(), Error> {
+            for mut subm in repo.submodules()? {
+                subm.update(true, None)?;
+                list.push(subm.open()?);
+            }
+            Ok(())
+        }
+
+        let mut repos = Vec::new();
+        add_subrepos(self, &mut repos)?;
+        while let Some(repo) = repos.pop() {
+            add_subrepos(&repo, &mut repos)?;
+        }
+        Ok(())
+    }
+
     /// Execute a rev-parse operation against the `spec` listed.
     ///
     /// The resulting revision specification is returned, or an error is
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Out of curiosity, is it safe to do this concurrently? I'd imagine that may muck with internal libgit2 state that wouldn't be safe to happen concurrently.

Also mind keeping the `update_submodules` method private for now?

stephaneyfx:
To do it concurrently, the idea would be to get `Repository` objects from `Submodule` objects using [`git_submodule_repo_init`](https://libgit2.github.com/libgit2/#HEAD/group/submodule/git_submodule_repo_init) and process these repositories concurrently. My understanding is that separate `Repository` objects are safe to use concurrently ([libgit2 thread safety](https://github.com/libgit2/libgit2/blob/master/THREADING.md)).

The reason I made `update_submodules` public is that it's actually useful in build scripts. Let's say you have a sys crate with a submodule for the C lib and the build script is in charge of updating the submodule. In this case, when the build script runs, the repository has already been cloned and the script just needs to update the submodules of this repo (similarly to the libgit2-sys crate whose build script updates its submodule). The build script could just get the `Submodule` object and update it (assuming there's only one direct submodule), but the repo the submodule points to may itself contain submodules, now or in the future, and the build script should not need to know. I am working on a crate that has just this scenario: Rust crate repo has C lib repo as a submodule and the C lib repo has a submodule of its own.

Please let me know if the purpose is still unclear. How do you want me to proceed?

1. Given the above explanations, would you agree that it is useful to have `update_submodules` public too? If not, do you have another design in mind for the use case I described?
2. Do you want me to look into updating submodules concurrently? It helps in the case of more than one submodule because fetching is done in parallel.

Thank you.

stephaneyfx:
I'm not sure if it's clear from my previous message: `Repository::update_submodules` is roughly equivalent to `git submodule update --init --recursive`, which I feel is useful to update all the submodules of an already-cloned repository recursively, for example in build scripts that update and build a foreign submodule (which could have its own submodules).

alexcrichton:
Thanks for the explanation! Makes sense to me. For now I'd prefer to keep this conservative as it doesn't expose all the flavors of options for how to update submodules. I figured if we added parallel support (which I'm sort of hesitant to do) we'd want to expose all the options instead of hardcoding the fetch options. Having a helper though as a constructor on `Repository` seems fine to me!

stephaneyfx:
I made `update_submodules` private as requested. Let me know if there's anything else you would like me to do.

alexcrichton:
Thanks!

Issue #238, opened 2017-08-12
Creating blobs from stream


katyo:
As I see the blobs can be created using files in filesystem and also binary data in memory. The libgit2 provides API to create blobs using streaming (https://libgit2.github.com/libgit2/#HEAD/group/blob/git_blob_create_fromstream).
Are bindings supports streaming feature? I would like to piping uploaded data directly to blobs without additional buffering and storing on filesystem.

alexcrichton:
Ah I dont think these are bound yet, but no reason they couldn't be!

katyo:
Also the libgit2 provides another way of writing blobs in streaming manner by using ODB API (https://libgit2.github.com/libgit2/#HEAD/group/odb/git_odb_open_wstream) but in this case we must know the length of stream before opening.

katyo:
I tried to implement blobs writer. I newbie in Rust (but have experience in C and Haskell) so I maybe not fully understand some specific things. Also I not found how correctly abort stream in libgit2 documentation so I implemented commit only, but the Drop trait might be implemented too.

katyo:
I looked in libgit2 sources and found that `free` callback of `git_writestream` is all we need to proper cleaning up uncommitted blob writing stream. So I also implemented `Drop` trait for `BlobWriter` now to complete solving this issue.

katyo:
I implemented another way of streamed blob writing (and reading) using git [ODB](https://libgit2.github.com/libgit2/#HEAD/group/odb) APIs.

alexcrichton:
I'm gonna close this since I think some new APIs have been bound and otherwise we're not currently tracking unbound APIs in this issue tracker.

PR #239, opened 2017-08-15
Added basic support of streamed blobs writing

katyo:
Fix for #238

### DIFF
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -72,12 +72,15 @@ impl<'repo> Drop for Blob<'repo> {
 /// A structure to represent a git writestream for blobs
 pub struct BlobWriter<'repo> {
     raw: *mut raw::git_writestream,
+    need_cleanup: bool,
     _marker: marker::PhantomData<Object<'repo>>,
 }
 
 impl<'repo> BlobWriter<'repo> {
     /// Finalize blob writing stream and write the blob to the object db
     pub fn commit(&mut self) -> Result<Oid, Error> {
+        // After commit we already doesn't need cleanup on drop
+        self.need_cleanup = false;
         let mut raw = raw::git_oid { id: [0; raw::GIT_OID_RAWSZ] };
         unsafe {
             try_call!(raw::git_blob_create_fromstream_commit(&mut raw, self.raw));
@@ -92,10 +95,20 @@ impl<'repo> Binding for BlobWriter<'repo> {
     unsafe fn from_raw(raw: *mut raw::git_writestream) -> BlobWriter<'repo> {
         BlobWriter {
             raw: raw,
+            need_cleanup: true,
             _marker: marker::PhantomData,
         }
     }
     fn raw(&self) -> *mut raw::git_writestream { self.raw }
+}
+
+impl<'repo> Drop for BlobWriter<'repo> {
+    fn drop(&mut self) {
+        // We need cleanup in case the stream has not been committed
+        if self.need_cleanup {
+            unsafe { ((*self.raw).free)(self.raw) }
+        }
+    }
 }
 
 impl<'repo> io::Write for BlobWriter<'repo> {
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -78,7 +78,7 @@ pub struct BlobWriter<'repo> {
 
 impl<'repo> BlobWriter<'repo> {
     /// Finalize blob writing stream and write the blob to the object db
-    pub fn commit(&mut self) -> Result<Oid, Error> {
+    pub fn commit(mut self) -> Result<Oid, Error> {
         // After commit we already doesn't need cleanup on drop
         self.need_cleanup = false;
         let mut raw = raw::git_oid { id: [0; raw::GIT_OID_RAWSZ] };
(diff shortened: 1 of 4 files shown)


src/repo.rs
```diff
@@ -802,6 +802,19 @@ impl Repository {
         }
     }
 
+    /// Create stream to write blob
```
alexcrichton: Could this copy some of the documentation from libgit2, including recommendations about what else to call and what sort of buffering this does?

alexcrichton: (also an explanation of what `hintpath` is)

src/blob.rs
```diff
@@ -68,10 +69,67 @@ impl<'repo> Drop for Blob<'repo> {
...
+}
+
+impl<'repo> io::Write for BlobWriter<'repo> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        unsafe {
+            let res = ((*self.raw).write)(self.raw, buf.as_ptr() as *const i8, buf.len());
+            if res < 0 {
+                Err(io::Error::new(io::ErrorKind::Other, "Write error"))
+            } else {
+                Ok(buf.len())
```
alexcrichton: I think `res` is the number of bytes transferred? In that the whole buffer may not be guaranteed to get written.

src/repo.rs
katyo: Okay

src/blob.rs
katyo: It's a common practice in C but the libgit2 do not follows to it.
I don't found info about `write` callback in library documentation, so I looked into sources. I found that the `write` callback calls `git_filebuf_write` ([src](https://github.com/libgit2/libgit2/blob/master/src/blob.c#L301-L306)) and `git_filebuf_write` calls its own `write` function by pointer. The implementations of `write` function returns 0 on success (for example [write_normal](https://github.com/libgit2/libgit2/blob/1560b5808e71af170d3a0c09f35cab7e973df5a5/src/filebuf.c#L148)).

src/repo.rs
```diff
@@ -802,6 +802,19 @@ impl Repository {
...
+    pub fn blob_writer(&self, hintpath: Option<&Path>) -> Result<BlobWriter, Error> {
+        let path = match hintpath {
+            Some(path) => try!(path.into_c_string()).into_raw(),
```
katyo: Can I use `.as_ptr` instead of `.into_raw` here to avoid leakage?

alexcrichton: Yes, but you'll want to create the `CString`, store it on the stack, then call `as_ptr` later

katyo: So, do you mean something like this?
```rust
        let path_cstr = match hintpath {
            Some(path) => Some(try!(path.into_c_string())),
            None => None,
        };
        let path = match path_cstr {
            Some(path) => path.as_ptr(),
            None => ptr::null(),
        };
```

alexcrichton: Sure yeah I think that could work!

src/blob.rs
```diff
@@ -68,10 +69,67 @@ impl<'repo> Drop for Blob<'repo> {
...
+/// A structure to represent a git writestream for blobs
+pub struct BlobWriter<'repo> {
+    raw: *mut raw::git_writestream,
+    need_cleanup: bool,
+    _marker: marker::PhantomData<Object<'repo>>,
+}
+
+impl<'repo> BlobWriter<'repo> {
+    /// Finalize blob writing stream and write the blob to the object db
+    pub fn commit(&mut self) -> Result<Oid, Error> {
```
alexcrichton: Oh one other thing I thought of, I think this should take `self` right to avoid a double-free from calling this twice?

src/repo.rs
```diff
@@ -802,6 +802,33 @@ impl Repository {
...
+    /// If the `hintpath` parameter is filled, it will be used to determine
+    /// what git filters should be applied to the object before it is written
+    /// to the object database.
+    pub fn blob_writer(&self, hintpath: Option<&Path>) -> Result<BlobWriter, Error> {
+        let path_str = match hintpath {
+            Some(path) => Some(try!(path.into_c_string())),
+            None => None,
+        };
+        let path = match path_str {
+            Some(path) => path.as_ptr(),
```
alexcrichton: This'll actually want to be `ref path` to avoid the use-after free I believe

src/blob.rs
katyo: I seems to be understand that you mean but I require some hint how to do that right.

src/repo.rs
katyo: Do you mean `Some(ref path)` here?

alexcrichton: Indeed!

src/blob.rs
alexcrichton: Oh you can just change this method from `&mut self` to `mut self`

katyo: Ok, thanks. It seems the borrowing methodology in Rust is much unlike the things I'v ever deal in C and Haskell :)

alexcrichton:
Looks great!

PR #240, opened 2017-08-16
Added ODB stream APIs

katyo:
An addition for solving the issue #238

### DIFF
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -1,0 +1,135 @@
+use std::marker;
+use std::io;
+use libc::{c_char};
+
+use {raw, Oid, Object, Error};
+use util::Binding;
+
+/// A structure to represent a git ODB rstream
+pub struct OdbReader<'repo> {
+    raw: *mut raw::git_odb_stream,
+    _marker: marker::PhantomData<Object<'repo>>,
+}
+
+impl<'repo> Binding for OdbReader<'repo> {
+    type Raw = *mut raw::git_odb_stream;
+
+    unsafe fn from_raw(raw: *mut raw::git_odb_stream) -> OdbReader<'repo> {
+        OdbReader {
+            raw: raw,
+            _marker: marker::PhantomData,
+        }
+    }
+    fn raw(&self) -> *mut raw::git_odb_stream { self.raw }
+}
+
+impl<'repo> Drop for OdbReader<'repo> {
+    fn drop(&mut self) {
+        unsafe { raw::git_odb_stream_free(self.raw) }
+    }
+}
+
+impl<'repo> io::Read for OdbReader<'repo> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        unsafe {
+            let ptr = buf.as_ptr() as *mut c_char;
+            let len = buf.len();
+            let res = raw::git_odb_stream_read(self.raw, ptr, len);
+            if res < 0 {
+                Err(io::Error::new(io::ErrorKind::Other, "Read error"))
+            } else {
+                Ok(len)
+            }
+        }
+    }
+}
+
+/// A structure to represent a git ODB wstream
+pub struct OdbWriter<'repo> {
+    raw: *mut raw::git_odb_stream,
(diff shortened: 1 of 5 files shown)


src/odb.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/// A structure to represent a git ODB wstream
+pub struct OdbWriter<'repo> {
+    raw: *mut raw::git_odb_stream,
+    _marker: marker::PhantomData<Object<'repo>>,
+}
+
+impl<'repo> OdbWriter<'repo> {
+    /// Finalize ODB writing stream and write the object to the object database
```
alexcrichton: Could you add a comment here about what happens if you call this twice and what happens if you call write after you call this?

src/odb.rs
```diff
@@ -0,0 +1,135 @@
...
+impl<'repo> io::Write for OdbWriter<'repo> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        unsafe {
+            let ptr = buf.as_ptr() as *const c_char;
+            let len = buf.len();
+            let res = raw::git_odb_stream_write(self.raw, ptr, len);
+            if res < 0 {
+                Err(io::Error::new(io::ErrorKind::Other, "Write error"))
+            } else {
+                Ok(buf.len())
```
alexcrichton: I think that `res` may be how much was actually written?

src/repo.rs
```diff
@@ -811,6 +825,30 @@ impl Repository {
...
     }
 
+    /// Create object database reading stream
+    pub fn odb_reader(&self, oid: Oid) -> Result<OdbReader, Error> {
+        let mut out = ptr::null_mut();
+        unsafe {
+            let mut db = ptr::null_mut();
+            try_call!(raw::git_repository_odb(&mut db, self.raw()));
+            try_call!(raw::git_odb_open_rstream(&mut out, db, oid.raw()));
+            raw::git_odb_free(db);
```
alexcrichton: I think this is leaked if the `open_rstream` fails, but in general it looks like there's room for an `Odb` abstraction here?

I'd imagine that, given enough time at least, there'd be an `odb` method and then from that `Odb` object you could create readers/writers?

src/odb.rs
katyo: Ok

src/odb.rs
katyo: It's a common practice in C but not in the libgit2. The write function returns 0 if the write succeeded and error code otherwise ([git_odb_stream_write](https://libgit2.github.com/libgit2/#HEAD/group/odb/git_odb_stream_write)). Also you can see usage example like [here](https://github.com/libgit2/libgit2/blob/master/src/blob.c#L66-L67).

src/repo.rs
katyo: Of course, initially I planned to introduce ODB objects, but after looking the libgit2 sources I seen that the common practice is accessing to ODB by weakptr because the ODB associated with repository. For reference you can look the source of  [git_blob_create_frombuffer](https://github.com/libgit2/libgit2/blob/master/src/blob.c#L62) function. I used `git_repository_odb__weakptr` like here. It works for me, but have the compilation error on your CI (may be it related to different versions of libgit2).

src/odb.rs
alexcrichton: Ah ok I see, thanks for the info!

src/repo.rs
alexcrichton: Are the weak pointer APIs public APIs? I'm not quite sure what you mean :(

katyo: I mean that introducing ODB objects is unnecessary here.
I think it will be sufficient to fix pointer leakage in this case.

alexcrichton: My worry is that idiomatically these functions on `Repository` won't exist over if we had an ODB object representation. Could a "bare bones" version be added which could get expanded later?

alexcrichton:
Egad sorry I missed the update here! Feel free to ping me whenever a rebase happens, but otherwise this looks great and thanks again!

PR #244, opened 2017-08-29
Make Repo::update_submodules public

dylanmckay:
The method is quite useful on its own.


alexcrichton:
Thanks! I'm sort of wary about doing this though as in theory it's all best done in parallel and there's a lot of various ways to tweak this, so I'm not sure if this is the best stable method to do this with.

dylanmckay:
That makes sense, I realised I can work around this by explicitly calling `Submodule::update` recursively, which also allows me to print status messages.

I'll close this MR

PR #248, opened 2017-09-11
Allow overriding lib/include dirs when cross-compiling.

jessicah:
Needed when cross-compiling cargo, to use provided shared libraries from Haiku packages.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -29,6 +29,21 @@ fn main() {
     if curl {
         register_dep("CURL");
     }
+
+    let target = env::var("TARGET").unwrap();
+    let host = env::var("HOST").unwrap();
+
+    if target != host {
+        let lib_dir = env::var_os("TARGET_LIBGIT2_LIB_DIR").map(PathBuf::from);
+        let include_dir = env::var_os("TARGET_LIBGIT2_INCLUDE_DIR").map(PathBuf::from);
+
+        if lib_dir.is_some() && include_dir.is_some() {
+            println!("cargo:rustc-link-search=native={}", lib_dir.unwrap().to_string_lossy());
+            println!("cargo:include={}", include_dir.unwrap().to_string_lossy());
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think though that this is the use case for [library overrides in Cargo](http://doc.crates.io/build-script.html#overriding-build-scripts)?

jessicah:
Huh, interesting. I will try this out! If it works, commit the cargo change and close these PRs...

jessicah:
Hmm, that doesn't seem portable though? Don't I need to put a path in there? That doesn't seem like something that should get committed to the cargo repo. And I don't need these overrides when compiling from Haiku natively, as pkg-config should work there.

However, I'd like to see Haiku CI happen, especially since we otherwise get left behind quickly, so that's why I want to get these merged rather than maintain on my own as I've done in the past. Took me almost a week to get Haiku building again >_<

alexcrichton:
Yes this form of configuration is typically committed with build configuration rather than source code, although sometimes that can live in the same place, like in rust-lang/rust

sfackler:
The buildscript override approach may be a bit iffy for openssl-sys since you'll need to be careful to match up the exact version of openssl and all of the compile time flags it was configured with.

alexcrichton:
It's true yeah, in general, but for libgit2/libssh2 that's what I'd expect to be used to basically override everything.

jessicah:
I don't think the library overrides are a good approach in this circumstance. That seems suited to temporary overrides by a developer working on new features or a pre-release library.

It would probably be useful to extract the [helper function](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/build.rs#L39-L50) used in the openssl crate into the gcc crate perhaps, so that there is a consistent override for environment variables for the cross-compilation story.

alexcrichton:
Hm so this is precisely why overrides exists, and I don't really want to support an exact mirror of such a feature in this crate. Do overrides not work for you?

alexcrichton:

PR #252, opened 2017-09-20
Odb foreach bindings

daniel-gustafsson:
Bindings for git_odb_foreach.

I noticed that much of this is duplicated in #240, so just the foreach call and callback should probably be rebased on that code first if there is interest in merging this.


alexcrichton:
Looks great to me, thanks!

PR #255, opened 2017-09-22
Add git_merge_trees bindings

uzytkownik:
I guess it can be workaround by creating dummy commits but it should be more straightforward.


alexcrichton:
Thanks!

PR #258, opened 2017-10-03
Implement Sync for git2::Repository

hwright:
git2::Repository already implements Send, which is required for multiple
readers across different threads.  However, both the comment above that
implementation, as well as the comments in libgit2 itself seem to
indicate that the underlying datastructures themselves, can be sync'd,
not just sent.

This has a practical application in cases when trying to share a
Repository instances inside an Iron handler, for example.


alexcrichton:
Thanks for the PR! Unfortunately though I don't think this is memory safe as the underlying type isn't safe to share across threads concurrently, but I think a `Mutex` should work?

Issue #260, opened 2017-10-09
Some structs are not documented


nbigaouette-eai:
I found two structs that are unfortunately not documented on docs.rs.

Can't remember the first one, but the second one is `SubmoduleUpdateOptions`:
* See `Submodule.update()` docs: https://docs.rs/git2/0.6.8/git2/struct.Submodule.html#method.update The `SubmoduleUpdateOptions` is not clickable.
* Searching for it returns no hit: https://docs.rs/git2/0.6.8/git2/index.html?search=SubmoduleUpdateOptions
* Source contains documentation: https://github.com/alexcrichton/git2-rs/blob/master/src/submodule.rs#L205-L210

What kind of glitch (if it's really one) could be causing this? Maybe a cargo issue?

nbigaouette-eai:
Weirdly enough, this module is private? I cannot use it... Maybe that's why that it wasn't documented?

```
error[E0603]: module `submodule` is private
   --> src/git.rs:182:53
    |
182 |                         let mut submodule_options = git2::submodule::SubmoduleUpdateOptions::new();
    |                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error
```

In any case because I cannot create a `git2::submodule::SubmoduleUpdateOptions`, I cannot call `git2::submodule::Submodule::update()`! :(

alexcrichton:
Oh oops! These should be reexported at the root for sure and that I think would force them to be documented, thanks for finding this!

Would you be interested in sending a PR for this?

nbigaouette-eai:
Sure, if it's as simple as the re-export, I'll give it a try!

PR #262, opened 2017-10-10
Add script to automatically generate tests for pub structs

nbigaouette:
This PR adds a bash script that will get all structs marked `pub` and, if their module has not been `pub` exported, will generate a simple test that try to `use` them.

Reported in #260, a struct was found to not have been re-exported at the lib level. I did remember that I found the same issue with another struct, but couldn't remember which one. Since there is many, I decided to script this to detect any error. Found 3 other structs (`ParentIds`, `StashCbData` and `TestTreeIter`). I'm not sure one of those was the one I had issue with initially, but they should still be fixed.

The PR will add the tests and their fix.

I'll push the commits one after the other to trigger the CI at every step.


nbigaouette:
Turns out, `TestTreeIter` is a test struct... urg. I've changed the script to start match at beginning of line.

I've force pushed to change this:

* 4c3de54 -> e6ace15
* 03d0073 -> f998bc7

nbigaouette:
Unfortunately, `StashCbData` is not documented, which will prevent the crate to build... :(

nbigaouette:
I've added (in 8cdfc78) the execution of the script to Travis config, making sure that the tests are always up to date.

I'm not sure about AppVeyor though (bash on windows? pipes? might be problematic...) but Travis is probably sufficient.

alexcrichton:
Thanks! My shell-fu is pretty awful, but I wonder if this could perhaps be a test that's generated from a build script?

nbigaouette:
I wish I was as fluent in rust as I'm in shell ;)

It would definitely be better in rust though. I'll see what I can do!

nbigaouette:
By the way, what should be done about the missing docstring? Is that struct ok being public? What is it's meaning?

alexcrichton:
It's ok to either fill in the docs or tag it with `#[allow(missing_docs)]` for now

nbigaouette:
I'm working on a `build.rs` version... should be coming soon ;)

alexcrichton:
Thanks! A few comments on the organization here (but otherwise looks great)

* Could this generate code into the build directory rather than the source directory? You should be able to include that via `include!` in the test itself.
* Could the build script be moved to a separate project? (something like a `git2-test` folder). That way we won't have to worry about this running by accident and/or failing when git2 is used as a crate (not with tests)

alexcrichton:

PR #268, opened 2017-11-10
reference: Expose ReferenceType

abhinav:
This adds a `kind()` method to `Reference` which uses
[`git_reference_type`][1] to get the type of the reference and converts
it to the enum `ReferenceType`.

[1]: https://libgit2.github.com/libgit2/#HEAD/group/reference/git_reference_type

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -294,6 +294,16 @@ pub enum ObjectType {
     Tag,
 }
 
+/// An enumeration of all possile kinds of references.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum ReferenceType {
+    /// A reference which points at an object id.
+    Oid,
+
+    /// A reference which points at another reference.
+    Symbolic,
+}
+
 /// An enumeration for the possible types of branches
 #[derive(PartialEq, Eq, Debug, Copy, Clone)]
 pub enum BranchType {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Instead of an enum here, should this perhaps be a newtype wrapper struct with boolean accessors?

abhinav:
@alexcrichton Are you imagining something like this?

```rust
let rtype = reference.kind();
if rtype.is_oid() {
    // ..
}
if rtype.is_symbolic() {
    // ..
}
```

Are you sure about that? It doesn't seem quite right because the two cases are
mutually exclusive. You also lose the ability to `match` on them. If it
matters, [pygit2][1] and [git2go][2] appear to be using enum-equivalents for
this.

I'm happy to make the change if you would still prefer it like that.

[1]: http://www.pygit2.org/references.html#pygit2.Reference.type
[2]: https://godoc.org/github.com/libgit2/git2go#ReferenceType

alexcrichton:
Hm ok, are we sure that these are the only two kinds of kinds that will ever exist? The problem with an `enum` in Rust is that it can never be extended, while enums in most other languages can be extended over time.

abhinav:
I believe that git only has two kinds of references: direct, which point to an
object ID, and symbolic, which point to another reference. libgit2 maps these
to OID and SYMBOLIC. Looking at libgit2 and existing bindings to it, these two
appear to be the only valid values.

libgit2 also declares a [`GIT_REF_LISTALL`][1] which appears to be completely
unused at this point in the library. It used to be a magic value for
[`git_reference_list` in libgit2 v0.18][2] to request all types of references
but that parameter was dropped in v0.19.

Barring any unforeseen breaking changes in libgit2, I'm fairly confident that
the enum will not need to be extended.

[1]: https://github.com/libgit2/libgit2/blob/fd1492e82783159231d046beb22a20d17029bc61/include/git2/types.h#L198

[2]: https://libgit2.github.com/libgit2/#v0.18.0/group/reference/git_reference_list

alexcrichton:
Ok! Sounds good enough for me

PR #270, opened 2017-11-17
Fix typo in Delta::Renamed's docstring

Phrohdoh:
https://github.com/libgit2/libgit2/blob/6be167f13076294721b8707197ce2d049313fb0b/include/git2/diff.h#L251


alexcrichton:
Thanks!

PR #275, opened 2017-12-10
Add some ignore rules APIs

kubo39:


src/repo.rs
```diff
@@ -1891,6 +1891,42 @@ impl Repository {
...
+    pub fn add_ignore_rule(&self, rule: &str) -> Result<(), Error> {
+        let rules = CString::new(rule)?;
+        unsafe {
+            try_call!(raw::git_ignore_add_rule(self.raw, rules));
+        }
+        Ok(())
+    }
+
+    /// Add ignore rules for a repository.
+    pub fn add_ignore_rules(&self, rules: &Vec<&str>) -> Result<(), Error> {
```
alexcrichton: In general this library tries to match upstream libgit2 as much as possible, so perhaps this could be omitted for now?

src/repo.rs
```diff
@@ -1891,6 +1891,42 @@ impl Repository {
             Ok(())
         }
     }
+
+    /// Add an ignore rule for a repository.
```
alexcrichton: Could this documentation be expanded a bit to mention that new lines are a llowed, for example?

alexcrichton:
Thanks!

PR #276, opened 2017-12-20
Don't shutdown libgit2 at exit

sfackler:
atexit handlers can be run while other threads are still alive, and in
particular still using git2 types. This can manifest as hitting this
assert, for example:
https://github.com/libgit2/libgit2/blob/master/src/mwindow.c#L106


sfackler:
Thanks! Could you cut a release when you have a chance?

alexcrichton:
Already done!

PR #277, opened 2017-12-25
Implement reading an object from the object database.

kaendfinger:
I've been using Rust for only two days, so feel free to criticize everything :)

### DIFF
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -1,6 +1,8 @@
 use std::marker;
 use std::io;
 use std::ptr;
+use std::slice;
+
 use libc::{c_char, c_int, c_void};
 
 use {raw, Oid, Object, ObjectType, Error};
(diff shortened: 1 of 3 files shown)


kaendfinger:
Related to #124, however this is a quite different API.
I'll look into a writing API too, but reading is much more important to have this level of an API, I feel.

src/odb.rs
```diff
@@ -66,6 +68,70 @@ impl<'repo> Odb<'repo> {
...
+    pub fn read(&self, oid: Oid) -> Result<OdbObject, Error> {
+        let mut out = ptr::null_mut();
+        unsafe {
+            try_call!(raw::git_odb_read(&mut out, self.raw, oid.raw()));
+            Ok(OdbObject::from_raw(out))
+        }
+    }
+
+    /// Checks if the object database has an object.
+    pub fn exists(&self, oid: Oid) -> Result<bool, Error> {
```
alexcrichton: Should this perhaps just return a `bool` instead of a `Result`? (not sure it can return an `Error`

src/odb.rs
```diff
@@ -66,6 +68,70 @@ impl<'repo> Odb<'repo> {
...
+
+impl<'a> Drop for OdbObject<'a> {
+    fn drop(&mut self) {
+        unsafe { raw::git_odb_object_free(self.raw) }
+    }
+}
+
+impl<'a> OdbObject<'a> {
+    /// Get the object type.
+    pub fn get_type(&self) -> Result<ObjectType, Error> {
```
alexcrichton: Similar to above (and the ones below) I don't think a `Result` needs to be returned here, right?

src/odb.rs
kaendfinger: That's correct, I'll fix that.

src/odb.rs
kaendfinger: I have fixed this, but it requires calling unwrap(), which feels wrong in a library, let me know what you think.

alexcrichton:
Thanks!

PR #286, opened 2018-01-17
Update to libgit2 f1323d9c

carlosmn:
This is the current latest as of now. I'll leave it up to you to decide when the sys crate needs a bump.

Most of the changes are mechanical, though the library did change how we specify which features are active, so I had to change how ssh support is detected. This might also mean that the `!msvc` part of the check can go away, but I don't have a Windows machine handy to test.

The status options has grown a pointer to a tree to optionally specify a base tree instead of HEAD, but when I tried to add it I got all tangled up in lifetimes and `Option`s so you can't make use of that in git2-rs for now.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -130,10 +130,10 @@ fn main() {
     // Make sure libssh2 was detected on unix systems, because it definitely
     // should have been!
     if ssh && !msvc {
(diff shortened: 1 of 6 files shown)


alexcrichton:
Ok I think CI is now wrangled so I'll merge, thanks @carlosmn!

PR #292, opened 2018-02-03
oid: Correct documentation of errors on from_str

kamalmarhubi:
The `from_str` implementation just calls the libgit2 function
`git_oid_fromstrn`, which accepts any non-empty hex-string with length
less than 40:

  https://github.com/libgit2/libgit2/blob/8abd514c270ba3b3fc42c6d5feefedc4cc65dd9b/src/oid.c#L31-L42


alexcrichton:
Thanks!

PR #295, opened 2018-02-05
oid: Allow Oid to represent short oids

### LINKED ISSUE #293: First class short oids?
The C side of an oid is a `char[GIT_OID_RAWSZ]`. In the various `{read,write,exists}_prefix` APIs, the oid is treated as short, with the number of set nibbles (hex chars in hex representation), eg,

```rust
impl <'repo> Odb<'repo> {
    // ...
    pub fn exists_prefix(&self, short_oid: Oid, len: usize) -> Result<Oid, Error> {
        // ...
    }
}
```

In the spirit of slice and str, it would be cleaner and rustier if the oid itself kept track of its length:

```rust
/// Unique identity of any object (commit, tree, blob, tag).
#[derive(Copy, Clone)]
pub struct Oid {
    raw: raw::git_oid,
    len: u32,
}
```

This way, the `_prefix` methods can get rid of need to pass length separately. This would be a breaking change though. :-(

It may be possible to introduce this in a non-breaking way by adding the `len` as an option, and having those functions treat length 0 as a signal to read from the oid; this would preserve behaviour for existing uses. New APIs such ODB backends could simply not have the `len` parameter.

Thoughts?

kamalmarhubi:
Add a `len` field to `Oid` to allow it to track if it's a short
oid to use as a prefix. This makes the `exists_prefix` API easier to
use, as you no longer need to track the length yourself.

BREAKING CHANGES:
- `Oid` now tracks its own length, allowing it to represent both short
and full oids.  As a result, `Odb::exists_prefix` no longer takes a
length parameter.
- The `OidArray` no longer implements `Deref<Target=[Oid]>`. To access
the contained oids, use the iterator returned by `OidArray::iter()`.

closes #293

### DIFF
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -129,7 +129,15 @@ impl<'repo> Odb<'repo> {
     }
 
     /// Potentially finds an object that starts with the given prefix.
+    ///
+    /// If `len` is 0, gets the length from `short_oid`. The `len` parameter
+    /// will be removed in the next major version.
     pub fn exists_prefix(&self, short_oid: Oid, len: usize) -> Result<Oid, Error> {
+        let len = match len {
+            0 => short_oid.len(),
+            n => n,
+        };
+
         unsafe {
             let mut out = raw::git_oid { id: [0; raw::GIT_OID_RAWSZ] };
             try_call!(raw::git_odb_exists_prefix(&mut out,
@@ -412,6 +420,9 @@ mod tests {
         let id_prefix_str = &id.to_string()[0..10];
         let id_prefix = Oid::from_str(id_prefix_str).unwrap();
         let found_oid = db.exists_prefix(id_prefix, 10).unwrap();
+        assert_eq!(found_oid, id);
+        // Check the short OID behaviour works.
+        let found_oid = db.exists_prefix(id_prefix, 0).unwrap();
         assert_eq!(found_oid, id);
     }
 }
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -129,21 +129,13 @@ impl<'repo> Odb<'repo> {
     }
 
     /// Potentially finds an object that starts with the given prefix.
-    ///
-    /// If `len` is 0, gets the length from `short_oid`. The `len` parameter
-    /// will be removed in the next major version.
-    pub fn exists_prefix(&self, short_oid: Oid, len: usize) -> Result<Oid, Error> {
-        let len = match len {
-            0 => short_oid.len(),
-            n => n,
-        };
-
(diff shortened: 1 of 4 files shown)


kamalmarhubi:
@alexcrichton not sure what to do re breaking. I was most of the way done before I realised that the `Deref<Target=[Oid]>` impl wouldn't work, so decided to send it and see what you think.

kamalmarhubi:
I guess since the change to `OidArray` is breaking might as well drop the `len` parameter to `Odb::exists_prefix` at the same time. Will wait for feedback before doing that.

alexcrichton:
FWIW I think there's already breaking changes queued up so a breaking change is fine, but in general I also wouldn't mind learning a bit more about this. How does libgit2 handle short oids in this case? I think there's a whole slew of functions that all take an oid, and do they all extend a sort id to a full id (failing if that can't be done)? Or do only some functions handle a short id?

kamalmarhubi:
> FWIW I think there's already breaking changes queued up so a breaking change is fine

How do you track breaking changes? Is my SHOUTING in the commit message good enough?

> How does libgit2 handle short oids in this case? I think there's a whole slew of functions that all take an oid, and do they all extend a sort id to a full id (failing if that can't be done)? Or do only some functions handle a short id?

I'm not 100% on this, but from looking at the odb and odb backend APIs, only some handle a short oid. The one's I've seen are labelled `_prefix` in the name, take a `git_oid` and a `size_t len`. Some, like `git_odb_exists_prefix` also take an out parameter for the full oid.

Functions that take a `git_oid` without the length parameter have no way to know how much of the oid to treat as significant. The docs on `git_object_lookup_prefix` give a good idea of what's going on: https://libgit2.github.com/libgit2/#HEAD/group/object/git_object_lookup_prefix

(Also I just learned that I need to reject length less than `GIT_OID_MINPREFIXLEN` instead of only zero.)

kamalmarhubi:
Quick thought here: we could have a second type for possibly-short oids, with conversions where appropriate:
- `impl From<Oid> for OidPrefix`
- `impl TryFrom<OidPrefix> for Oid` (the equivalent given that's not stable yet)

Then we can be really clear in the API which functions accept which kind.

alexcrichton:
Hm ok so yeah I think it sounds like we shouldn't change `Oid` itself? It sounds like lots of functions in libgit2 (unrelated to the odb) already assume that it's a "full" oid? If that's true then yeah a separate type sounds fine by me.

In terms of management, I personally just go with `publish = false` in the manifest which is a reminder to my future self

kamalmarhubi:
I'm closing this and will open a replacement with a `ShortOid` struct soonish.

PR #296, opened 2018-02-07
Support Refspec generation by using libgit2's internal parsing api

cynecx:
This PR adds `Refspec::try_parse` which can parse a refspec string into a Refspec object. This uses libgit2's internal apis (`git_refspec__parse`, `git_refspec__free `).


alexcrichton:
Thanks! I think the internal apis should be avoided by consumers, though, right? (the CI errors also show that the refspect struct is intended to be opaque and private)

cynecx:
Well, I thought it would be a great idea to add support for this, as this might be important if you are dealing with git repositories but I didn't want to reimplement all the logic and since this library provides mostly bindings, so I thought most viable solution would be to use the internal api, since it's accessible. But yeah, these are internal apis which shouldn't be used but looking at other git as a library implementation it would be a loss since they support it ([go-git](https://github.com/src-d/go-git) for example).

If you think that this shouldn't be included, I could try to open a PR upstream and see where it goes.

alexcrichton:
Ok yeah a PR upstream to libgit2 sounds great!

cynecx:
JFYI: https://github.com/libgit2/libgit2/pull/4519

cynecx:
I am closing this for now. Just waiting for a new upstream release with should include the new api which we can then use :)

PR #298, opened 2018-02-11
Forward error from credential callback.

### LINKED ISSUE #297: Credentials callback discards error klass and message.
The signature of the credentials callback is `FnMut(&str, Option<&str>, CredentialType) -> Result<Cred, Error>`, so users expect to be able to return a meaningful error, for example https://github.com/rust-lang/crates.io/blob/47998b76328e82341ba374d74036231fcce7720f/src/git.rs#L208

However because of https://github.com/alexcrichton/git2-rs/blob/308ba2a0e74a4bad9a993b743c27937901cba76b/src/remote_callbacks.rs#L282 all the information in the error except for the error code is discarded.

Possible solutions:

1) Update the docs (https://github.com/alexcrichton/git2-rs/blob/308ba2a0e74a4bad9a993b743c27937901cba76b/src/remote_callbacks.rs#L44) to include a note about this.
2) Change the signature of the callback to return `Option<Cred>` instead, discarding the code as well, but making the signature more intuitive.
3) Set the error in the callback wrapper (#298).

Coneko:
Fixes #297.

### DIFF
--- a/src/remote_callbacks.rs
+++ b/src/remote_callbacks.rs
@@ -1,4 +1,4 @@
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 use std::marker;
 use std::mem;
 use std::slice;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #300, opened 2018-02-22
Update vendored libgit2 to v0.27.0-rc1

carlosmn:
The new release should be out shortly, so here's an update to the first RC.


alexcrichton:
Awesome, thanks! It looks like the `readstream` field may have changed as well for one of the structs?

carlosmn:
You're right. I completely missed that `lib.rs` contained a definition for the odb backend interface. I've updated and pushed with the change there.

alexcrichton:
Woo thanks!

PR #163, opened 2016-10-20
Add bindings for creating and extracting GPG signatures

cmr:
These were added in libgit2 v24

PR #301, opened 2018-02-22
Add bindings for creating and extracting GPG signatures

gchronis:
These were added in libgit2 v24. This is a clone of #163 by cmr which was closed due to inactivity. Comments from the original PR have been addressed, including changing the method name to `commit_signed` from `signed_commit`, being explicit about the ordering of return arguments, and removing the version bump.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -998,6 +998,56 @@ impl Repository {
         }
     }
 
+    /// Create a commit object from the given buffer and signature
+    ///
+    /// Given the unsigned commit object's contents, its signature and the
+    /// header field in which to store the signature, attach the signature to
+    /// the commit and write it into the given repository.
+    ///
+    /// Use `None` in `signature_field` to use the default of `gpgsig`, which is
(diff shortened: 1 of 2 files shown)


src/repo.rs
```diff
@@ -998,6 +998,56 @@ impl Repository {
...
+                             signature_field: Option<&str>)
+                             -> Result<(Buf, Buf), Error> {
+        let signature_field = try!(::opt_cstr(signature_field));
+        let signature = Buf::new();
+        let content = Buf::new();
+        unsafe {
+            try_call!(raw::git_commit_extract_signature(signature.raw(),
+                                                        content.raw(),
+                                                        self.raw(),
+                                                        commit.id().raw() as *mut _,
```
alexcrichton: If only the id of the commit is used here, should this function take an `Oid` instead?

gchronis: Thanks for pointing that out. Fixed.

alexcrichton:
Thanks!

PR #302, opened 2018-02-22
Fix lifetime of Object::peel_to_* result

### LINKED ISSUE #299: Wrong (elided) lifetimes for peel_to_*?
The following code doesn't compile:
```rust
    let env_tree = tree.get_name("foo")
        .unwrap()
        .to_object(repo)?
        .peel_to_tree()?;
```
because the resulting `Tree` has the lifetime of the short-lived object reference, instead of the repository. Am I missing something, or is that wrong?

Currently, the signature for `peel_to_tree` is
```rust
fn peel_to_tree(&self) -> Result<Tree, Error>
```
, but shouldn't it be
```rust
fn peel_to_tree(&self) -> Result<Tree<'repo>, Error>
```
?

flodiebold:
The peeled object lives as long as the repo.

Fixes #299.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -43,22 +43,22 @@ impl<'repo> Object<'repo> {
     }
 
     /// Recursively peel an object until a blob is found
-    pub fn peel_to_blob(&self) -> Result<Blob, Error> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #307, opened 2018-04-08
fix(cargo): Update minimal required dependencies

klausi:
This is necessary so that this crate compiles with `CARGO_REGISTRY_INDEX=https://github.com/klausi/crates.io-index cargo build -Z minimal-versions`

Note that the custom registry is needed because of the `links` entries that are missing on the main crates.io index.


klausi:
Urgs, the build failed because apparently "0.1.2" means that only a maximum version in the 0.1.x series can be selected. So we explicitly need to specify ">= 0.1.2".

alexcrichton:
Thanks!

alexcrichton:
Done!

PR #311, opened 2018-05-17
Add Config.remove_multivar

daniel-ferguson:
This is modelled after https://github.com/alexcrichton/git2-rs/pull/170

The test is getting a tiny bit complex - do you think it would be better split into separate cases?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -135,6 +135,17 @@ impl Config {
         }
     }
 
+    /// Remove multivar config variables in the config file with the highest level (usually the
+    /// local one).
+    pub fn remove_multivar(&mut self, name: &str, regexp: &str) -> Result<(), Error> {
+        let name = try!(CString::new(name));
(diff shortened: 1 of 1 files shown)


daniel-ferguson:
Hmm, it looks like `libgit2`'s multivar functions treat regexes differently on OSX?

https://travis-ci.org/alexcrichton/git2-rs/jobs/380466162
```
[...]
failures:
---- config::tests::multivar stdout ----
	thread 'config::tests::multivar' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: -1, klass: 8, message: "empty (sub)expression" }', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
[...]
```

I can replace the regex in `cfg.remove_multivar("foo.bar", "").unwrap();` with `".*"` for the same effect.

alexcrichton:
Thanks!

Issue #279, opened 2018-01-02
Status unusable without libgit2-sys


glfmn:
In order to properly use the `Status` struct, I have to know which bit corresponds to which flag, which requires `libgit2-sys` as far as I can tell, unless I do lots of very manual testing to figure out which is which.

For more context, I'm writing a binary which pretty-prints git stats, and I'm using git2 to interact with the repository to gather the information I need.

alexcrichton:
Sounds bad! Seems fine though to add various accessors as necessary to avoid needing libgit2-sys

glfmn:
From the usage examples in the bit-flags crate, it seems that it should be possible to use the flags as follows:

```rust
bitflags! {
    struct Flags: u32 {
        const FLAG_A       = 0b00000001;
        const FLAG_B       = 0b00000010;
        const FLAG_C       = 0b00000100;
        const FLAG_ABC     = Self::FLAG_A.bits
                           | Self::FLAG_B.bits
                           | Self::FLAG_C.bits;
    }
}

fn main() {
    let e1 = Flags::FLAG_A | Flags::FLAG_C;
    let e2 = Flags::FLAG_B | Flags::FLAG_C;
    assert_eq!((e1 | e2), Flags::FLAG_ABC);   // union
    assert_eq!((e1 & e2), Flags::FLAG_C);     // intersection
    assert_eq!((e1 - e2), Flags::FLAG_A);     // set difference
    assert_eq!(!e2, Flags::FLAG_A);           // set complement
}
```

I think the problem may stem from a simple lack of `pub` on the declarations, though I need to double check.  Accessors that check for each bit and return a `bool` directly could definitely still be useful, less boilerplate that way.  Would you mind if I submitted a PR for this?

alexcrichton:
Nah yeah definitely feel free to send a PR!

glfmn:
**It turns out it's just an error on my part, I was looking at the wrong version of the `bitflags` crate.**

The associated constants shown above appeared in version 1.0 of `bitflags`.  It's a breaking change in `bitflags` and would likely be a breaking change in `git2-rs` as well.  If you would like I can still bump the `bitflags` version to 1.0 and make all the fix all the broken tests and such in `git2-rs`, but it would need to be in the next release.

I can also still add the helper methods.

**The fix** on my part was simply to use the unqualified `STATUS_WT_NEW` family of constants from git2.

alexcrichton:
[...]

PR #312, opened 2018-05-21
Create helper methods on bitflag generated structs

glfmn:
Addresses #279 

When checking the value of a single flag, rather than typing out:

```rust
let s = Status::CURRENT;
s.intersect(Status::CURRENT);
// or
s & Status::CURRENT == STATUS::CURRENT
```

Add helper methods to simplify the above to:

```rust
let s = Status::CURRENT;
s.is_current();
```

Also, add a `Default` impl for any bitflag struct which defines a `DEFAULT` constant which simply returns the `DEFAULT` constant, allowing for more idiomatic code.  I think it would make sense to do the same for bitflag structs which define a `NONE` constant with no `DEFAULT`.

I only checked the top level `lib.rs` file, if there are more bitflags I'll happily track them down this week.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -432,6 +432,12 @@ impl CredentialType {
     is_bit_set!(is_username, CredentialType::USERNAME);
 }
 
+impl Default for CredentialType {
+    fn default() -> Self {
+        CredentialType::DEFAULT
+    }
+}
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Nice!

PR #313, opened 2018-05-21
Automate commit testing with proper stashing

glfmn:
Add a git hook with installation instructions to automate testing done on each commit.  Promotes proper handling of edge cases related to testing code that will not be in the commit, which masks errors that show up later as a nasty surprise.

It's opt-in, to install it simply use:

```sh
$ ln -s ../../.pre-commit.sh .git/hooks/pre-commit
$ chmod +x .pre-commit.sh # for good measure
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,18 @@ $ cd git2-rs
 $ cargo build
 ```
 
+### Automating Testing
+
+Running tests and handling all of the associated edge cases on every commit
+proves tedious very quickly.  To automate tests and handle proper stashing and
+unstashing of unstaged changes and thus avoid nasty surprises, use the
+pre-commit hook found [here][pre-commit-hook] and place it into the
+`.git/hooks/` with the name `pre-commit`.  You may need to add execution
+permissions with `chmod +x`.
+
+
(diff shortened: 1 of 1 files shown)


glfmn:
Just running

```sh
cargo check &&
cargo test
```

Is probably enough, but for sake of completeness I can always make equivalent to travis-ci with some version of:

```sh
cargo test --no-default-features
cargo test
cargo run --manifest-path systest/Cargo.toml --release
```

alexcrichton:
Thanks! Since hooks have to be manually set up though I'm not sure if this belongs in the repo? Perhaps this could be listed in the README with a link to a gist?

glfmn:
It depends on what you want to prioritize I suppose; I imagine it's less likely someone would go through the process of downloading it and copy pasting versus just doing a symlink.  But I'm happy to go that route!  It's definitely just my preference to do it this way.

alexcrichton:
Ok sure yeah, let's go with the route of a snippet in the README

glfmn:
Gist found [here](https://gist.github.com/glfmn/0c5e9e2b41b48007ed3497d11e3dbbfa).

Updated, it should be good!

alexcrichton:
Thanks!

PR #315, opened 2018-05-24
Add bindings for rebase

### LINKED ISSUE #146: Please provide bindings for the rebase API

quodlibetor:
This is almost just a rebase of #215, which in turn closes #146. The only changes that I made are in the last two commits, which I believe resolves the open issues.

All conflicts other than a pair of overlapping `use` lines were "clean" in that it was obvious that both sides of the conflict should be kept.

Other than running the tests I haven't actually checked to see if this is fit for purpose.


alexcrichton:
Thanks for the PR! It looks like CI may be failing though in that the sys crate is slightly different than the headers? A missing `const` maybe?

quodlibetor:
Yup, the failure is in systest which is passing for me, I'll look into it.

I've also now verified that this stuff does work, and I think I've got an idea for how tests should work, so I might try and add some.

quodlibetor:
On second thought I'm not really sure what to do about the failure, it is failing for me, I just wasn't running systest locally.

I think that the error is in a `ctest`-generated test that verifies the type of the `git_rebase_operation.id` field. This is the function that the error is pointing to:

```c
// $OUT_DIR/all.c
git_oid* __test_field_type_git_rebase_operation_id(git_rebase_operation* b) {
    return &b->id;
}
```

Which is I believe verifying that

```rust
// libgit2-sys/lib.rs
#[repr(C)]
pub struct git_rebase_operation {
    pub kind: git_rebase_operation_t,
    pub id: git_oid,
    pub exec: *const c_char
}
```

has a `const git_oid id` because of the C struct:

```C
// libgit2-sys/libgit2/include/git2/rebase.h
typedef struct {
    git_rebase_operation_t type;
    const git_oid id;
    const char *exec;
} git_rebase_operation;
```

Which does indeed want a `const git_oid`.

Sadly, I can't find examples of *any other structs* in `libgit2` that use the `const <bare_field>` style -- everything is like `const char *`, which can be hidden behind a `*const c_char` on the rust side.

A cursory glance through the `build.rs` scripts doesn't reveal anything to me. *Maybe* this is a `ctest`-related issue?

alexcrichton:
Aha interesting! In that case it's probably fine to ignore the field here, ctest should have plenty of options to skip things

This does sound like a great bug though! I'm not sure there is an option to have this successfully tested right now

src/rebase.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+impl<'rebase, 'repo: 'rebase> Iterator for RebaseOperations<'rebase, 'repo> {
+    type Item = Result<RebaseOperation<'rebase, 'repo>, Error>;
+    fn next(&mut self) -> Option<Result<RebaseOperation<'rebase, 'repo>, Error>> {
+        if self.count >= self.rebase.operation_count() {
+            None
+        } else {
+            self.count += 1;
+            Some(self.rebase.next())
```
alexcrichton: Hm this means that calling `.operations_iter()` will change over time, right? Should this perhaps use `operation_at_index`?

systest/build.rs
```diff
@@ -34,5 +34,6 @@ fn main() {
         }
     });
     cfg.skip_type(|t| t == "__enum_ty");
+    cfg.skip_field_type(|s, field| s == "git_rebase_operation" && field == "id");
```
alexcrichton: Could you add a TODO here pointing at the ctest issue?

src/rebase.rs
```diff
@@ -0,0 +1,377 @@
...
+        unsafe {
+            try_call!(raw::git_rebase_inmemory_index(&mut ret, self.raw));
+            Ok(Binding::from_raw(ret))
+        }
+    }
+
+    /// Performs the next rebase operation and returns the `RebaseOperation` about it.
+    ///
+    /// This is the fundamental operation that `operation_iter` relies upon
+    pub fn next<'rebase>(&self) -> Result<RebaseOperation<'rebase, 'repo>, Error> {
```
alexcrichton: Should this `'rebase` be tied to `&self`?

src/rebase.rs
```diff
@@ -0,0 +1,377 @@
...
+        let mut ret = 0 as *mut raw::git_rebase_operation;
+        unsafe {
+            try_call!(raw::git_rebase_next(&mut ret, self.raw));
+            Ok(Binding::from_raw(ret))
+        }
+    }
+
+    /// Iterator of rebase operations
+    pub fn operation_iter(&self) -> RebaseOperations {
```
alexcrichton: Perhaps this could be called `operations`?

src/rebase.rs
quodlibetor: tl;dr: I've changed this to `self.rebase.current_operation_index().map(|_| self.rebase.next())` and removed the `count` field.

This function is the only place where `self.count` is ever accessed or mutated. It appears to just be a way of keeping track of where we are in the rebase to make sure we don't call the underlying `git_rebase_operation_next` function incorrectly. I believe that git's internal `current_operation_index` function will be re-entrant, where `count` will fail if you try to use this iterator.

Based on [the docs for `git_rebase_next`](https://libgit2.github.com/libgit2/#HEAD/group/rebase/git_rebase_next) I believe that we need to call it in order to actually make progress in the rebase. [`git_rebase_operation_byindex`](https://libgit2.github.com/libgit2/#HEAD/group/rebase/git_rebase_operation_byindex) doesn't say anything that suggests that it interacts with any sort of internal state machine. So I don't think it's fine to swap to just using `operation_at_index`.

systest/build.rs
quodlibetor: Done.

src/rebase.rs
quodlibetor: Oh yeah. Also several others.

src/rebase.rs
quodlibetor: Seems nicer, yes.

quodlibetor:
I've added a test and, while I was wrong that I could just `map` over `current_operation_index` the basic idea of just asking git what's up seems to work correctly. There is now also a test.

src/repo.rs
```diff
@@ -2051,6 +2052,54 @@ impl Repository {
...
         Ok(ignored == 1)
     }
+
+    /// Initializes a rebase operation to rebase the changes in `from_branch`
+    /// relative to `upstream` onto '`to_branch`. To begin the rebase
+    /// process, call `next()` or `operations()` on the returned object.
+    pub fn rebase_init(&self,
+                       from_branch: Option<AnnotatedCommit>,
+                       upstream: Option<AnnotatedCommit>,
+                       to_branch: Option<AnnotatedCommit>,
```
alexcrichton: Since this doesn't need ownership, could these be `&AnnotatedCommit`?

src/repo.rs
```diff
@@ -2051,6 +2052,54 @@ impl Repository {
...
+            drop(from_branch);
+            drop(upstream);
+            drop(to_branch);
+            drop(options_raw);
+            Ok(Binding::from_raw(ret))
+        }
+    }
+    /// Opens an existing rebase that was previously started by either an
+    /// invocation of `rebase_init` or by another client.
+    pub fn rebase_open(&self, options: Option<RebaseOptions>) -> Result<Rebase, Error> {
```
alexcrichton: I think it should be safe for this to be `Option<&mut RebaseOptions>`

src/rebase.rs
```diff
@@ -0,0 +1,487 @@
...
+    /// raw value of options
+    pub unsafe fn raw(&mut self) -> raw::git_rebase_options {
+        let mut checkout_options: raw::git_checkout_options = mem::zeroed();
+        raw::git_checkout_init_options(&mut checkout_options, raw::GIT_CHECKOUT_OPTIONS_VERSION);
+        if let Some(ref mut checkout_builder) = self.checkout_builder {
+            checkout_builder.configure(&mut checkout_options);
+        }
+        let mut merge_options: raw::git_merge_options = mem::zeroed();
+        if let Some(ref opts) = self.merge_options {
+            ptr::copy(opts.raw(), &mut merge_options, 1);
```
alexcrichton: I think this can be `merge_options = opts.raw()` perhaps?

src/rebase.rs
```diff
@@ -0,0 +1,487 @@
...
+    /// Perform an in-memory rebase, will not updated the repository to be in a rebasing-state
+    /// or modify the working directory.
+    pub fn in_memory(&mut self, in_memory: bool) -> &mut Self {
+        self.in_memory = in_memory;
+        self
+    }
+
+    /// The name of the notes reference used to rewrite notes
+    /// for rebased commits when finishing the rebase
+    pub fn rewrite_notes_ref(&mut self, rewrite_notes_ref: Option<CString>) -> &mut Self {
```
alexcrichton: I think idiomatically currently this library uses `fn rewrite_notes_ref<T: IntoCString>(&mut self, ref: T) -> ...` to avoid the `Option<CString>`

src/rebase.rs
```diff
@@ -0,0 +1,487 @@
...
+    ///
+    /// Verifies that there _is_ a next operation and then calls
+    /// `Rebase::next`, propagating any errors from there.
+    fn next(&mut self) -> Option<Result<RebaseOperation<'rebase, 'repo>, Error>> {
+        // The current_operation_index() returns the value for the _previous_
+        // invocation of rebase.next(), which means it is None on the first
+        // invocation and we need to offset idx + 1 for the general "are we
+        // done with the loop" query.
+        match self.rebase.current_operation_index() {
+            None => Some(self.rebase.next()),
```
alexcrichton: I think the usage of `next()` here will cause the iterator to not work when run twice, so perhaps this could use `operation_at_index`?

src/rebase.rs
```diff
@@ -0,0 +1,487 @@
...
+    ///
+    /// If the operation is one that applies a patch (which is any operation
+    /// except GIT_REBASE_OPERATION_EXEC) then the patch will be applied and
+    /// the index and working directory will be updated with the changes. If
+    /// there are conflicts, you will need to address those before committing
+    /// the changes.
+    ///
+    /// Iterating over the result of `operations()` correctly handles begin and
+    /// end conditions for you.
+    pub fn next<'rebase>(&'rebase self) -> Result<RebaseOperation<'rebase, 'repo>, Error> {
```
alexcrichton: To avoid weird iterator issues perhaps this could take `&'rebase mut self`?

src/repo.rs
quodlibetor: Okay.

src/repo.rs
quodlibetor: Done.

src/rebase.rs
quodlibetor: I had to deref it from a `*const` but `merge_options = *opts.raw();` seems semantically correct?

quodlibetor: Oh wait no I get `cannot move out of dereference of raw pointer` on compile (but not from rls). The `git_merge_options` struct seems like it would be safe to derive copy/clone on, and there are a few Copy types in `libgit2-sys/lib.rs`, so I've done that. Happy to change it back, though.

src/rebase.rs
quodlibetor: Hmm `IntoCString` doesn't support returning an `Option<CString>`, just a `Result`, and it's not implemented on `Option<IntoCString>`. I could do the Into dance:

```rust
    pub fn rewrite_notes_ref<T, U>(&mut self, rewrite_notes_ref: T) -> &mut Self
    where
        T: Into<Option<U>>,
        U: IntoCString,
    {
        self.rewrite_notes_ref = rewrite_notes_ref.into().map(|s| s.into_c_string().unwrap());
        self
    }
```

Possibly transforming the return value into a result, but I think that the chance that a ref isn't a valid CString _probably_ requires pretty crazy input. I'll put that up as the change, but lmk if you'd prefer some other approach.

src/rebase.rs
quodlibetor: I'm not sure exactly which weird iterator issues you're talking about, the main issue from my perspective isn't intra-process race conditions (which are weird) but the fact that rebases stop all the time and it's not crazily uncommon for rebases to start in one tool and then have some of the work happen in another tool.

For example, starting a rebase in an IDE, running into a conflict, editing in the terminal, then running `git rebase --continue` in the terminal, before jumping back into the IDE. There is a similar problem with the `operation_count`, inasmuch as it's possible for a rebase to be completed and a new one to be started which will cause unexpected errors. But in that case the entire rebase is invalid so I think it's a more reasonable to throw an error there.

I guess the other position is that keeping track of where the in-memory rebase is supposed to be allows the developer to decide how to handle the situation.

Also I can't seem to get the borrow checker to agree with addng that `mut`.

src/rebase.rs
quodlibetor: I think that each `RebaseOprations` needs to be considered a 1-shot iterator, because of the state that git keeps internally.

src/rebase.rs
quodlibetor: I guess another way of saying that is that `Rebase` can't be immutable: outside processes can affect its state.

Additionally, `RebaseOperations` should probably be either fully stateless -- getting rid of the `operation_count` field, and not requiring uniq access. Or it should be stateful and I think everything would be easier if `fn operations()` took self by-value. I've implemented that as well as adding the `mut` to the borrow (which I never got working) let me know what you think.

src/rebase.rs
```diff
@@ -0,0 +1,492 @@
...
+            rebase: self,
+            operation_count: op_count,
+        }
+    }
+
+    /// Gets the rebase operation specified by the given index.
+    pub fn operation_at_index<'rebase>(
+        &'rebase self,
+        index: usize,
+    ) -> Option<RebaseOperation<'repo>> {
```
alexcrichton: Should `'repo` here be `'rebase`?

src/rebase.rs
```diff
@@ -0,0 +1,492 @@
...
+            } else {
+                Some(index as usize)
+            }
+        }
+    }
+
+    /// Convenience function to get the operation at the current index.
+    ///
+    /// Returns `None` if the first operation has not yet been applied.
+    pub fn current_operation<'rebase>(&'rebase self) -> Option<RebaseOperation<'repo>> {
```
alexcrichton: Similar to above, should `'repo` be `'rebase`?

src/rebase.rs
```diff
@@ -0,0 +1,492 @@
...
+    ///
+    /// If the operation is one that applies a patch (which is any operation
+    /// except GIT_REBASE_OPERATION_EXEC) then the patch will be applied and
+    /// the index and working directory will be updated with the changes. If
+    /// there are conflicts, you will need to address those before committing
+    /// the changes.
+    ///
+    /// Iterating over the result of `operations()` correctly handles begin and
+    /// end conditions for you.
+    pub fn next<'rebase>(&'rebase mut self) -> Result<RebaseOperation<'repo>, Error> {
```
alexcrichton: Should `'repo` here be `'rebase`?

src/repo.rs
```diff
@@ -2051,6 +2052,51 @@ impl Repository {
...
+            try_call!(raw::git_rebase_init(&mut ret,
+                                           self.raw(),
+                                           from_branch_raw,
+                                           upstream_raw,
+                                           to_branch_raw,
+                                           options_raw.as_ref().map(|opts| opts as *const _)));
+            drop(from_branch);
+            drop(upstream);
+            drop(to_branch);
+            drop(options_raw);
```
alexcrichton: Are the explicit `drop` calls here required?

src/rebase.rs
```diff
@@ -0,0 +1,492 @@
...
+///
+/// [2]: https://libgit2.github.com/libgit2/#HEAD/type/git_rebase_operation
+///
+/// It has a lifetime of the rebase it belongs to.
+pub struct RebaseOperation<'repo> {
+    raw: *mut raw::git_rebase_operation,
+    _marker: marker::PhantomData<Rebase<'repo>>,
+}
+
+pub struct RebaseOperations<'repo> {
```
alexcrichton: I think this `'repo` may want to be renamed to `'rebase`

src/rebase.rs
```diff
@@ -0,0 +1,492 @@
...
+    raw: *mut raw::git_rebase,
+    _marker: marker::PhantomData<&'repo Repository>,
+}
+
+/// A structure representing a single [operation][2] to be performed during the rebase.
+///
+/// [2]: https://libgit2.github.com/libgit2/#HEAD/type/git_rebase_operation
+///
+/// It has a lifetime of the rebase it belongs to.
+pub struct RebaseOperation<'repo> {
```
alexcrichton: I think this `'repo` may want to be renamed to `'rebase`

quodlibetor:
> If that's the case I think we should remove Rebase::next as well as Rebase::current_operation. Then instead just have Rebase::operations be an iterator that exclusively uses operation_at_index.
>
> Does that make sense? Or is that not how you think Rebase is supposed to be used?

Right, from my initial reading I think that you are required to call `git_rebase_next` to advance all the state files that git itself keeps on disk, otherwise you just get into a corrupted state. Also, I think that it's multiprocess safe to just do the next operation in a rebase without asking the underlying git state files what the next thing is supposed to be.

I will add a test that verifies my hypotheses, though.

quodlibetor:
And thanks for taking the time to review this!

alexcrichton:
Oh ok I see, so if you're purely doing in-memory things `git_rebase_next` isn't very useful but if you want to actually touch the filesystem (like an interactive CLI tool for example) you'd want `next`?

If that's the case then I think we should keep the iterator and the `next` method, but I think the iterator shouldn't use `next` and should instead only use the in-memory methods perhaps?

quodlibetor:
Hey, sorry I've been busy and expect to be for a couple more weeks. If anybody wants to take this and work on it I won't be upset. I do expect to come back to it some time in the next ~month, though.

alexcrichton:
I'm gonna close this since it's been inactive for some time, but it can of course be resubmitted!

PR #318, opened 2018-05-30
Update libgit2 to include fixes for LibreSSL 2.7

heirecka:
I did not update it to a even newer version to no import that many
changes. Nevertheless it contains a few fixes, for a memleak
amongst other things.

### DIFF
--- a/systest/build.rs
+++ b/systest/build.rs
@@ -30,6 +30,8 @@ fn main() {
             "git_push_transfer_progress" |
             "git_push_negotiation" |
             "git_packbuilder_progress" => true,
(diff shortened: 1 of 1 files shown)


heirecka:
cargo build && cargo check && cargo test passed successfully locally with LibreSSL 2.7.3

alexcrichton:
Thanks! Looks like though some checks may be failing with a type changing in libgit2?

heirecka:
Thanks for fixing the test and the new release. Would you mind publishing 0.7.3 on crates.io as well?

PR #331, opened 2018-07-16
Add the ssh_key_from_memory optional feature.

promethe42:
Fix for https://github.com/alexcrichton/git2-rs/issues/320

This is done through a feature that will enable the `GIT_SSH_MEMORY_CREDENTIALS` flag when building `libgit2-sys`.

This feature is optional so that the current behavior is unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ default = ["ssh", "https", "curl"]
 ssh = ["libgit2-sys/ssh"]
 https = ["libgit2-sys/https", "openssl-sys", "openssl-probe"]
 curl = ["libgit2-sys/curl"]
+ssh_key_from_memory = ["libgit2-sys/ssh_key_from_memory"]
 
 [workspace]
 members = ["systest", "git2-curl"]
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! It looks like though this is defined by libgit2's build system if it detects a symbol in libssh2, and maybe the detection is broken? Are you sure the libssh2 build you're using has the `libssh2_userauth_publickey_frommemory` symbol?

promethe42:
> Are you sure the libssh2 build you're using has the libssh2_userauth_publickey_frommemory symbol?

When I enable the feature, everything works as expected.
So I'm guessing it does, yes.

alexcrichton:
In that case perhaps this should be enabled unconditionally? Assuming that we put it on ourselves to find an appropriate libssh2 build?

promethe42:
> In that case perhaps this should be enabled unconditionally? Assuming that we put it on ourselves to find an appropriate libssh2 build?

Maybe. I made it optional because I don't know what would happen if someone uses it with a system libssh2 that does not support it.

In my case I'm also using the `ssh2` crate. So I'm guessing the system's libssh2 is not used when I use `git2-rs`?

alexcrichton:
Hm yeah the API there seems to have been added 4 years ago, so it seems fine to unconditionally enable this.

promethe42:
> Hm yeah the API there seems to have been added 4 years ago, so it seems fine to unconditionally enable this.

Does that mean I remove the cargo "feature" thing, or make it enabled by default?

alexcrichton:
Indeed!

promethe42:
@alexcrichton Here you go!

PR #333, opened 2018-07-19
Added missing is_wt_deleted method

haxelion:
While using this awesome binding I noticed that method ``is_wt_deleted`` is missing for the ``Status`` struct. So here's a patch for it.


alexcrichton:
Thanks!

alexcrichton:
Thanks!

PR #335, opened 2018-07-20
work around for windows path separator for is_path_ignored

### LINKED ISSUE #334: is_path_ignored not working on windows
steps to repro:
start a new cargo project `C:\> cargo new git2-test` then
check that `.gitignore` includes `/target`
add `git2 = "*"` to `Cargo.toml` and the following to `main.rs`

```rust
extern crate git2;

fn main() {
    let p_exe = std::env::current_exe().unwrap();
    println!("./target/debug == {:?}", p_exe.display());
    let repo = git2::Repository::discover(&p_exe).unwrap();
    println!("./.git == {:?}", repo.path().display());
    let ignored = repo.is_path_ignored(&p_exe).unwrap();
    println!("true == {:?}", ignored);
}
```

(or unzip the file [git2-test.zip](https://github.com/alexcrichton/git2-rs/files/2211357/git2-test.zip)) and do `cargo run`.

output is:
```cmd
C:\git2-test>cargo run
   Compiling git2-test v0.1.0 (file:///C:/git2-test)
    Finished dev [unoptimized + debuginfo] target(s) in 1.78s
     Running `target\debug\git2-test.exe`
./target/debug == "C:\\git2-test\\target\\debug\\git2-test.exe"
./.git == "C:/git2-test/.git/"
true == false

```
I expected the last line to be `true == true`

tested on windows 10 locally, reduced from appveyor testing https://github.com/rust-lang/cargo/pull/5733

Eh2406:
This is the workaround requested in #334 and therefore closes #334. When this is pushed we can proceed on https://github.com/rust-lang/cargo/pull/5733.

I wish there was a way to do this that did not involve two extra allocations. `to_string_lossy` and `replace`.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2043,6 +2043,20 @@ impl Repository {
     }
 
     /// Test if the ignore rules apply to a given path.
+    #[cfg(windows)]
+    pub fn is_path_ignored<P: AsRef<Path>>(&self, path: P) -> Result<bool, Error> {
+        // `git_ignore_path_is_ignored` dose not work with windows path separator
+        // so we convert \ to /
+        let path = try!(::std::ffi::CString::new(path.as_ref().to_string_lossy().replace('\\', "/")));
+        let mut ignored: c_int = 0;
+        unsafe {
+            try_call!(raw::git_ignore_path_is_ignored(&mut ignored, self.raw, path));
+        }
+        Ok(ignored == 1)
+    }
+
+    /// Test if the ignore rules apply to a given path.
+    #[cfg(not(windows))]
     pub fn is_path_ignored<P: AsRef<Path>>(&self, path: P) -> Result<bool, Error> {
         let path = try!(path.as_ref().into_c_string());
         let mut ignored: c_int = 0;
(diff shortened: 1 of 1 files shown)


src/repo.rs
```diff
@@ -2540,4 +2554,18 @@ mod tests {
         let _ = repo.clear_ignore_rules();
         assert!(!repo.is_path_ignored(Path::new("/foo")).unwrap());
     }
+
+    #[test]
+    #[cfg(windows)]
```
alexcrichton: Could the be listed below with an `if cfg!(windows)` guard instead?

src/repo.rs
```diff
@@ -2043,6 +2043,20 @@ impl Repository {
     }
 
     /// Test if the ignore rules apply to a given path.
+    #[cfg(windows)]
```
alexcrichton: Instead of `#[cfg]`, could `if cfg!(windows)` be used to conditionally do the `to_string_lossy` instead?

Eh2406:
Changes made.

Also it looks like bild tymes are rather high, and rls reports significant time in `build_script_build`. Is it possible that a build script is missing a `rerun-if-changed` and is being run every time?

alexcrichton:
Looks great, thanks! And yeah I'd be fine tweaking the build script if necessary

PR #339, opened 2018-07-31
Make some patch functions only require &self

bbqsrc:
I checked the documentation for libgit2 and could see no reason these two functions required `&mut self`. Without this patch it is very practically difficult to iterate through all hunks of a patch, and each line of each hunk, as they rely on the same parent `git2::Patch`.


alexcrichton:
thanks!

Issue #340, opened 2018-08-02
more path not working on windows


Eh2406:
This is a follow up to #334. Are there other methods that require the same fix? https://github.com/rust-lang/cargo/pull/5820#issuecomment-409021860 suggests that `status_file()` may need it? Are there others? https://github.com/rust-lang/cargo/pull/5848 is waiting on getting this fixed, but I am not sure for witch methods?

How can we track down all of them, and make sure this doesn't come back?

PR #341, opened 2018-08-02
work around for windows path separator for status_file

Eh2406:
part of #340. Suggested by https://github.com/rust-lang/cargo/pull/5820#issuecomment-409021860


alexcrichton:
👍 Thanks!

PR #344, opened 2018-08-11
Add binding for index_add_frombuffer (already existed in libgit2-sys)

knewt:
After starting to experiment with git2-rs (thanks so much for writing this btw), discovered there wasn't an existing binding for index_add_frombuffer, so thought I'd have a go at adding it


alexcrichton:
Awesome, thanks!

PR #346, opened 2018-08-13
Don't panic in shallow clones

kornelski:
Attempt to navigate history of a repo with a shallow clone causes `panic!` inside the iterator, which can't be handled by the application.


alexcrichton:
Hm interesting! This'll be good to log for the next major release as these probably want to become iterators of results, but otherwise looks great, thanks!

PR #352, opened 2018-09-12
Tree walk: bugfix

asayers:
From https://github.com/alexcrichton/git2-rs/pull/343:

> I'm seeing some strange behaviour with this feature. I have a callback which returns `TreeWalkResult::Ok` unconditionally, and I expect every node in the tree to be visited. Instead I see a subset of the top-level nodes only, and this subset is not deterministic across runs! Looks like something nasty is happening. Returning 0 from the callback instead gives the expected behaviour.

I haven't been able to observe this behaviour in this branch.

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -43,14 +43,14 @@ pub enum TreeWalkMode {
 }
 
 /// Possible return codes for tree walking callback functions.
-#[allow(dead_code)]
+#[repr(i32)]
 pub enum TreeWalkResult {
     /// Continue with the traversal as normal.
     Ok = 0,
     /// Skip the current node (in pre-order mode).
     Skip = 1,
     /// Completely stop the traversal.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

asayers:
`Tree::walk()` should now be safe, but `treewalk_cb` is unsafe.  However, I can't _mark_ it as `unsafe`, because `raw::git_tree_walk` expects a safe function.  Options:

* Document the invariant and cross fingers.
* Monomorphise `tree::Walk` and force the user to return `TreeWalkResult`.
* Some way to mark `treewalk_cb` as `unsafe` that I'm not thinking of?

alexcrichton:
Ah that's ok, it's an internal function so it should be fine to leave as "safe", but adding documentation would also be great!

PR #353, opened 2018-09-13
Remove dependency on `cmake`

alexcrichton:
I've spent too many hours of my life bending over backwards trying to satisfy
native libraryies' build systems. The most recent pain is that informing a
native build system of its dependencies (such as telling libgit2 where libssh2
is installed) is an absolute never-ending nightmare.

The towel is now thrown in as `cmake` is jettisoned and this is now just using
the `cc` crate to directly compile all the various C code.

For some more info see alexcrichton/curl-rust#225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,13 +42,3 @@ env:
 notifications:
   email:
     on_success: never
-
-addons:
-  apt:
-    sources:
-      - kalakris-cmake
-    packages:
-      - cmake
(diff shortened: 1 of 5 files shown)


alexcrichton:
Merged in https://github.com/alexcrichton/git2-rs/commit/76f4b74aef2bc2a54906ddcbf7fbe0018936a69d

PR #355, opened 2018-09-21
Bring back MinGW CI

mati865:
With latest MinGW fix and this workaround applied build and tests work but it takes about [16 minutes for x86_64-gnu toolchain to finish](https://ci.appveyor.com/project/mati865/git2-rs/build/1.0.18/job/ldhp2vdy2be13815). One of the gnu toolchains can be dropped to save time.

Nightly x86_64-gnu doesn't work at all so it's temporary using beta to see what you think about it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,8 +9,8 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
   - TARGET: i686-pc-windows-msvc
 install:
-  - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-nightly-${env:TARGET}.exe"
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I'm hesitant to add this though as MinGW has historically been incredibly difficult to keep working on CI with very little benefit. It would have caught a regression like the MinGW toolchain not working, but that was altogether a pretty simple regression to fix and one that happens vanishing rarely (and was discovered quite quickly). 

To that end I'd prefer to not add this for now, but it's good to have the knowledge here of how to get it working for other projects!

mati865:
No problem, I'll keep the branch around so it's easier to access it.

PR #358, opened 2018-10-04
Fix documentation for Sort::TOPOLOGICAL

flodiebold:
The libgit2 documentation says "no parents before all of its children are
shown", which is exactly the opposite of what the rust docstring said.

This got me quite confused for a bit :smile:


alexcrichton:
Thanks!

PR #360, opened 2018-10-11
Support sticky redirects in git2-curl.

ehuss:
If you use a gitlab url such as "https://gitlab.com/user/repo", then gitlab will
redirect the first GET request to "https://gitlab.com/user/repo.git" (with the
`.git` at the end). However, gitlab does not redirect the next POST, causing a
404. This change keeps track of the original redirect so that subsequent actions
use the new base url. This roughly mirrors what is done in libgit2's transport.

Fixes rust-lang/cargo#6114

### DIFF
--- a/git2-curl/src/lib.rs
+++ b/git2-curl/src/lib.rs
@@ -36,13 +36,18 @@ use url::Url;
 
 struct CurlTransport {
     handle: Arc<Mutex<Easy>>,
+    /// The URL of the remote server, e.g. "https://github.com/user/repo"
+    ///
+    /// This is an empty string until the first action is performed.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh jeez, that really is pretty icky! I think this is good enough for now though. I may have been a bit too conservative in requiring things to be threadsafe, and that'd make this much nicer with interior mutability via `RefCell` rather than mutexes. In any case, thanks so much for investigating this!

alexcrichton:
I've included this in 0.8.2 and folded the version bump into https://github.com/rust-lang/rust/pull/54919

Issue #365, opened 2018-10-22
Segmentation fault while checking out bare repository with notifications


minijackson:
I don't know if the problem comes from `git2` or `libgit2` itself, but here's the reduced code:

```rust
extern crate git2;

use git2::build::{CheckoutBuilder, RepoBuilder};
use git2::{CheckoutNotificationType, Repository};

fn main() {
    let repo = RepoBuilder::new()
        .bare(true)
        .clone(
            "https://github.com/alexcrichton/git2-rs",
            "/tmp/segfault_poc_bare".as_ref(),
        )
        .unwrap();

    let tree = repo
        .revparse_single(&"master")
        .unwrap()
        .peel_to_tree()
        .unwrap();
    let mut index = repo.index().unwrap();
    index.read_tree(&tree).unwrap();

    let mut checkout_opts = CheckoutBuilder::new();
    checkout_opts.target_dir("/tmp/segfault_poc_checkout".as_ref());
    checkout_opts.notify_on(CheckoutNotificationType::all());
    checkout_opts.notify(|notif, path, file1, file2, file3| {
        println!("{:?}", file1.path());
        true
    });
    repo.checkout_index(Some(&mut index), Some(&mut checkout_opts))
        .unwrap();
}
```

It seems that trying to print either `file1.path()` or `file3.path()` in the notify callback produces the segfault (but not `file2` for some reason).

Doing the same with a non-bare repository seems to work fine.

Tested with Rust 1.31.0-nightly and 1.27.0 (no difference).

Thank you for the great work!

minijackson:
After reading the source code for libgit2 ([here in particular](https://github.com/libgit2/libgit2/blob/master/src/checkout.c#L94-L152)) I've found that each of the diff file pointers passed to the callback have cases where they can be null. One way to fix the segfault would be to change the notify callback in git2 to take `Option<DiffFile>` instead of `DiffFile` as arguments. What do you think?

alexcrichton:
Sounds like a reasonable change to me!

minijackson:
Closing since it was fixed in linked PR

PR #366, opened 2018-10-23
Use Option for DiffFiles in CheckoutBuilder's notify callback

minijackson:
Fixes issue #365 

With a regression test for good measure


alexcrichton:
Thanks!

PR #376, opened 2018-11-17
Expose openssl vendored feature

rukai:
I'm developing on arch and deploying to ubuntu and get this error when I run the binary.
```
./pf_sandbox_website: /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1: version `OPENSSL_1_1_1' not found (required by ./pf_sandbox_website)
```
I was able to resolve the issue by statically linking as per this PR.

Is exposing it like this the correct solution?
What about `libgit2-sys`? I didn't need to change anything in its `cargo.toml` even though it uses `openssl-sys` as well.


alexcrichton:
Thanks for the PR! Since this only affects OpenSSL could the feature be renamed to something like `vendored-openssl` perhaps?

rukai:
Thanks for the feedback!

PR #382, opened 2018-12-07
Introduce rebases

carlosmn:
I've tried to follow the conventions and patterns from the rest of the library. It includes iterator for going through the operations in order.

I only ralised #315 existed after writing this but that PR is stalled/abandoned so I guess this doesn't hurt too much.


carlosmn:
Huh, maybe I messed up some signatures, but I can run that systest command just fine locally on my rustc 1.28

carlosmn:
Nope nevermind, I can repro locally, I was just on the wrong branch before.

alexcrichton:
Ah good point! The `ctest` crate in use here has various options to configure how it generates tests, but it can definitely be a bit of a black box. I think a diff like this could help fix CI though?

https://gist.github.com/alexcrichton/8a57d317c4f1715bd7de6f8ca7662f5b

carlosmn:
Since `kind` is already known to transform to `type`, we can avoid adding an extra rule rule and awkward name by renaming the field to `kind` which is what we use for the accessor anyway; that way we only add the special handling for `id`.

alexcrichton:
Perfect, thanks!

PR #385, opened 2018-12-13
Implement wrapper around git_index_conflict_iterator.

iancormac84:


alexcrichton:
One option would be to add instance methods to return the struct fields, but in this case it's perhaps best to simply have them be public fields?

iancormac84:
Thanks! It wasn't registering to me that the fields were private, and therefore essentially useless.

iancormac84:
I'm not sure how to fix the errors that are showing up on the Travis CI.

PR #387, opened 2018-12-15
Implemented wrapper for git_index_conflict_iterator.

iancormac84:


alexcrichton:
Hm oddly enough it looks like the signatures match on windows, but not on unix? (causing travis to fail)

PR #389, opened 2018-12-21
Adds git attributes functionality. 

Kleptine:
This is my first venture into anything unsafe/c-bindings, so if you are able to take a close eye on that part, that'd be fantastic. I've tested it a fair bit, but it's entirely possible I'm doing things incorrectly. Thanks!

This only contains git_attr_get for now, but sets up the types for others to be implemented.

---
This change is [Reviewable](https://reviewable.io/reviews/alexcrichton/git2-rs/389)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,6 +80,7 @@ use std::fmt;
 use std::str;
 use std::sync::{Once, ONCE_INIT};
 
+pub use attr::{AttributeType};
 pub use blame::{Blame, BlameHunk, BlameIter, BlameOptions};
 pub use blob::{Blob, BlobWriter};
 pub use branch::{Branch, Branches};
@@ -626,6 +627,36 @@ impl MergePreference {
     is_bit_set!(is_fastforward_only, MergePreference::FASTFORWARD_ONLY);
 }
 
+
+bitflags! {
+    /// Check attribute flags: Reading values from index and working directory.
+    ///
+    /// When checking attributes, it is possible to check attribute files
+    /// in both the working directory (if there is one) and the index (if
+    /// there is one).  You can explicitly choose where to check and in
+    /// which order using the following flags
+    ///
+    /// Core git usually checks the working directory then the index,
+    /// except during a checkout when it checks the index first.  It will
+    /// use index only for creating archives or for a bare repo (if an
+    /// index has been specified for the bare repo).
+    pub struct CheckAttributeFlags: u32 {
+        #[allow(missing_docs)]
+        const FILE_THEN_INDEX = raw::GIT_ATTR_CHECK_FILE_THEN_INDEX as u32;
+
+        #[allow(missing_docs)]
+        const INDEX_THEN_FILE = raw::GIT_ATTR_CHECK_INDEX_THEN_FILE as u32;
+
+        #[allow(missing_docs)]
+        const INDEX_ONLY = raw::GIT_ATTR_CHECK_INDEX_ONLY as u32;
+
+        /// Normally, attribute checks include looking in the /etc (or system
+        /// equivalent) directory for a `gitattributes` file.  Passing this
+        /// flag will cause attribute checks to ignore that file.
+        const NO_SYSTEM = raw::GIT_ATTR_CHECK_NO_SYSTEM as u32;
+    }
+}
+
 #[cfg(test)] #[macro_use] mod test;
 #[macro_use] mod panic;
 mod call;
@@ -637,6 +668,7 @@ pub mod string_array;
 pub mod oid_array;
 pub mod transport;
 
+mod attr;
 mod blame;
 mod blob;
 mod branch;
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -10,8 +10,9 @@ use libc::{c_int, c_char, size_t, c_void, c_uint};
 use {raw, Revspec, Error, init, Object, RepositoryOpenFlags, RepositoryState, Remote, Buf, StashFlags};
 use {ResetType, Signature, Reference, References, Submodule, Blame, BlameOptions};
 use {Branches, BranchType, Index, Config, Oid, Blob, BlobWriter, Branch, Commit, Tree};
-use {AnnotatedCommit, MergeOptions, SubmoduleIgnore, SubmoduleStatus, MergeAnalysis, MergePreference};
+use {AnnotatedCommit, MergeOptions, SubmoduleIgnore, SubmoduleStatus, MergeAnalysis, MergePreference, CheckAttributeFlags};
 use {ObjectType, Tag, Note, Notes, StatusOptions, Statuses, Status, Revwalk};
+use {AttributeType};
 use {RevparseMode, RepositoryInitMode, Reflog, IntoCString, Describe};
 use {DescribeOptions, TreeBuilder, Diff, DiffOptions, PackBuilder, Odb};
 use {Rebase, RebaseOptions};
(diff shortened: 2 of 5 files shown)


src/lib.rs
```diff
@@ -626,6 +627,36 @@ impl MergePreference {
...
+    /// When checking attributes, it is possible to check attribute files
+    /// in both the working directory (if there is one) and the index (if
+    /// there is one).  You can explicitly choose where to check and in
+    /// which order using the following flags
+    ///
+    /// Core git usually checks the working directory then the index,
+    /// except during a checkout when it checks the index first.  It will
+    /// use index only for creating archives or for a bare repo (if an
+    /// index has been specified for the bare repo).
+    pub struct CheckAttributeFlags: u32 {
```
Kleptine: I'm not honestly sure exactly how these flags behave for libgit2, I think it's for how it decides which .gitattributes files to use, but it's a bit unclear, so I haven't added unit tests for them.

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
         }
         Ok(ignored == 1)
     }
+
+    /// Look up the value of one git attribute for path. Equivalent to libgit2::git_attr_get.
+    pub fn get_attr(&self, flags: CheckAttributeFlags, path: &str, name: &str) -> Result<AttributeType, Error> {
+        let mut value_out : *mut c_char = 0 as *mut c_char;
```
Kleptine: I tried to figure out how to use ptr::null but I couldn't get the type quite right. :\

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
...
+        let value_out_ptr : *mut *mut c_char = &mut value_out; 
+        
+        let c_flags = flags.bits() as c_uint;
+        let c_path = CString::new(path).expect("The path value contained an invalid 0 byte within the string");
+        let c_name = CString::new(name).expect("The name value contained an invalid 0 byte within the string");
+
+        let attribute_type = unsafe {
+            try_call!(raw::git_attr_get(value_out_ptr, self.raw, c_flags, c_path, c_name));
```
Kleptine: Is try_call the correct macro here? I borrowed it from the other usages, but I'm not entirely sure what it provides. It seems to do some auto-conversions?

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
...
+            let attr_type = raw::git_attr_value(*value_out_ptr);
+
+            let attribute_type = match attr_type {
+                raw::GIT_ATTR_UNSPECIFIED => AttributeType::Unspecified,
+                raw::GIT_ATTR_TRUE => AttributeType::True,
+                raw::GIT_ATTR_FALSE => AttributeType::False,
+                raw::GIT_ATTR_VALUE => {
+                    let value = CStr::from_ptr(*value_out_ptr);
+                    let string_value = value.to_string_lossy().to_owned().to_string();
+                    AttributeType::Value(string_value)
```
Kleptine: For the returned attribute string value, I've just copied it out lossily to a rust string. It might be possible to hold a borrow, but that seems like a pretty big pain, and I'm not entirely sure how long that memory is valid anyhow.

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
...
+            try_call!(raw::git_attr_get(value_out_ptr, self.raw, c_flags, c_path, c_name));
+
+            let attr_type = raw::git_attr_value(*value_out_ptr);
+
+            let attribute_type = match attr_type {
+                raw::GIT_ATTR_UNSPECIFIED => AttributeType::Unspecified,
+                raw::GIT_ATTR_TRUE => AttributeType::True,
+                raw::GIT_ATTR_FALSE => AttributeType::False,
+                raw::GIT_ATTR_VALUE => {
+                    let value = CStr::from_ptr(*value_out_ptr);
```
Kleptine: Do I need to do anything with a CStr (ie. free it manually) or will this just take care of itself?

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
...
         Ok(ignored == 1)
     }
+
+    /// Look up the value of one git attribute for path. Equivalent to libgit2::git_attr_get.
+    pub fn get_attr(&self, flags: CheckAttributeFlags, path: &str, name: &str) -> Result<AttributeType, Error> {
+        let mut value_out : *mut c_char = 0 as *mut c_char;
+        let value_out_ptr : *mut *mut c_char = &mut value_out; 
+        
+        let c_flags = flags.bits() as c_uint;
+        let c_path = CString::new(path).expect("The path value contained an invalid 0 byte within the string");
```
alexcrichton: I think the `expect` here can be replaced with `?` perhaps?

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
         }
         Ok(ignored == 1)
     }
+
+    /// Look up the value of one git attribute for path. Equivalent to libgit2::git_attr_get.
+    pub fn get_attr(&self, flags: CheckAttributeFlags, path: &str, name: &str) -> Result<AttributeType, Error> {
+        let mut value_out : *mut c_char = 0 as *mut c_char;
+        let value_out_ptr : *mut *mut c_char = &mut value_out;
```
alexcrichton: I think you can avoid this pointer and just use `&mut` in `try_call!` below?

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
         }
         Ok(ignored == 1)
     }
+
+    /// Look up the value of one git attribute for path. Equivalent to libgit2::git_attr_get.
+    pub fn get_attr(&self, flags: CheckAttributeFlags, path: &str, name: &str) -> Result<AttributeType, Error> {
```
alexcrichton: Could this perhaps be split into two APIs? One that returns an opaque `Attribute`, and then an API on `Attribute` to interpret what type it is?

src/repo.rs
```diff
@@ -2100,6 +2101,40 @@ impl Repository {
...
+
+            let attr_type = raw::git_attr_value(*value_out_ptr);
+
+            let attribute_type = match attr_type {
+                raw::GIT_ATTR_UNSPECIFIED => AttributeType::Unspecified,
+                raw::GIT_ATTR_TRUE => AttributeType::True,
+                raw::GIT_ATTR_FALSE => AttributeType::False,
+                raw::GIT_ATTR_VALUE => {
+                    let value = CStr::from_ptr(*value_out_ptr);
+                    let string_value = value.to_string_lossy().to_owned().to_string();
```
alexcrichton: I think we'll want to avoid the `_lossy` here, and if we go with a dedicated `Attribute` API then we can have a bytes and optional `str` accessor like most other types

src/repo.rs
Kleptine: Thanks! 

Part of this was with the goal to make the API a bit safer (ie. representing the implicit enum of `git_attr_t` explicit). The C API returns something quite close to an enum, so it seems like an enum makes a lot of sense as a return value. 

Do you think it'd be better to keep it closer to the original API here? It would allow you to squeak out a bit more performance, perhaps?

Either way, I'll swap the to_string_lossy to a &[u8] for the value itself, and rename AttributeType to Attribute. 

I added some comments above as well, just general questions -- let me know if there's a better place to add that kind of stuff.

I'm following along from the header file [here](https://github.com/libgit2/libgit2/blob/master/include/git2/attr.h).

src/repo.rs
Kleptine: Ah, moving to just using a &mut reference solves this issue too.  :)

src/repo.rs
alexcrichton: It would yeah, although I figured it'd be best to return "rusty" types like `&[u8]` and avoid showing *too* much of the underlying C. At this point it's probably best to stick with the rest of the conventions, but another future update could switch it all to `CStr`!

src/repo.rs
alexcrichton: Ah I'm not too too worried about performance here but more idioms and following the original documentation. Most of this library is written assuming you're also following along with the documentation of upstream libgit2 (as it's far more comprehensive), so having the same idioms and styles of APIs I think are important for doing the translation.

Kleptine: I've been working on splitting the API up into two, but ran into a (potential?) safety problem.

The attribute pointer returned by the C implementation can only be read as byte string if the attribute is of type value. If we expose an `as_bytes` function on the opaque Attribute, we would first have to make the check to verify that it does indeed have the `git_attr_t` type of Value. 

The LibGit2 documentation is a bit unclear on the behaviour of trying to read a non-value attribute as a string. It seems undefined.

Does that seem reasonable? It defeats the purpose of splitting the API on a procedural level, although I'm happy to keep with this strategy, as it does make sure to stick with the documentation. 

Thanks for the help!

alexcrichton: Ah ok then yeah in that case we'd probably want it to return an `Option` and then it'd get checked  when the bytes are accessed

alexcrichton:
I'm gonna close this since it's been inactive for some time now, but it can of course be resubmitted!

PR #402, opened 2019-02-19
Don't quote absolute path helper commands

hmarr:
Quoting treats the full command as the path to binary, which prevents arguments from being provided.

According to the [Git documentation][git-docs] arguments should be accepted. This also [matches the behaviour][git-source] of git itself.

I discovered this issue through Cargo's use of this library.

[git-docs]: https://git-scm.com/docs/api-credentials#_credential_helpers
[git-source]: https://github.com/git/git/blob/ca1b4116483b397e78483376296bcd23916ab553/credential.c#L258-L261


alexcrichton:
Thanks!

PR #406, opened 2019-02-27
Fixing misspelling

damiancarrillo:
This just fixes a misspelled word ("user" instead of "uses") in the documentation.


ehuss:
Thanks!

PR #407, opened 2019-03-07
Add Status::is_current convenience method for completeness

joliss:
Now we have convenience methods for [all the constants](https://github.com/rust-lang/git2-rs/blob/718799c265d4e6fea635906cb57379a441021f3c/src/lib.rs#L922-L950).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -952,6 +952,7 @@ bitflags! {
 }
 
 impl Status {
+    is_bit_set!(is_current, Status::CURRENT);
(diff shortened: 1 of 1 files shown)


joliss:
Wait, this is not correct. GIT_STATUS_CURRENT is just 0; it denotes the absence of any set bits. Let me see if I can fix this.

alexcrichton:
I'm gonna close this since it's been inactive for some time now, but it can of course be resubmitted!

PR #409, opened 2019-03-10
Amended IndexConflict to make it sensitive to the possibility that gi…

iancormac84:
…t_index_conflict_iterator returns a NULL pointer for one of the trio of IndexEntries. Also added Reference::find_reference_dwim to make it a little easier to find refs.


iancormac84:
Are any changes needed to make this ready for merge?

alexcrichton:
The changes here look fine to me yeah but it's a breaking API change so I'd prefer that these were batched up (although I'm not certain when the batch would happen)

alexcrichton:
👍

PR #415, opened 2019-04-26
Fix incorrect method name in Branch doc.

DanAlbert:
There is no `unwrap`. The method is called `into_reference`.


alexcrichton:
Thanks and sorry for the delay!

PR #417, opened 2019-05-10
use ? instead of try! macro

foamdino:
Sorry to bug you, I was looking into the example code to learn how to use `revwalk` and noticed that IntelliJ was warning about deprecated `try!` macro.


alexcrichton:
Thanks!

PR #423, opened 2019-05-26
Refactor: Extract function util::c_cmp_to_ordering

emlun:
I noticed some repeated code, and also added a test for good measure.

### DIFF
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -181,6 +181,7 @@ mod tests {
 
     use tempdir::TempDir;
     use {ObjectType};
+    use super::Error;
     use super::Oid;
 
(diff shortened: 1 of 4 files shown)


ehuss:
Thanks!

emlun:
You're welcome!

By the way, I also experimented a bit with adding a `trait OrdC { fn c_cmp(...) -> c_int }` and an `impl<T: OrdC> Ord for T`, but it seems like that's not allowed by Rust's orphan rules (it would work if `Ord` had been defined in this crate). If you have any ideas on how something like that could be doable, I'd be happy to experiment a bit more with it to see what happens.

PR #425, opened 2019-05-31
Update to libgit2 0.28.2

cuviper:
ref: [Upstream changelog](https://github.com/libgit2/libgit2/blob/maint/v0.28/docs/changelog.md)

This includes a semver bump for all crates, as a few APIs have been
renamed in 0.28. Upstream provides `deprecated.h` aliases, if we'd like
to follow suit, but I feel it's better to just move forward.

- `giterr_*` -> `git_error_*`
- `GITERR_*` -> `GIT_ERROR_*`
- `git_otype` -> `git_object_t`
- `GIT_OBJ_*` -> `GIT_OBJECT_*`
- `git_ref_t` -> `git_reference_t`
- `GIT_REF_*` -> `GIT_REFERENCE_*`
- `git_idxentry_*` -> `git_index_entry_*`
- `GIT_IDXENTRY_*` -> `GIT_INDEX_ENTRY_*`
- `git_buf_free` -> `git_buf_dispose`

Proxy support is now builtin to `libgit2`, so `libcurl` is no longer a
dependency at all. Proxies were also the reason for `git2-curl` to
exist, which seems to be unnecessary for a while now, since `libgit2`
used `libcurl` itself since 0.23 anyway. Maybe `git2-curl` can just be
retired now, but I've updated it just for the sake of compatibility.

The minimum pkg-config libgit2 is now documented and tested for 0.28.
While it previously claimed support for 0.25.1, this was broken by
`git_config_entry`'s `include_depth` field, which was added *after* the
0.27 release branched.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "libgit2-sys/libgit2"]
 	path = libgit2-sys/libgit2
 	url = https://github.com/libgit2/libgit2
+	branch = maint/v0.28
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "git2"
-version = "0.8.0"
+version = "0.9.0"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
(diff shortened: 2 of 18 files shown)


ehuss:
This looks great, thanks for doing this!  I'll try to look at it closer sometime soon.

@alexcrichton Do you have any idea why the azure pipeline didn't launch?  It was working before.

ehuss:
I think git2-curl will still be needed by things like Cargo.  Cargo performs a variety of configurations on the curl handle (certificate settings, user-agent, debugging, etc.), and those don't seem to have direct analogues in libgit2.

I've been testing this out a bit and I haven't found any problems.

alexcrichton:
Thanks for the @cuviper! Agreed with @ehuss that we do use the curl support for non-proxy stuff, but it's indeed minor (ish) compared to the proxy support which was critical at the time! In any case looks like this is all ship shape, so I'm gonna merge.

alexcrichton:
And now published on crates.io!

PR #426, opened 2019-06-02
Fix `bare_trait_objects` warning.

ehuss:
`bare_trait_objects` is now a warning on nightly (via https://github.com/rust-lang/rust/pull/61203). This is required to get tests passing on nightly.


alexcrichton:
👍

PR #427, opened 2019-06-03
A first stab at a pull example.

zaphar:
I'm sure I'm missing edge cases and stuff in here since I had to basically piece this together over a couple weeks but since I've done it I figured other people could learn from my now hard won experience.

### DIFF
--- a/examples/pull.rs
+++ b/examples/pull.rs
@@ -153,7 +153,7 @@ fn do_merge<'a>(
             }
             Err(_) => {
                 // The branch doesn't exist so just set the reference to the
-                // commit direcly. Usually this is because you are pulling
+                // commit directly. Usually this is because you are pulling
                 // into an empty repository.
                 repo.reference(
                     &refname,
--- a/examples/pull.rs
+++ b/examples/pull.rs
@@ -12,13 +12,13 @@
  * <http://creativecommons.org/publicdomain/zero/1.0/>.
  */
 
-extern crate git2;
 extern crate docopt;
+extern crate git2;
 #[macro_use]
 extern crate serde_derive;
 
 use docopt::Docopt;
-use git2::{Repository};
+use git2::Repository;
 use std::io::{self, Write};
 use std::str;
 
@@ -38,14 +38,19 @@ fn do_fetch<'a>(
     // Print out our transfer progress.
     cb.transfer_progress(|stats| {
         if stats.received_objects() == stats.total_objects() {
-            print!("Resolving deltas {}/{}\r", stats.indexed_deltas(),
-                    stats.total_deltas());
+            print!(
+                "Resolving deltas {}/{}\r",
+                stats.indexed_deltas(),
+                stats.total_deltas()
+            );
         } else if stats.total_objects() > 0 {
-            print!("Received {}/{} objects ({}) in {} bytes\r",
-                    stats.received_objects(),
-                    stats.total_objects(),
-                    stats.indexed_objects(),
-                    stats.received_bytes());
+            print!(
+                "Received {}/{} objects ({}) in {} bytes\r",
+                stats.received_objects(),
+                stats.total_objects(),
+                stats.indexed_objects(),
+                stats.received_bytes()
+            );
         }
         io::stdout().flush().unwrap();
         true
@@ -54,7 +59,7 @@ fn do_fetch<'a>(
     let mut fo = git2::FetchOptions::new();
     fo.remote_callbacks(cb);
     fo.download_tags(git2::AutotagOption::All); // Always fetch all tags.
-    // Perform a download and also update tips
+                                                // Perform a download and also update tips
     println!("Fetching {} for repo", remote.name().unwrap());
     remote.fetch(refs, Some(&mut fo), None)?;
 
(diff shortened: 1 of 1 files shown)


zaphar:
I was going to run `cargo fmt` but it reformatted way too many files to pollute this diff with so I left it as is.

examples/pull.rs
```diff
@@ -0,0 +1,186 @@
...
+            println!("Nothing to do...");
+        }
+        Ok(())
+}
+
+fn run(args: &Args) -> Result<(), git2::Error> {
+    let remote_name = args.arg_remote.as_ref().map(|s| &s[..]).unwrap_or("origin");
+    let remote_branch = args.arg_branch.as_ref().map(|s| &s[..]).unwrap_or("master");
+    let repo = Repository::open(".")?;
+    let mut remote = repo.find_remote(&remote_name)?;
```
zaphar: Just realized that this section won't compile. I'll push out a new change later today.

zaphar:
I think this is ready for review if anyone want's to tackle it :-D It builds and runs for the limited set of cases I think I'm handling correctly.

ehuss:
Thanks @zaphar, I'll try to get to it sometime soon.

examples/pull.rs
```diff
@@ -0,0 +1,200 @@
...
+    remote_branch: &str,
+    fetch_commit: git2::AnnotatedCommit<'a>,
+) -> Result<(), git2::Error> {
+    // 1. do a merge analysis
+    let analysis = repo.merge_analysis(&[&fetch_commit])?;
+
+    // 2. Do the appopriate merge
+    if analysis.0.is_fast_forward() {
+        // do a fast forward
+        match repo.find_reference(remote_branch) {
```
ehuss: I'm having a hard time following how this should work.  In my testing, the fast-forward never works because this line fails.  Calling `find_reference` on the string `master` isn't going to work.

From my limited understanding of how `pull` should work, shouldn't the fast-forward code path do the following?
1. Get the `HEAD` ref
2. Look up `FETCH_HEAD` (or maybe not necessary, since this has `fetch_commit`).
3. repo.checkout_tree(FETCH_HEAD)
4. Move HEAD to oid of FETCH_HEAD

examples/pull.rs
```diff
@@ -0,0 +1,200 @@
...
+    lb.set_target(rc.id(), &msg)?;
+    Ok(())
+}
+
+fn normal_merge(
+    repo: &Repository,
+    local: &git2::AnnotatedCommit,
+    remote: &git2::AnnotatedCommit,
+) -> Result<(), git2::Error> {
+    let local_tree = repo.find_tree(local.id())?;
```
ehuss: This code path isn't working for me.  I am getting:

`error: the requested type does not match the type in ODB; class=Invalid (3); code=NotFound (-3)`

Since `local` is a commit, I assume `find_tree` does not dereference the commit to a tree.  I'm not sure how this should work.

examples/pull.rs
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along
+ * with this software. If not, see
+ * <http://creativecommons.org/publicdomain/zero/1.0/>.
+ */
+
+extern crate docopt;
+extern crate git2;
+#[macro_use]
+extern crate serde_derive;
```
ehuss: When you rebase on master, the project has been updated to 2018 edition, so you can drop these `extern crate` lines.  Will still need a `use serde_derive::Deserialize;`, though.

examples/pull.rs
zaphar: It is probably wrong :-D. I was very much feeling my way through the dark on these.

zaphar: Okay I've reloaded my state on this and I think my poor naming made it harder to reason through.

I could be wrong but I think the way this is supposed to work is that we want to set the reference we are pulling specifically the branch in your case 'master' to the result of the fetched commit. I should probably also be setting head as well which I don't think I'm doing but that isn't necessarily clear to me if that should happen or not.

I've updated the naming in this code to better indicate what I'm doing for clarity.

examples/pull.rs
zaphar: Hrmmm. Your right that does indeed fail. I'm not sure what the proper way to do this is actually. Looks like I have some reading to do.

examples/pull.rs
```diff
@@ -0,0 +1,200 @@
...
+            );
+        }
+        io::stdout().flush().unwrap();
+        true
+    });
+
+    let mut fo = git2::FetchOptions::new();
+    fo.remote_callbacks(cb);
+    fo.download_tags(git2::AutotagOption::All); // Always fetch all tags.
+                                                // Perform a download and also update tips
```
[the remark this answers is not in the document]
zaphar: done

examples/pull.rs
zaphar: Aha I was getting mired in the whole object db semantics. looks like the correct way is to do
`let local_tree = repo.find_commit(local.id())?.tree()?'`. I'll try to test this a little bit later when I have some more time this weekend.

zaphar: @ehuss I've tested this and it seems to work. I've also handled the conflicts on merge case.

examples/pull.rs
zaphar: I'll do that at the end so I only have to take the hit once. :-D

zaphar:
I think this is ready for another pass

zaphar:
@ehuss Ping

DarrienG:
I used this as an example and it was super helpful. Thanks :)

kallisti5:
super helpful. Why isn't this in the docs?

examples/pull.rs
```diff
@@ -0,0 +1,214 @@
...
+    };
+    let msg = format!("Fast-Forward: Setting {} to id: {}", name, rc.id());
+    println!("{}", msg);
+    lb.set_target(rc.id(), &msg)?;
+    repo.set_head(&name)?;
+    repo.checkout_head(Some(git2::build::CheckoutBuilder::default()
+       // For some reason the force is required to make the working directory actually get updated
+       // I suspect we should be adding some logic to handle dirty working directory states
+       // but this is just an example so maybe not.
+       .force()))?;
```
jder: 👋Thanks for putting this PR together; it was really helpful. 

I ran into this safe vs force problem as well and posted an answer [here](https://stackoverflow.com/a/60453573/3737) about what I think is going on. The summary is: safe mode will work if you checkout_branch before setting the head.

zaphar: Yeah that analysis sounds about right.

ehuss:
@zaphar  I apologize for letting this fall through the cracks. Would you be able to rebase on master? Afterwards I'll go ahead and merge it.

zaphar:
@ehuss, I'll try to get to that this week. Work and such has been a bit crazy with CoVID-19 preparedness and response stuff.

zaphar:
I successfully rebased and have updated to `StructOpt` which looks like it is what got chosen instead of `Docopt`. Some comletely unrelated tests failed on my local so I'm not sure if the above tests are going pass or not.

PR #432, opened 2019-06-25
Add cherrypick method on Repository object

vthib:
Hello. I saw that the previous PR to implement the cherrypick method was abandoned: #183 

I am interested in having this method exposed myself, so I have taken the existing PR and I have updated it:
* rebased, squashed and updated for master
* I have applied both of your remaining comments
* I have added a test in repo.rs
I figured it would be better to start a new PR, given the age of the previous one and the conflicts it had with current master.

Can you take a look when you have the time? Thank you :)


src/cherrypick.rs
```diff
@@ -0,0 +1,63 @@
...
+    mainline: u32,
+    checkout_builder: Option<CheckoutBuilder<'cb>>,
+    merge_opts: Option<MergeOptions>,
+}
+
+impl<'cb> CherrypickOptions<'cb> {
+    /// Creates a default set of cherrypick options
+    pub fn new() -> CherrypickOptions<'cb> {
+        CherrypickOptions {
+            mainline: 1,
```
vthib: This is wrong, should be 0 by default, otherwise will return an error when cherry-picking a non-merge commit.

src/cherrypick.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Obtain the raw struct
+    pub fn raw(&mut self) -> raw::git_cherrypick_options {
+        unsafe {
+            let mut checkout_opts: raw::git_checkout_options = mem::zeroed();
+            raw::git_checkout_init_options(&mut checkout_opts, raw::GIT_CHECKOUT_OPTIONS_VERSION);
+            if let Some(ref mut cb) = self.checkout_builder {
+                cb.configure(&mut checkout_opts);
+            }
+            let mut merge_opts = mem::zeroed();
```
vthib: a git_merge_init_options is missing

src/cherrypick.rs
```diff
@@ -0,0 +1,63 @@
...
+            if let Some(ref opts) = self.merge_opts {
+                ptr::copy(opts.raw(), &mut merge_opts, 1);
+            }
+            raw::git_cherrypick_options {
```
vthib: a call to git_cherrypick_init_options is missing too

src/cherrypick.rs
vthib: Done

src/cherrypick.rs
vthib: Done

src/cherrypick.rs
vthib: Done

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "git2"
-version = "0.9.1"
+version = "0.9.2"
```
vthib: That's a mistake, I didn't mean to include this

vthib:
I'm not sure I understand why it fails, The systests works fine for me

vthib:
Looks like the fails are not related to my PR, I can reproduce them on origin/master on a fresh ubuntu container

src/repo.rs
```diff
@@ -2346,6 +2347,23 @@ impl Repository {
...
+
+    /// Perform a cherrypick
+    pub fn cherrypick(
+        &self,
+        commit: &Commit<'_>,
+        options: Option<&mut CherrypickOptions<'_>>,
+    ) -> Result<(), Error> {
+        unsafe {
+            let raw_opts = match options {
+                Some(opts) => &opts.raw(),
```
alexcrichton: I think this needs to be moved into a temporary because otherwise this creates a raw pointer which, when used, is a use-after-free

src/cherrypick.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates a default set of cherrypick options
+    pub fn new() -> CherrypickOptions<'cb> {
+        CherrypickOptions {
+            mainline: 0,
+            checkout_builder: None,
+            merge_opts: None,
+        }
+    }
+
+    /// Set the mainline value
```
alexcrichton: I'm not really sure what this field is and the documentation here isn't too helpful, could this be expanded?

vthib: Yes! I added the comment of the field from libgit2.

src/repo.rs
vthib: Indeed, that was a mistake.

I'm not entirely familiar with Rust, but I believe I fixed it. I suppose Rust stores an Option<git_cherrypick_options> as a struct + bool, so pattern matching + take the ref of the Some value should be ok as long as it is used in the same scope as the Option.

vthib:
Thank you for the review! I forced pushed because I rebased this PR on the fix for #434, so the tests can pass.

src/repo.rs
```diff
@@ -2346,6 +2347,24 @@ impl Repository {
...
     }
+
+    /// Perform a cherrypick
+    pub fn cherrypick(
+        &self,
+        commit: &Commit<'_>,
+        options: Option<&mut CherrypickOptions<'_>>,
+    ) -> Result<(), Error> {
+        let raw_opts = options.map(|o| o.raw());
+        let ptr_raw_opts = match raw_opts {
```
alexcrichton: Oh this'll want to be `raw_opts.as_ref()` to avoid moving `raw_opts` and creating a dangling pointer again

src/cherrypick.rs
```diff
@@ -0,0 +1,72 @@
...
+            let mut checkout_opts: raw::git_checkout_options = mem::zeroed();
+            raw::git_checkout_init_options(&mut checkout_opts, raw::GIT_CHECKOUT_OPTIONS_VERSION);
+            if let Some(ref mut cb) = self.checkout_builder {
+                cb.configure(&mut checkout_opts);
+            }
+
+            let mut merge_opts: raw::git_merge_options = mem::zeroed();
+            raw::git_merge_init_options(&mut merge_opts, raw::GIT_MERGE_OPTIONS_VERSION);
+            if let Some(ref opts) = self.merge_opts {
+                ptr::copy(opts.raw(), &mut merge_opts, 1);
```
alexcrichton: Could this be `merge_opts = opts.raw()`?

src/repo.rs
vthib: Yes, this as_ref() method was what I was missing. Done

src/cherrypick.rs
vthib: opts.raw() returns a pointer, and merge_opts is a struct, so we need to do something like `merge_opts = *opts.raw()`. This ptr::copy came from the previous PR, I don't know if there is a better way to do this

alexcrichton: Ah ok that's fine

alexcrichton:
Thanks!

vthib:
Thanks a lot! Any idea when a new release could be done? So that I can use this method in my project.

alexcrichton:
Sure! After https://github.com/rust-lang/git2-rs/pull/440 goes through we can do a release

PR #433, opened 2019-06-26
Update versions in readme

pluehne:
The version of the `git2` crate mentioned in the readme as well as the currently used version of `libgit2` were outdated. This patch brings both up-to-date.


ehuss:
Thanks!

PR #436, opened 2019-07-12
Fix azure rustc installation.

ehuss:
Azure has pre-installed rustup on Linux and Windows, but not macOS.  The current install script doesn't handle if rustup is already installed, and thus wasn't testing the correct versions.


ehuss:
This should fail because the tests are currently broken on beta/nightly.

ehuss:
This should work after #437 is merged.

PR #440, opened 2019-07-13
Update libgit2 to latest master.

ehuss:


ehuss:
Let me know if you have any questions, I have some notes on some of the changes.

I switched the SHA1 implementation because the Windows HTTPS version seems to be broken. "CollisionDetection" is the default now on libgit2, so I imagine it's probably good enough?

Let me know if there is a better way to handle the settings for the PCRE stuff.  I couldn't find a way to tell the `cc` crate to build a set of files with different settings.  It seems to be all or nothing?

libgit2-sys/build.rs
```diff
@@ -31,33 +31,54 @@ fn main() {
...
 
-    // Always use bundled regex for now
-    cfg.include("libgit2/deps/regex")
-        .file("libgit2/deps/regex/regex.c");
+    // Use the included PCRE regex backend.
+    //
+    // Ideally these defines would be specific to the pcre files (or placed in
+    // a config.h), but since libgit2 already has a config.h used for other
+    // reasons, just define on the command-line for everything. Perhaps there
+    // is some way with cc to have different instructions per-file?
```
alexcrichton: Typically the way to go for this is to just use multiple `cc::Build` instances, but it's probably fine to do the same config for all of libgit2

alexcrichton:
Following upstream for defaults sounds reasonable to me! Did the regex backend get dropped in favor of a pcre2 dependency? It'd be sort of neat if we could patch in the `regex` crate instead of having libgit2 use its own backend, but that's probably a more long-term feature dependency. I'm also not entirely sure what it's even used for :)

ehuss:
> Did the regex backend get dropped in favor of a pcre2 dependency?

Yes, here's the PR: https://github.com/libgit2/libgit2/pull/4935  It looks like they were having some issues with the old glibc implementation.

> It'd be sort of neat if we could patch in the `regex` crate instead of having libgit2 use its own backend

It's kind hard for me to wrap my head around how linking would work for that. Presumably you'd have something that had `extern "C"` definitions for the API that libgit2 expects that would translate to the equivalent Rust code?  I'm not really sure how that would get injected into the right time and place during linking, and deal with all the rust side of things like libstd.

> I'm also not entirely sure what it's even used for :)

A few places:
- Config stuff, like searching through config values.
- Used in the diff engine.
- "rev parse" supports patterns.

alexcrichton:
Ok that all sounds good to me! Thanks again for taking this on!

As for how it'd use the `regex` crate itself, that's mostly just an idle musing. It's probably invalid anyway with different syntaxes, but I'd hope for something like we intialize libgit2 with a set of function pointers that it can call back into to do regex operations. Anyway that's pie-in-the-sky and not at all related to this, so I'll merge!`

PR #444, opened 2019-07-23
packbuilder: allow users to set number of threads

bk2204:
libgit2 provides the option for the pack builder to be multi-threaded during deltification. On repositories of any significant size, such as the Linux kernel, threading provides a significant performance benefit. Allow users to specify the thread count used by a pack builder.

Opinions about the best interface here are welcome. The interface accepts 0 for "number of CPUs" and returns the number of threads actually set, which will always be 1 if threading is disabled. I'm unsure if we want an `Option<u32>` instead (with `None` as 0) or if specifying 0 as the value is better. Also, I considered checking the return value and coercing it into a Result if threads are disabled, but there isn't a really good value here (`GIT_EINVALID`? `GIT_ENOTFOUND`?).

I'm happy to revise (including the documentation) with whatever's decided if there's a better way forward.

### DIFF
--- a/src/packbuilder.rs
+++ b/src/packbuilder.rs
@@ -138,6 +138,13 @@ impl<'repo> PackBuilder<'repo> {
         Ok(())
     }
 
+    /// Set the number of threads to be used.
+    ///
+    /// Returns the number of threads to be used.
+    pub fn set_threads(&mut self, threads: u32) -> u32 {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

alexcrichton:
I think mirroring the C interface is pretty reasonable here, so should be all good!

PR #445, opened 2019-07-24
Update depenency on `url` to 2.0

alexcrichton:
Looks like no changes needed!


ehuss:
Looks like msrv failed:

```
error: local variables in const fn are unstable
  --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/percent-encoding-2.0.0/lib.rs:72:13
   |
72 |         let chunk = self.mask[byte as usize / BITS_PER_CHUNK];
   |             ^^^^^

error: aborting due to previous error
```

PR #451, opened 2019-08-08
Correct dependency versions

kornelski:
This crate does *not* compile with bitflags 1.0.0 or log 0.4.0, newer versions are required.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,9 +21,9 @@ azure-devops = { project = "rust-lang/git2-rs", pipeline = "rust-lang.git2-rs", 
 
 [dependencies]
 url = "2.0"
-bitflags = "1.0"
+bitflags = "1.1.0"
 libc = "0.2"
-log = "0.4"
+log = "0.4.8"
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!  I'm curious how you came up with these versions?  They don't seem to work for me.  That is:
```
rm Cargo.lock
cargo generate-lockfile
cargo update -p bitflags:1.1.0 --precise 1.0.2
cargo update -p log:0.4.8 --precise 0.4.1
cargo check
```

Or using `-Zminimal-versions` with this PR.

This fails with a bunch of errors.  The minimum versions that I can determine are:
```
bitflags = "1.0.4"
log = "0.4.4"
```
Although frankly I think using the latest versions would be reasonable.

kornelski:
I've started with -Z minimal-versions and kept bumping them until errors went away.

ehuss:
Did you delete `Cargo.lock` between runs?  `-Zminimal-versions` seems to not play well with lock files.

kornelski:
Ah, that was it. I've set it latest just in case.

ehuss:
Thanks!

PR #452, opened 2019-08-13
Update libgit2 submodule

carlosmn:
Includes a fix for a potential DoS/memory corruption and adds checks for
configuration file in ProgramData.


ehuss:
Thanks!  `3828d7afdd08b595584048e8e4dab6ddd4506ed1` doesn't appear to be a valid commit, though?

carlosmn:
It is valid, but it's actually rolling back to the v0.27 series and I'm not sure how it got there. I meant to update to HEAD. I'll push up a correction.

carlosmn:
Unfortunately it looks like there hasn't been an update in a while so we still need to adjust to some API changes.

alexcrichton:
Thanks!

PR #456, opened 2019-08-15
Bump version numbers of crates

alexcrichton:
* Bump main crate to 0.10.0
* Bump libgit2-sys to 0.9.0
* Bump git2-curl to 0.11.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "git2"
-version = "0.9.2"
+version = "0.10.0"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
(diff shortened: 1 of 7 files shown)


ehuss:
Just want to be clear on the exact reason for this, is this due to the Option changes? Should this be done every time libgit2 is updated?

I think a version bump should update a few other things:
* The three instances of `html_root_url`.
* The versions in the README.

If that's too much hassle, perhaps it would be easiest to just change the `html_root_url` to use `*` for the version? I'm pretty sure docs.rs handles that, the only downside is that the user is redirected to the *latest* version regardless of what they are actually using. 

Or we could use a script to bump the version?

alexcrichton:
Correct yeah, only if libgit2-sys has an API-breaking change do I think we need to bother with a major version bump. Once libgit2-sys has a breaking version bump though that cascades to everything else

I'll edit those versions too, and yeah it's probably best to have a script (or at least instructions somewhere) at this point!

PR #459, opened 2019-08-20
Fix some memory issues around rebase

### LINKED ISSUE #457: Memory errors in RebaseOptions
There appears to be two issues in `RebaseOptions`:

- [`RebaseOptions::rewrite_notes_ref`](https://github.com/rust-lang/git2-rs/blob/c423e2b712cef58ea85a1e60f22daeb758ae48f7/src/rebase.rs#L61-L65) appears to have a dangling pointer to a temporary `CString`.  I believe what is typically done in git2 is to add an `Option<CString>` to `RebaseOptions` to hold the reference.

- [`RebaseOptions::raw`](https://github.com/rust-lang/git2-rs/blob/c423e2b712cef58ea85a1e60f22daeb758ae48f7/src/rebase.rs#L83-L95) has a call to `mem::forget` on a reference, which does nothing.  I'm not sure how this should work, since it is a little convoluted, and behaves differently from other `raw` functions, so I'm not sure why it is written the way it is.

carlosmn:
Fixes #457

### DIFF
--- a/src/rebase.rs
+++ b/src/rebase.rs
@@ -86,7 +86,6 @@ impl<'cb> RebaseOptions<'cb> {
         unsafe {
             if let Some(opts) = self.merge_options.as_mut().take() {
                 ptr::copy_nonoverlapping(opts.raw(), &mut self.raw.merge_options, 1);
-                mem::forget(opts);
             }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #465, opened 2019-09-03
Add merge_base_many and merge_bases_many for Repository

nossralf:
This adds `merge_base_many` and `merge_bases_many` methods to the `Repository` struct.

I also added a smoke test that covers `merge_base` and `merge_base_many`, and reused the preexisting smoke test for `merge_bases` to also cover `merge_bases_many`.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2879,11 +2879,25 @@ extern "C" {
         two: *const git_oid,
     ) -> c_int;
 
+    pub fn git_merge_base_many(
+        out: *mut git_oid,
+        repo: *mut git_repository,
+        length: size_t,
+        input_array: *const git_oid,
+    ) -> c_int;
+
     pub fn git_merge_bases(
         out: *mut git_oidarray,
         repo: *mut git_repository,
         one: *const git_oid,
         two: *const git_oid,
+    ) -> c_int;
+
+    pub fn git_merge_bases_many(
+        out: *mut git_oidarray,
+        repo: *mut git_repository,
+        length: size_t,
+        input_array: *const git_oid,
     ) -> c_int;
 
     // pathspec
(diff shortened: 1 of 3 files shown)


nossralf:
One specific thing that's worth looking extra at in this case, as I'm not experienced at all with Rust FFI, is the conversion from a slice of `Oid` Rust structs to a C array of `git_oid` structs (it's *not* an array of pointers to `git_oid` structs). Dereferencing the pointer required unsafe and does include the possibility of null pointer errors and similar things. I'm not quite sure how to bridge the FFI gap in this case, and I welcome feedback to avoid hidden footguns.

I found no prior art in git2-rs where a libgit2 function takes an array of C structs, from what I could find basically only `git_merge_base_many`, `git_merge_bases_many` and `git_merge_base_octopus` in libgit2 take a C array of `git_oid` structs.

alexcrichton:
Thanks for the PR! If you'd like, want to put `#[repr(C)]` on `Oid` and then we can just convert `*const Oid` to `*const raw::git_oid`?

nossralf:
Trying to wrap my head around this. So, by using `#[repr(C)]` on `Oid`, the iterator chain would look something like this, does that make sense?
```rust
let raw_oids = oids
    .iter()
    .map(|oid| *(oid as *const Oid as *const raw::git_oid))
    .collect::<Vec<_>>();
```
Or am I misunderstanding something and a simpler solution is possible?

In my admittedly limited understanding this looks sort of like the current solution in this PR. I guess the difference is that in the above snippet, we know that the pointer we dereference isn't null since we get it from the reference of an `Oid` struct that rustc guarantees won't be `null`?

alexcrichton:
Oh sorry with the repr attribute that means `&[Oid]` is equivalent to `&[git_oid]` so no new allocation is needed

nossralf:
Ah, of course! Silly me.

alexcrichton:
👍

nossralf:
Thanks for the help. I really appreciate it!

PR #479, opened 2019-10-07
Exposes dirty state for DescribeResult struct

fser:


alexcrichton:
Thanks! I think though this is an incomplete `struct` in C so we're not supposed to be defining and using it?

fser:
Indeed...We didn't notice until the failed build :(

PR #482, opened 2019-10-22
Accept all refspecs that implement AsRef<str>

pluehne:
This changes the signature of the `Remote::fetch` and `Remote::push` methods to accept refspecs as slices over not only `&str` but anything that implements `AsRef<str>` (as well as `IntoCString` and `Clone`). In this way, refspecs can be provided, for example, by a `&Vec<String>`, which wasn’t possible before.

The reason I’d like to contribute this change is that I’m in a situation where I retrieve remote references with `RemoteConnection::list` and turn them into a list of refspecs I want to fetch (or push), leaving me with a `Vec<String>` of refspecs. However, in order to call `Remote::fetch` (or `Remote::push`), I currently have to allocate a new `Vec<&str>` from this `Vec<String>`, which I’d like to avoid if possible.

I’m still in the process of learning Rust, so I hope that this change and my explanations make sense :slightly_smiling_face:. As far as I can tell, it’s not a breaking change in the sense that existing code using `Remote::fetch` and `Remote::push` should still be working.

### DIFF
--- a/examples/fetch.rs
+++ b/examples/fetch.rs
@@ -81,7 +81,7 @@ fn run(args: &Args) -> Result<(), git2::Error> {
     // progress.
     let mut fo = FetchOptions::new();
     fo.remote_callbacks(cb);
-    remote.download(&[], Some(&mut fo))?;
+    remote.download(&[] as &[&str], Some(&mut fo))?;
 
     {
         // If there are local objects (we got a thin pack), then tell the user
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -224,9 +224,9 @@ impl<'repo> Remote<'repo> {
     /// let repo = git2::Repository::discover("rust").unwrap();
     /// fetch_origin_master(repo).unwrap();
     /// ```
-    pub fn fetch(
+    pub fn fetch<Str: AsRef<str> + crate::IntoCString + Clone>(
         &mut self,
-        refspecs: &[&str],
+        refspecs: &[Str],
         opts: Option<&mut FetchOptions<'_>>,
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for this!

Due to this being a breaking change with respect to type inference though I'm gonna hold off on merging for now until the next breaking change for git2 is queued up

pluehne:
@alexcrichton: Great, I’m happy to hear that this changed is queued for the next API-breaking release!

Talking about other breaking changes to bundle with this one, I’m starting to think that it may make sense to look for other places in `git2` that could benefit from accepting `AsRef<str>` arguments rather than plain `&str` arguments—for consistency and in order to avoid breaking the API twice when once would do.

`Remote::download` comes to mind for example: https://github.com/rust-lang/git2-rs/blob/e7973c4f9b877bb2005eae704a89e141d0e9f074/src/remote.rs#L180 Generally, I feel that `AsRef<str>` is mostly useful in situations where we’d otherwise deal with `&[&str]` because having to reallocate arrays or vectors of something like `String` just to match the `&[&str]` parameter type is costly. What do you think?

alexcrichton:
Makes sense to me!

pluehne:
@alexcrichton: Thanks for the feedback. I updated `Remote::download` accordingly. I can’t find any other occurrence of the `&[&str]` type in `git2`, so should we leave it at this?

alexcrichton:
Yeah this seems fine and we can do a quick audit before the next release as well

pluehne:
@alexcrichton: Perfect, thanks a lot!

PR #487, opened 2019-10-28
added git-ignore rules for files that the intellij editor produces

alexanderkjall:


alexanderkjall:
After making this pr i learned that I can have a global git-ignore file, feel free to reject this pr as it's not needed.

alexcrichton:
Thanks for the PR! Yeah these sorts of files typically go in a global gitignore rather than a per-project gitignore, so I'm going to close.

Issue #469, opened 2019-09-11
Support getting remotes of branches


keirlawson:
So far as I can see presently there is no way to do this, however libgit2 does have `git_branch_remote_name` and `git_branch_upstream_remote` that could presumably be used to provide this.

PR #488, opened 2019-11-03
Add support for remotes of branches

lunaryorn:
Wrap `git_branch_upstream_name` and `git_branch_upstream_remote` which retrieve the configured upstream remote and branch, but do not attempt to resolve them, so unlike `Branch::upstream` these functions don't fail if the branch has a configured upstream which doesn't exist.  This way we can detect "gone" branches, i.e. branches whose upstream got deleted on the remote and was subsequently pruned by `git fetch`.

I don't know whether I put the functions in the correct place, and I'm not sure how to test these; I'm happy to amend this pull request if you would review it and give me directions.

Closes GH-469

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2424,6 +2424,16 @@ extern "C" {
     ) -> c_int;
     pub fn git_branch_upstream(out: *mut *mut git_reference, branch: *const git_reference)
         -> c_int;
+    pub fn git_branch_upstream_name(
+        out: *mut git_buf,
+        repo: *mut git_repository,
+        refname: *const c_char,
+    ) -> c_int;
+    pub fn git_branch_upstream_remote(
+        out: *mut git_buf,
+        repo: *mut git_repository,
+        refname: *const c_char,
+    ) -> c_int;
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I wonder if it'd be better though here to return the `Buf` directly instead of forcing a utf-8 check and extra allocation?

lunaryorn:
I don’t know.  For my purpose either is fine, but I don’t think I know Rust and Git so well as to make a general choice.  I’d like to leave this to you, but I’m happy to update my pull request accordingly 🙂

alexcrichton:
Ah ok, in that case yeah let's switch to returning `Buf` directly!

lunaryorn:
@alexcrichton Thanks for the feedback; I'll do that soon :)

lunaryorn:
@alexcrichton Thanks for the feedback; I'll do that soon :)

lunaryorn:
@alexcrichton Done :)

alexcrichton:
👍

lunaryorn:
Many thanks for merging so quickly 🙏. And thanks for this great library!

Issue #485, opened 2019-10-26
Missing binding for git_commit_create_buffer function


alexanderkjall:
I'm trying to use the commit_signed function (https://docs.rs/git2/0.10.1/git2/struct.Repository.html#method.commit_signed) and I need to supply the commit object as a buffer.

From what I can understand from the libgit2 documentation I get that buffer from the git_commit_create_buffer function but it doesn't have a wrapper here.

Is there some other way to get the commit as a buffer in memory?

Also, should the arguments really be of type &str? And not a u8 buffer?

alexcrichton:
Thanks for the report! Not all functions are bound in this library and libgit2 is constantly growing over time as well, but feel free to send a PR which binds this function!

alexanderkjall:
I'll try. But there might be some beginner mistakes.

alexanderkjall:
I think this might be a start: https://github.com/alexanderkjall/git2-rs/commit/d180aaa7ca5f0704aef19a781127de4ab7e685f3

But I have a hard time figuring out how to test it in my project since libgit2-sys and git2 seems to be in the same git repository.

If i add this as a dependency: git2 = { git = "https://github.com/alexanderkjall/git2-rs", branch = "add-git_commit_create_buffer-bindings" }

I get this error:
```
    Updating git repository `https://github.com/alexanderkjall/git2-rs`
error: failed to select a version for `libgit2-sys`.
    ... required by package `git2 v0.10.1 (https://github.com/alexanderkjall/git2-rs?branch=add-git_commit_create_buffer-bindings#d180aaa7)`
    ... which is depended on by `ripasso v0.3.0-alpha (/home/capitol/project/ripasso)`
    ... which is depended on by `ripasso-cursive v0.3.0-alpha (/home/capitol/project/ripasso/cursive)`
versions that meet the requirements `^0.9.1` are: 0.9.1

the package `libgit2-sys` links to the native library `git2`, but it conflicts with a previous package which links to `git2` as well:
package `libgit2-sys v0.9.1`
    ... which is depended on by `git2 v0.10.1`
    ... which is depended on by `ripasso-cursive v0.3.0-alpha (/home/capitol/project/ripasso/cursive)`

failed to select a version for `libgit2-sys` which could resolve this conflict
```

ehuss:
@alexanderkjall You might have better luck using a `[patch]` table in the top of your workspace, as it looks like you have multiple things depending on `git2`, maybe something like:

```toml
[dependencies]
[...]

PR #489, opened 2019-11-10
added bindings for git_commit_create_buffer

alexanderkjall:
In reference to bug #485 .

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -1089,6 +1089,40 @@ impl Repository {
         }
     }
 
+    /// Create a commit object and return that as a string.
+    ///
+    /// This string is suitable to be passed to the `commit_signed` function,
+    /// the arguments behave the same as in the `commit` function.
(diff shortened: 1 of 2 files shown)


src/repo.rs
```diff
@@ -1089,6 +1089,40 @@ impl Repository {
...
+                self.raw(),
+                author.raw(),
+                committer.raw(),
+                ptr::null(),
+                message,
+                tree.raw(),
+                parents.len() as size_t,
+                parent_ptrs.as_mut_ptr()
+            ));
+            Ok(str::from_utf8(&buf).unwrap().to_string())
```
alexcrichton: Could this perhaps return the `Buf` instead of converting to a `String`?

alexanderkjall: @alexcrichton fixed, I converted it to a string due to that commit_signed took a string as an argument.

PR #494, opened 2019-11-18
Add a missing initialization in `Patch::from_buffers`.

### LINKED ISSUE #491: Segfault on uninitialized git global allocator
The following code causes a segfault in [patch_generated_with_delta_alloc](https://github.com/libgit2/libgit2/blob/01ea911b9e14a38f94a0018c13e5f308719aced6/src/patch_generate.c#L508).

```rust
fn main() {
    git2::Patch::from_buffers(
        "abc".as_bytes(),
        None,
        "abc".as_bytes(),
        None,
        None,
    );
}
```

I believe this is because git's global allocator never gets initialized. A workaround is to add `libgit2_sys::init()` at the top of main. Should this be necessary, or should the safe git2 binding be calling init in more places?

alexcrichton:
Closes #491

### DIFF
--- a/src/patch.rs
+++ b/src/patch.rs
@@ -100,6 +100,7 @@ impl Patch {
         new_path: Option<&Path>,
         opts: Option<&mut DiffOptions>,
     ) -> Result<Patch, Error> {
+        crate::init();
         let mut ret = ptr::null_mut();
         let old_path = into_opt_c_string(old_path)?;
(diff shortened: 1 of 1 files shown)


zertosh:
Can a release be made with this fix please?

alexcrichton:
Sure thing, pushed up as https://github.com/rust-lang/git2-rs/pull/496

PR #495, opened 2019-11-18
Implement Debug for Patch and related diff structs

zertosh:
These are tricky structs to debug and understand. They can really benefit
from being Debug-printable.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -444,6 +444,17 @@ impl<'a> Binding for DiffDelta<'a> {
     }
 }
 
+impl<'a> std::fmt::Debug for DiffDelta<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        f.debug_struct("DiffDelta")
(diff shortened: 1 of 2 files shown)


src/diff.rs
```diff
@@ -863,6 +888,23 @@ impl<'a> Binding for DiffLine<'a> {
...
+        let mut ds = f.debug_struct("DiffLine");
+        if let Some(old_lineno) = &self.old_lineno() {
+            ds.field("old_lineno", old_lineno);
+        }
+        if let Some(new_lineno) = &self.new_lineno() {
+            ds.field("new_lineno", new_lineno);
+        }
+        ds.field("num_lines", &self.num_lines())
+            .field("content_offset", &self.content_offset())
+            .field("content", &self.content())
```
zertosh: I wonder if it makes more sense to print this field like so:

```
.field("content", &format_args!("[{} bytes]", self.content().len()))
```

alexcrichton:
Nice! 

I'm down for improving this if necessary, but given the built-in debug formatters we may not want to add too much custom formatting.

PR #502, opened 2019-12-06
Allow delegating all linking

krallin:
This is necessary when an external build system (e.g. Buck) is taking
care of linking in git-sys, and we don't want the Rust compiler to emit
any linking arguments.

To do so, this adds a flag to just bail out of the build script and not
build any native crate if LIBGIT2_SYS_DELEGATE_LINKING is set in the
environment.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -4,6 +4,14 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 fn main() {
+    let k = "LIBGIT2_SYS_DELEGATE_LINKING";
+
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This feature though I think is handled by [build script overrides](https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts), so I don't think there's a need to develop a libgit2-sys-specific solution here.

krallin:
Ah, thanks a lot for the pointer, @alexcrichton! I'll update on our end to try and use this instead :)

PR #503, opened 2019-12-06
Update libgit2 submodule.

nossralf:
This updates the libgit2 submodule to latest master.

As several public struct members as well as some functions' parameters/return values have changed fundamental types (e.g from `git_off_t` which is an `int64_t` to `git_object_size_t` which is an `uint64_t`) I bumped the minor versions of both libgit2-sys and git2-rs (meaning it's equivalent to a major version bump for 0.x versions). As I'm not familiar with the version increment policy, let me know if that's wrong.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "libgit2-sys/libgit2"]
 	path = libgit2-sys/libgit2
 	url = https://github.com/libgit2/libgit2
-	branch = maint/v0.28
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git2"
-version = "0.10.2"
+version = "0.11.0"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
@@ -20,7 +20,7 @@ url = "2.0"
 bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
-libgit2-sys = { path = "libgit2-sys", version = "0.9.2" }
+libgit2-sys = { path = "libgit2-sys", version = "0.10.0" }
 
 [target."cfg(all(unix, not(target_os = \"macos\")))".dependencies]
 openssl-sys = { version = "0.9.0", optional = true }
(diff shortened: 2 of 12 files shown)


nossralf:
I guess I'll have to get a VM with Windows up and running tomorrow to figure out why the Windows build fails.

nossralf:
The change from `enum` to `uint32_t` for flags in some structs is what tripped me up. On Windows with msvc, where `git_enum!` uses `i32`, the raw `enum` values must be cast to `u32` for rustc to be satisfied.

The [libgit2 pull request](https://github.com/libgit2/libgit2/pull/5242) which introduces this change mentions the ergonomic downsides for C++ that results from using an `uint32_t`, which may be worth mentioning.

As the usage of raw flags is hidden behind methods in the Rust API, the numeric casts don't affect users of git2-rs unless they are using raw values directly.

alexcrichton:
Everything looks great to me here thanks!

For the `enum` changes, could the type definitions for each type be switched to `u32` all platforms too? That should remove the need ideally for the `as` cast when using the named constant values.

nossralf:
I tried to do that (by using `pub enum git_merge_flag_t: u32` in the `git_enum!` definition for all the affected structs) but it led to errors on Windows for the systest suite related to the structs.

alexcrichton:
Sorry for the delayed response, but could you gist what the errors look like in that case?

nossralf:
This is the output I get when I set the `git_enum!` type to `u32` explicitly for these three enums.
```
RUNNING ALL TESTS
bad git_merge_flag_t signed: rust: 0 (0x0) != c 1 (0x1)
bad git_merge_file_flag_t signed: rust: 0 (0x0) != c 1 (0x1)
bad git_stash_apply_flags signed: rust: 0 (0x0) != c 1 (0x1)
```
The tests that fail all seem to be generated as such, with just the enum name changed:
```rust
#[inline(never)]
#[allow(non_snake_case)]
fn sign_git_stash_apply_flags() {
  extern {
    #[allow(non_snake_case)]
    fn __test_signed_git_stash_apply_flags() -> u32;
  }
  unsafe {
    same(((!(0 as git_stash_apply_flags)) < (0 as git_stash_apply_flags)) as u32,
      __test_signed_git_stash_apply_flags(), "git_stash_apply_flags signed");
  }
}
```
I don't understand what information the systests are generated from, but that a test that covers signedness fails when the type is set to unsigned at least makes some sense based on the name alone. (If it makes sense to someone who understands the tests more deeply is unclear to me.)

alexcrichton:
Ah ok no worries! I think that the definitions may be a bit odd at the C layer, but that's ok. Let's go ahead with this and we can improve it later if necessary!

PR #505, opened 2019-12-09
Make it possible to link in http_parser and pcre

krallin:
This adds features for using dynamically linked http_parser and pcre
libraries. This makes it possible to use git2-rs in a binary that
otherwise links those in already without duplicate linking issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,13 @@ If the `ssh` feature is enabled (and it is by default) then this library depends
 on libssh2 which depends on OpenSSL. To get OpenSSL working follow the
 [`openssl` crate's instructions](https://github.com/sfackler/rust-openssl#macos).
 
+## Linking to system libraries
+
+The underlying libgit2 library vendors two of its dependencies: pcre and
+http_parser`. By default, the libgit2-sys crate builds those two dependencies
+along with libgit2. If youd like to link them in instead, enable the
+`link_pcre` and / or `link_http_parser` features for the libgit2-sys crate.
+
 # License
 
 This project is licensed under either of
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -32,3 +32,5 @@ openssl-sys = { version = "0.9", optional = true }
 ssh = ["libssh2-sys"]
 https = ["openssl-sys"]
 ssh_key_from_memory = []
+link_http_parser = []
+link_pcre = []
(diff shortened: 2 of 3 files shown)


krallin:
@alexcrichton: I'd like to follow up on the discussion we had in #502 a bit. What I noted after going down that road is that libgit2-sys doesn't really target a stable libgit2, because there is no such libgit2. This puts me in a situation where I need to get libgit2 built through libgit2-sys, but would still want its transitive dependencies to not be compiled in.

The reason for that is that we link all our binaries statically, so if libgit2-sys contains libgit2 and its transitive dependencies, it contains PCRE symbols, and that conflicts with linking in PCRE itself (which is getting linked in since it's dependend)

So, in an ideal world, I'd like to use libgit2's `build.rs` script, but not have it emit linker arguments for the libraries it's linking in. Would you be OK with me sending that change on top of this PR?

alexcrichton:
Seems reasonable to me!

Want to document these features in `Cargo.toml` and I can take a look at the further changes?

krallin:
@alexcrichton: sounds good, thanks; I updated the README here to explain the features.

The further changes are over here in this commit: https://github.com/krallin/git2-rs/commit/4d64ec2a9d6c953cff7de9226aadcec942a8ea12 — I put the documentation inline since this feels like a pretty narrow use case. Let me know if this looks reasonable? If it does, I'll add this in to this PR (unless you'd rather I submit it as a separate PR? just let me know)

Thanks!

README.md
```diff
@@ -47,6 +47,13 @@ If the `ssh` feature is enabled (and it is by default) then this library depends
 on libssh2 which depends on OpenSSL. To get OpenSSL working follow the
 [`openssl` crate's instructions](https://github.com/sfackler/rust-openssl#macos).
 
+## Linking to system libraries
+
+The underlying libgit2 library vendors two of its dependencies: pcre and
+http_parser`. By default, the libgit2-sys crate builds those two dependencies
+along with libgit2. If youd like to link them in instead, enable the
```
ehuss:
```suggestion
along with libgit2. If you would like to link them in instead, enable the
```

joshtriplett:
@krallin I thought those build systems typically used the cargo feature to fully *override* build scripts so that they don't run at all, and supplied the libraries themselves? This is the first time I've seen any notion of running a build script but not actually linking the libraries, and that's an unusual use case that I'd want to see more explanation and exploration of.

In the meantime, though, I'd be happy to merge the change to dynamically link an external http_parser and pcre, if you rebase this to remove the conflicts.

ehuss:
Closing due to inactivity.

PR #506, opened 2019-12-13
Add missing error classes and error codes

nossralf:
This makes the git2-rs error classes and error codes be in sync with libgit2's.


alexcrichton:
Thanks!

Issue #516, opened 2020-02-05
libgit2-sys does not support detail error output in init()


cosmo0920:
Currently, libgit2-sys does not output error detail but just error code only.
We'd struggled why libgit2-sys complains `-1` error.

OS: Ubuntu 18.04 LTS
libgit2-rs: 0.10.0 also hit in 0.8.1
OpenSSL: 1.1.1d compiled with `--with-rand-seed=devrandom` w/o os seed
Within Virtual machine Yocto building environment and using [meta-rust](https://github.com/meta-rust/meta-rust)

cosmo0920:
#517 is merged. Closing.

PR #517, opened 2020-02-05
Support libgit2 error message report when assertion failed

cosmo0920:
Addresses #516 

Current libgit2-sys implementation does not call `git_error_last()` to get *git_error object* and reporting its *message member* contents.
E.g. cargo just complains the following error when `git_libgit2_init()` is failing with the current implementation:

```
$ ./cargo --version
thread 'main' panicked at 'couldn't initialize the libgit2 library: -1', /usr/src/debug/cargo-native/1.37.0-r0/rustc-1.37.0-src/vendor/libgit2-sys/lib.rs:3356:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -7,6 +7,8 @@ extern crate libz_sys as libz;
 use libc::{c_char, c_int, c_uchar, c_uint, c_void, size_t};
 #[cfg(feature = "ssh")]
 use libssh2_sys as libssh2;
+use std::ffi::CStr;
+use std::ptr;
 
 pub const GIT_OID_RAWSZ: usize = 20;
 pub const GIT_OID_HEXSZ: usize = GIT_OID_RAWSZ * 2;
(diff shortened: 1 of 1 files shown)


libgit2-sys/lib.rs
```diff
@@ -3551,7 +3553,21 @@ pub fn init() {
...
+        if !error_msg.is_null() {
+            assert!(
+                r >= 0,
+                "couldn't initialize the libgit2 library: {}, error: {}",
+                r,
+                CStr::from_ptr(error_msg).to_string_lossy()
+            );
+        } else {
+            assert!(r >= 0, "couldn't initialize the libgit2 library: {}", r);
+        }
```
alexcrichton: Could this perhaps be structured to check the return code, and then if it's negative fetch the error message and print it?

cosmo0920: Sure. Will do.

cosmo0920: I've commit https://github.com/rust-lang/git2-rs/pull/517/commits/83c5b25ba429772c77bf5e9f7395095624f30fa5 for your suggested comment.

cosmo0920:
Thanks!

Issue #515, opened 2020-02-04
Support non defined value in ConfigEntry


Vitroxyn:
Currently it is assumed that ConfigEntry has always a value, but this is not the case with a variable defined without `= <value>`, which is a [shorthand to represent a true value](https://git-scm.com/docs/git-config#Documentation/git-config.txt-true).

For example this git config:
```
[core]
    commitGraph
```

Trying to access the value results in panicking.
```rust
let config_entry = cfg.get_entry("core.commitGraph").unwrap();
config_entry.value_bytes(); //Panics
config_entry.value(); //Panics
```

Backtracking this issue:
```rust
pub fn value_bytes(&self) -> &[u8] {
    unsafe { crate::opt_bytes(self, (*self.raw).value).unwrap() }
}
```
The panic is caused by unwrapping the result of `opt_bytes` which is `None`, because `value` is a null pointer.

To not introduce breaking api changes, there should be atleast a function to check for existance of a value.

For example this function inside `ConfigEntry` would allow checking for a defined value:
```
pub fn has_value(&self) -> bool {
    !self.raw.value.is_null()
}
```

Another approach would be to change the function `value()` in `ConfigEntry` to return also `None` when the value is not defined.

Also the documentation could be improved to inform about panicking when a value is not defined.

PR #518, opened 2020-02-06
Add support for non defined value in ConfigEntry

Vitroxyn:
Addresses #515

To not introduce breaking api changes `ConfigEntry` has a new method `has_value()` which returns `true` when a value is defined. The methods `value_bytes()` and `value()` are panicing when a non defined value is present this could be prevented by checking `has_value()` first.


alexcrichton:
Thanks!

PR #519, opened 2020-02-16
Add examples

rgreinho:
While starting to use this library, I had some trouble understanding how to 
perform the 3 following tasks, therefore I thought I would contribute
some examples to the documentation.

The examples should help to:

* Clone/push using SSH
* Stage all files

Hopefully it will also help some other people.


alexcrichton:
Thanks!

PR #523, opened 2020-02-21
Tie Patch lifetime to its input buffers

yuyoyuppe:
It's currently possible to get a use-after-free errors when using `Patch::from_buffers`. Here's a small code to repro:

```rust
use git2::*;

fn main() {
    let repo = Repository::open(REPO_PATH).unwrap();
    let old_file = repo
        .find_blob(Oid::from_str(BLOB_SHA1).unwrap())
        .unwrap();
    let new_file = repo
        .find_blob(Oid::from_str(BLOB_SHA2).unwrap())
        .unwrap();

    let diff_options = &mut DiffOptions::new();
    let patch = Patch::from_buffers(
        old_file.content(),
        None,
        new_file.content(),
        None,
        Some(diff_options),
    )
    .unwrap();

     std::mem::drop(old_file);
     std::mem::drop(new_file);

    for hunk_idx in 0..patch.num_hunks() {
        for line_idx in 0..patch.num_lines_in_hunk(hunk_idx).unwrap() {
            let diff_line = patch.line_in_hunk(hunk_idx, line_idx).unwrap();
            let line = String::from_utf8(diff_line.content().into())
                .map(|s| String::from(s.trim_end()))
                .unwrap_or_else(|e| {
                    println!("utf8 decoding error: {}", e);
                    String::from("!USE AFTER FREE!")
                });
            println!("line: {}", line);
        }
    }
}
```
On my system it fails in ~5% cases, so it may be hard to repro. That happens because we're passing `old_file.content()`/`old_file.content()` pointers, [free them with `mem::drop`](https://docs.rs/git2/0.11.0/src/git2/blob.rs.html#81-83) and try to use them again.

We also don't attempt to double free the pointers in [`patch_generated_free`](https://github.com/libgit2/libgit2/blob/2376cd26226771dcf8ef5dfd04c83a50c50bf3d4/src/patch_generate.c#L37-L38) during `Patch` drop, since we don't have `GIT_DIFF_FLAG__FREE_DATA` flags on theirs [`git_diff_file_content`](https://github.com/libgit2/libgit2/blob/9bcf10e97e9b7c8fc12961e3950ce0a8e64231ce/src/diff_file.h#L21).

I understand that this change could break a lot of code and probably is overly defensive, since I'm not sure if other `Patch::from_*` functions behave the same, so I'm open to suggestions on how to improve this PR!

### DIFF
--- a/src/patch.rs
+++ b/src/patch.rs
@@ -1,4 +1,5 @@
 use libc::{c_int, c_void};
+use std::marker::PhantomData;
 use std::path::Path;
 use std::ptr;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oops thanks so much for catching this, and thanks for the PR!

PR #527, opened 2020-03-13
Update libgit2

ehuss:
Update to latest master.

Brings in https://github.com/libgit2/libgit2/pull/5432 which should fix https://github.com/rust-lang/cargo/issues/7988

Brings in https://github.com/libgit2/libgit2/pull/5435 which should fix https://github.com/rust-lang/cargo/issues/7466


ehuss:
Note that due to https://github.com/gnzlbg/ctest/issues/90, the systest is now failing on the latest nightly.  I ran with previous versions and it passed.  I'm not sure what the scope is of fixing ctest.  I'd be happy to help out, though I'm not really up-to-speed on the best strategy to fix things.

alexcrichton:
I've pushed up https://github.com/rust-lang/git2-rs/commit/d70b64599a080ea91f6052cb234a4a485f5dcdff which I think should fix the issue, want to rebase and see if tests are green?

ehuss:
Thanks @alexcrichton!  Looks like the tests are green.

PR #529, opened 2020-03-14
Libgit2 has a new error class to represent HTTP errors. Add it to the…

iancormac84:
… appropriate enums.


ehuss:
Shouldn't [`Error::class`](https://github.com/rust-lang/git2-rs/blob/30f013875ec69e6c794392a41e3d8f364ea8ec5b/src/error.rs#L123) and [`Error::raw_class`](https://github.com/rust-lang/git2-rs/blob/30f013875ec69e6c794392a41e3d8f364ea8ec5b/src/error.rs#L204) be updated?

alexcrichton:
Thanks!

Issue #403, opened 2019-02-22
git apply patch `missing`


niklasad1:
`git apply patch` support has landed in [libgit2](https://github.com/libgit2/libgit2/issues/1980), any plans to support it?

alexcrichton:
Support can be added at any time, so feel free to send a PR! We currently don't keep issues open for unimplemented libgit2 APIs here though o I'm going to close this.

PR #531, opened 2020-03-20
add apply sys api

Extrawurst:
added missing sys calls for missing `apply` API. is it ok to split this from implementing a safe API or does it have to be in one PR?

(this is related to #403)

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -23,6 +23,7 @@ pub const GIT_SUBMODULE_UPDATE_OPTIONS_VERSION: c_uint = 1;
 pub const GIT_ODB_BACKEND_VERSION: c_uint = 1;
 pub const GIT_REFDB_BACKEND_VERSION: c_uint = 1;
 pub const GIT_CHERRYPICK_OPTIONS_VERSION: c_uint = 1;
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

libgit2-sys/lib.rs
```diff
@@ -1728,6 +1729,35 @@ pub struct git_cherrypick_options {
...
+    }
+}
+
+#[repr(C)]
+pub struct git_apply_options {
+    pub version: c_uint,
+    pub delta_cb: git_apply_delta_cb,
+    pub hunk_cb: git_apply_hunk_cb,
+    pub payload: *mut c_void,
+    pub flags: u32,
```
Extrawurst: Hey @alexcrichton thanks for merging!

Out of curiosity: why is it not possible to idiomatically use the enum types in the structs? This only seemed to fail on windows:
https://github.com/rust-lang/git2-rs/runs/521702373

I see many structs using the `u32` instead of the actual enum type, but some seem to be able to use them, what am I missing?

alexcrichton: I think it has to do with the way MSVC handles enums and whether they're signed or not. I don't remember a ton of specifics but I know it's slightly different from gcc which has caused issues. In any case I'm fine with tweaking things as necessary!

PR #542, opened 2020-04-05
add apply support to repository

extrawurst:
after adding the necessary sys code (https://github.com/rust-lang/git2-rs/pull/531) this adds the new `apply` method to `Repository` and a new `apply` module handling the new `ApplyOptions` and c-callbacks.

(this is related to #403)

### DIFF
--- a/src/apply.rs
+++ b/src/apply.rs
@@ -1,0 +1,208 @@
+//! git_apply support
+//! see original: https://github.com/libgit2/libgit2/blob/master/include/git2/apply.h
+
+use crate::{panic, raw, util::Binding, DiffDelta, DiffHunk};
+use libc::c_int;
+use std::{ffi::c_void, mem};
+
+/// Possible application locations for git_apply
+/// see https://libgit2.org/libgit2/#HEAD/type/git_apply_options
+#[derive(Copy, Clone, Debug)]
+pub enum ApplyLocation {
+    /// Apply the patch to the workdir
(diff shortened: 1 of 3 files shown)


alexcrichton:
Nice!

extrawurst:
@alexcrichton can we have a point release with this? 😏

alexcrichton:
Sure thing, done!

SirWindfield:
Hey! Sorry to kind of necro this, but I have a question regarding this PR. Is this meant to be used with files produced by `git diff`'s output? E.g. `.patch` files that can be applied to repositories? If so, how exactly? The documentation mentions that the `Diff` object is created with one of the `diff_` methods on `Repository`, but that wouldn't apply for the above example.

Maybe I just misunderstood the purpose of this PR...

extrawurst:
@SirWindfield I think this kind of question fits more into libgit2, `git2-rs` is only a wrapper around it. I am using `apply` to do hunk by hunk staging in gitui, maybe this helps you: https://github.com/extrawurst/gitui/blob/master/asyncgit/src/sync/hunks.rs

SirWindfield:
I opened a discussion on their repository. The feature is relatively new and I am the first one, but just for reference: https://github.com/libgit2/libgit2/discussions/5588

@extrawurst hunk by hunk (as you did it) does still compare between the index and the current file, doesn't it?

PR #544, opened 2020-04-06
Make Repository::update_submodules public.

iancormac84:


alexcrichton:
Thanks for this! I've been hesitant to make this public in the past since I'd prefer to not stray too far from libgit2 itself, and this is a pretty inflexible method with how things are updated (and pretty easy to inline locally if necessary).

iancormac84:
No problem. I see where you're coming from, given the difficulty I've had with it, myself.

Issue #164, opened 2016-10-29
status_should_ignore only works for absolute paths


jacwah:
Calling `Repository::status_should_ignore` on a path of the form `./filename` always returns true regardless of the current gitignore file. I tried canonicalizing the path instead which lead to the correct result being returned.

Is this a bug in git2-rs/libgit2? Am I missing something?

Cheers

alexcrichton:
Ah unfortunately this may be behavior relative to libgit2, it looks like we just pass the file unmodified down to the lower layer.

alexcrichton:
Closing in favor of https://github.com/libgit2/libgit2/issues/3979

PR #549, opened 2020-04-13
Attempt to uniformly handle Windows-style paths.

### LINKED ISSUE #340: more path not working on windows

ehuss:
Passing a Windows path with backslashes to a git2 API that expects a repo path is always wrong. This attempts to consolidate the translation of backslashes, and I tried to cover every method that takes a repo path or pathspec.

This also checks if the path is absolute on all platforms. git2 repo paths are never absolute (AFAIK).  The intent is to avoid footguns like #164 where passing filesystem-like paths with leading non-normal components never works (except for maybe `is_path_ignored`).

I flip-flopped a lot on how to approach this, and I'm not entirely happy with it (though I can't pinpoint why).  I generally wanted to avoid changing any public APIs.

Closes #340

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -503,7 +503,7 @@ impl<'cb> CheckoutBuilder<'cb> {
     /// If no paths are specified, then all files are checked out. Otherwise
     /// only these specified paths are checked out.
     pub fn path<T: IntoCString>(&mut self, path: T) -> &mut CheckoutBuilder<'cb> {
-        let path = path.into_c_string().unwrap();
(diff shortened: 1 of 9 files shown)


alexcrichton:
👍 

Seems reasonable to me!

PR #553, opened 2020-04-22
Implement Diff::format_email

Techcable:
The test is a little ugly, but at least it passes :)

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -1208,7 +1208,7 @@ pub struct git_diff_format_email_options {
     pub id: *const git_oid,
     pub summary: *const c_char,
     pub body: *const c_char,
-    pub author: *const git_signature
+    pub author: *const git_signature,
 }
 
 pub const GIT_DIFF_FORMAT_EMAIL_OPTIONS_VERSION: c_uint = 1;
@@ -3265,11 +3265,11 @@ extern "C" {
     pub fn git_diff_format_email(
         out: *mut git_buf,
         diff: *mut git_diff,
-        opts: *const git_diff_format_email_options
+        opts: *const git_diff_format_email_options,
     ) -> c_int;
     pub fn git_diff_format_email_options_init(
         opts: *mut git_diff_format_email_options,
-        version: c_uint
+        version: c_uint,
     ) -> c_int;
 
     // patch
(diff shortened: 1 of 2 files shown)


libgit2-sys/lib.rs
```diff
@@ -9,6 +9,8 @@ use libc::{c_char, c_int, c_uchar, c_uint, c_void, size_t};
 use libssh2_sys as libssh2;
 use std::ffi::CStr;
 
+pub const LIBGIT2_VERSION: &str = "1.0.0";
```
alexcrichton, line 12: Hm ideally this'd be auto-synthesized from libgit2 headers, would it be possible to defer this to later?

src/diff.rs
```diff
@@ -1337,9 +1373,44 @@ impl DiffFindOptions {
...
+        } else {
+            self.raw.flags &= !opt;
+        }
+        self
+    }
+
+    /// Exclude `[PATCH]` from the subject header
+    pub fn exclude_subject_patch_header(&mut self, should_exclude: bool) -> &mut Self {
+        self.flag(raw::GIT_DIFF_FORMAT_EMAIL_EXCLUDE_SUBJECT_PATCH_MARKER, should_exclude)
+    }
```
alexcrichton, line 1408: Would you be interested in expanding this to include other options too?

Techcable: As of right now, these are [the only flags implemented](https://github.com/libgit2/libgit2/blob/fb439c9/include/git2/diff.h#L1367-L1374)

libgit2-sys/lib.rs
Techcable: This was only used for the test to match the version at the end. I changed it to do fuzzy matching.

Techcable:
Thanks for the quick review! I fixed the formatting errors raised by the initial checks.

I can't seem to figure out why they're still failing.

Techcable:
I can confirm using that patch files created with this method can be parsed by #560 (Diff::parse_buffer).
I can take an input patch file (made using libgit2), parse it, apply it, then get an identical patch file if I output it again.

There are some minor formatting differences with the patch files generated by `git format-patch`, but they don't affect patch application.

alexcrichton:
Ok looks good to me, and thanks!

I think with a rustfmt this should be good to go.

PR #554, opened 2020-04-22
Allow absolute paths for Repository::is_path_ignored.

ehuss:
Some users are passing absolute paths to `Repository::is_ignored_path`.  The changes from #549 cause these to be rejected.  I don't think it necessarily needs to take a principled stance of "no absolute paths", and also since this breaks existing usage, I think it should just allow absolute paths.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2597,7 +2597,7 @@ impl Repository {
 
     /// Test if the ignore rules apply to a given path.
     pub fn is_path_ignored<P: AsRef<Path>>(&self, path: P) -> Result<bool, Error> {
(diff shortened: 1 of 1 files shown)


ehuss:
@alexcrichton Sorry, I should have tested this more. As I noted in #549, I suspected this function might not need to be restricted.

This will be needed for https://github.com/rust-lang/cargo/pull/8095, so if you could push another release, I'd appreciate it.  Also, I noticed with the last bump that you didn't push a commit to this repo with the `Cargo.toml` updates.  Sorry about the churn.

alexcrichton:
No worries! I should get better about the release process anyway...

PR #556, opened 2020-04-23
Fixed compile issue on windows-gnu regarding DiffDeltas::flag.

### LINKED ISSUE #555: DiffDeltas flag method fails to compile on Windows (GNU)
git2-rs version 0.13.3 fails to compile on windows-gnu because the `as_u32` method is variant on `windows`, but the git_enum has its variant defined by `msvc`.

pJunger:
Changed the `as_u32` function inside `flag` to also use u32 as parameter in case the environment is windows-gnu.

Closes #555.


pJunger:
Thank you!

PR #557, opened 2020-04-23
Fix more Windows paths.

ehuss:
I missed a bunch of functions in #549 for converting windows paths to git repo paths.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -267,6 +267,7 @@ impl<'cb> RepoBuilder<'cb> {
         }
 
         let url = CString::new(url)?;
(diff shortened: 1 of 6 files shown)


ehuss:
Sorry Alex. 😞   I'm not sure if I had a brain fart or what, somehow I missed all the calls to `into_c_string` (I think I searched for `IntoCString` and then forgot to check the function calls.)

Needed for https://github.com/rust-lang/cargo/pull/8095

ehuss:
Yep, retriggered CI and it seems to be green.

PR #559, opened 2020-04-28
rebase: Add orig_head_name and orig_head_id

abhinav:
Add methods `orig_head_name` and `orig_head_id` to `git2::Rebase`. These
are Rust-level surface for libgit2's [`git_rebase_orig_head_name`][1]
and [`git_rebase_orig_head_id`][2].

  [1]: https://libgit2.org/libgit2/#HEAD/group/rebase/git_rebase_orig_head_name
  [2]: https://libgit2.org/libgit2/#HEAD/group/rebase/git_rebase_orig_head_id

`rebase_head_name` needs the rebase head to have a name; just a commit
won't do. To test it, use `refs/heads/master` as the head instead of the
head commit.

Additionally, since these operations are valid only on merge rebases,
switch the test to on-disk from in-memory.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -3596,6 +3596,8 @@ extern "C" {
         rebase: *mut git_rebase,
         idx: size_t,
     ) -> *mut git_rebase_operation;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #560, opened 2020-04-29
Implement Diff::from_buffer for parsing patch files

Techcable:
I question the decision to make diff live for 'static.
Libgit2 doesn't require that it's dependent on a repository
but it seems inconsitent with all the other methods of
creating diffs.


alexcrichton:
Seems reasonable to me, thanks!

PR #566, opened 2020-05-18
stash_save2 adds support for optional message

### LINKED ISSUE #565: Repository::stash_save NULL message not supported
`git_stash_save` supports NULL as an optional message and will do internal logic to define a stash name.
see https://libgit2.org/libgit2/#HEAD/group/stash/git_stash_save

`git2-rs` currently does not support this behaviour.

If I were to create a PR how do you guys handle this sort of change? the *correct* api would take message as `Option<&str>` and pass null or the converted cstring. this change would break the API. is that ok?

extrawurst:
closes #565

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2511,11 +2511,22 @@ impl Repository {
         message: &str,
         flags: Option<StashFlags>,
     ) -> Result<Oid, Error> {
+        self.stash_save2(stasher, Some(message), flags)
+    }
+
+    /// Save the local modifications to a new stash.
+    /// unlike `stash_save` it allows to pass a null `message`
+    pub fn stash_save2(
+        &mut self,
+        stasher: &Signature<'_>,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could this update the previous method to call the new one too?

extrawurst:
great idea! incoming

extrawurst:
@alexcrichton done

extrawurst:
@alexcrichton cool, then I only need a release now since cargo does not allow git references as dependencies :)

alexcrichton:
Sure thing, published!

extrawurst:
@alexcrichton thanks! how do we make sure we keep track of this to refactor it in a breaking release?

alexcrichton:
Typically open PRs are left, otherwise it's "I hope this sticks in my brain". Not exactly an infallible system!

PR #569, opened 2020-06-14
oid: Add ShortOid, a length-aware oid prefex

### LINKED ISSUE #293: First class short oids?

mjsir911:
I'm not the author of this, just trying to push it through:

This adds a `ShortOid` type that's aware of its length. A regular `Oid`
is convertible to a `ShortOid`. The `Odb::exists_prefix` method is
changed to take `impl Into<ShortOid>` and no length parameter.

The limit of one, and can hinder readability by, eg,
disallowing extra space between top-level items.

BREAKING CHANGES:
- `Odb::exists_prefix` no longer takes a length parameter.

replaces #295
this closes #293, the issue has been closed but it has not been resolved nor has the feature been added

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,7 +97,7 @@ pub use merge::{AnnotatedCommit, MergeOptions};
 pub use message::{message_prettify, DEFAULT_COMMENT_CHAR};
 pub use note::{Note, Notes};
 pub use object::Object;
-pub use oid::Oid;
+pub use oid::{Oid, ShortOid};
 pub use packbuilder::{PackBuilder, PackBuilderStage};
 pub use pathspec::{Pathspec, PathspecMatchList, PathspecFailedEntries};
 pub use pathspec::{PathspecDiffEntries, PathspecEntries};
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -7,7 +7,7 @@ use std::ffi::CString;
 
 use libc::{c_char, c_int, c_void, size_t};
 
-use {raw, Oid, Object, ObjectType, Error};
+use {raw, Oid, ShortOid, Object, ObjectType, Error};
 use panic;
 use util::Binding;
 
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! The context for this is unfortunately super old for me, though. Can you describe to me again the purpose for adding this?

mjsir911:
Yeah, no problem. This PR just adds an abstraction on to the OID type for "partial" OIDs that are not the full length hashes. Git has some support for shorter OIDs, and this will accurately reflect that.

I probably want to add on to this as well and have the short OID type be returned by the `Object.short_id` method since that is literally describing how to get the type.

I didn't realize this, but I'm probably going to have to resubmit the PR under a branch under a repo owned by me, I don't have edit access to resolve issues on this PR

joshtriplett:
@mjsir911 In addition to you needing to resubmit this, this also needs to be reworked to not make a breaking change to `Odb::exists_prefix`. Please add a separate method that takes a `ShortOid`, and keep the existing method that takes an `Oid` and a length. That avoids the need for a breaking change.

PR #573, opened 2020-06-19
Add bindings for git_revert{,_commit}

### LINKED ISSUE #571: add support for `git revert`
Support for `git revert` seems not to be implemented yet. I couldn't find anything relevant in the docs. Also, grepping the source for libgit2's `git_revert` didn't yield any results.

I would find it useful to add support for `git revert`. Please consider adding it.

toxeus:
Closes #571

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -16,6 +16,7 @@ use crate::stash::{stash_cb, StashApplyOptions, StashCbData};
 use crate::string_array::StringArray;
 use crate::util::{self, path_to_repo_path, Binding};
 use crate::CherrypickOptions;
+use crate::RevertOptions;
 use crate::{
     init, raw, AttrCheckFlags, Buf, Error, Object, Remote, RepositoryOpenFlags, RepositoryState,
     Revspec, StashFlags,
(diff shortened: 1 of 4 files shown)


src/repo.rs
```diff
@@ -2721,6 +2722,46 @@ impl Repository {
             Ok(())
         }
     }
+
+    /// Perform a revert.
```
extrawurst, line 2726: looking at the two comments it sounds misleading as If this is not changing the index. reading up on https://libgit2.org/libgit2/#HEAD/group/revert/git_revert and https://libgit2.org/libgit2/#HEAD/group/revert/git_revert_commit it sounds more like both will alter the index. 
some of the methods in git2-rs went about plain duplicating the doc comment from the original. maybe this makes sense here aswell?

toxeus: Yes, both change the index. One changes only the index. Will duplicate the doc comment :+1:

extrawurst:
@alexcrichton I think this looks good to go

alexcrichton:
Thanks!

Issue #576, opened 2020-06-24
Custom errors


kim:
When developing custom transports, I found it a bit unsatisfying that the only way to construct a custom `git2::Error` is via `Error::from_str` -- as the code and class are very generic, it essentially always bubbles up as an "unknown error". I think this means that `transport::subtransport_action` should also call `libgit2_sys::git_error_set_str`, but if that would be the case, it would still be nice to also be able to tweak the code and class (eg. by making `Error::from_raw` public).

Would you be willing to accept a patch which does these two things?

alexcrichton:
I don't think I'd want to stabilize `from_raw` literally as-is, but providing more constructors and/or modifiers makes sense to me!

kim:
That makes sense, although I'm unsure where to draw the line. I've submitted a PR for the first part, and will ponder on the second for a bit.

kim:
Closed via #577, #579, #580

PR #577, opened 2020-06-24
Transport: set libgit2's last_error (#576)

kim:
Previously, the propagated error from a custom transport would always be the
default "unknown error" as returned by `Error::last_error` if `git_error_last`
returns a null pointer. The attached test case is rather trivial, but can be
used to reproduce the old and new behaviour.


alexcrichton:
Thanks! Should there perhaps be a specialized version of `set_err`, though, which also forwards the class/code instead of overriding it with `GIT_ERROR_NET`?

alexcrichton:
👍

PR #579, opened 2020-06-25
Add modifiers for error code and class (#576)

kim:
Arguably the most naive / liberal way to allow for custom errors. There might be reasons to only allow a subset of the error codes / classes, but I can't make up my mind which ones that should be. If that's the case, perhaps a set of specialised constructors would be preferrable, otherwise I'd lean towards amending this to define a `new` constructor which takes the code and class directly.

Aside: I noticed that the `ErrorCode` enum is missing some possible values from the binding. I'll submit that separately.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,6 +16,18 @@ pub struct Error {
 }
 
 impl Error {
+    /// Creates a new error.
+    ///
+    /// This is mainly intended for implementors of custom transports or
+    /// database backends, where it is desirable to propagate an [`Error`]
+    /// through `libgit2`.
+    pub fn new<S: AsRef<str>>(code: ErrorCode, class: ErrorClass, message: S) -> Self {
+        let mut err = Error::from_str(message.as_ref());
+        err.set_code(code);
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Could these take `&mut self` though since I think that's more idiomatic?

kim:
> I think that's more idiomatic

I'm seeing "setters" used in a few different flavours across the ecosystem, so I can't tell, really :) `mut Self -> Self`, `&mut self -> &mut Self`, and `&mut self -> ()` all yield very different ergonomics, and I chose the first because I figured that one would mostly want to construct custom errors, not actually modify them. So, what do you think of just `new(class, code, message)` or `custom(...)`?

alexcrichton:
Adding `new` also sounds good to me! Typically with what I've seen `with_x` is `self -> self` while `set_x` is just `&mut self -> ()`. Builder-style `x` is where `&mut self -> &mut self` comes into play

alexcrichton:
Thanks!

PR #580, opened 2020-06-25
Add more ErrorCode variants

kim:
It's possible that those were omitted for a reason (ie. they can't actually propagate). Please just disregard if that's the case.


alexcrichton:
👍

PR #583, opened 2020-06-29
Add bindings for `git_config_multivar_iterator_new`

kim:
Yields a slightly more convenient API over just filtering the iterator returned
by `entries`, as we don't have to unwrap the `entry` only to discard it.


alexcrichton:
👍

PR #586, opened 2020-07-05
Document IndexEntry.path a bit

pkgw:
I've just started trying to use this library and had some questions about the semantics of the `IndexEntry.path` field. I've added a brief docstring to record my findings for posterity.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -66,6 +66,21 @@ pub struct IndexEntry {
     pub id: Oid,
     pub flags: u16,
     pub flags_extended: u16,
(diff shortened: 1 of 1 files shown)


alexcrichton:
@pkgw want to add a reference and/or extra bits from the docs @ehuss linked?

pkgw:
@alexcrichton Done, plus mentioning the existence of `util::bytes2path` if one wants to see how to turn these bytes into a path.

alexcrichton:
👍

Issue #585, opened 2020-07-05
Segfault in `Index::add_all()` with empty `pathspecs` slice and `Some(cb)`


cpick:
If `Index::add_all()` is called with an empty (`&[]`) `pathspecs` slice and `Some(...)` callback then it will crash.

If `pathspecs` is set to anything (eg `&[*]`) or `cb` is set to `None` then it runs without a crash.
I suspect the issue is that it's crashing trying to create the `matched_pathspec` argument for the callback and that the type of that argument should be changed to be an `Option<&[u8]>`.

Full steps to reproduce:
```sh
$ cargo new git2-crash
$ cd git2-crash
$ echo 'git2 = "0.13.6"' >> Cargo.toml
$ cat >|src/main.rs <<EOF
fn main() {
    let pathspecs: &[&str] = &[];
    git2::Repository::open(".")
        .unwrap()
        .index()
        .unwrap()
        .add_all(pathspecs, Default::default(), Some(&mut |_path, _matched_pathspec| 0))
        .unwrap();
}
EOF
$ cargo run
    Updating crates.io index
...
   Compiling git2-crash v0.1.0 (/Users/cpick/src/delme/git2-crash)
    Finished dev [unoptimized + debuginfo] target(s) in 14.64s
     Running `target/debug/git2-crash`
Segmentation fault: 11
```

ehuss:
Yea, it looks like the [`matched_pathspec`](https://github.com/rust-lang/git2-rs/blob/4015887b56ae579a840d528955542370844772ef/src/index.rs#L590) should allow NULL as a value, per the behavior of [`git_pathspec__match`](https://github.com/libgit2/libgit2/blob/8720ae8a2c9045cb9494d328e3fc6370decbb774/src/pathspec.c#L209-L210) which uses NULL if none of the pathspecs match.

alexcrichton:
Thanks for digging in @ehuss!

I've posted https://github.com/rust-lang/git2-rs/pull/587 to at least panic instead of segfaulting, but a breaking change will be needed to fully fix this to use `Option` instead.

cpick:
I definitely understand the hesitation to introduce a breaking change and panicking is strictly better than segfaulting, but I think there may be some room for improvement on that too.

[...]

PR #587, opened 2020-07-06
Translate a segfault to a panic

alexcrichton:
At least fixes the segfault from #585, but a breaking change will be
needed to fix this at the source.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -591,10 +591,9 @@ extern "C" fn index_matched_path_cb(
     payload: *mut c_void,
 ) -> c_int {
     unsafe {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Er I need to work on the test, but the general idea is here.

ehuss:
I think the repo needs to have at least one diff entry. An untracked file or a modified file should work. 

I'm not a huge fan of `should_panic` tests, since any of those `unwrap`s could fail, which would give a false positive.

alexcrichton:
Sadly I don't have time right now to handle this, so I'm going to close this.

PR #404, opened 2019-02-25
Better Worktree Support

mkeeler:
I have a need to have support for worktrees in the library so I figured I would take a crack at implementing this.

The PR pulls in the latest upstream libgit2 release (required for some of the worktree support) and implements a bunch of worktree related functionality. I have some more polish to put on the PR but figured I would go ahead and submit the draft to get feedback before spending much more time on it. Please let me know if these changes would be acceptable and if so what a path to getting this merged should look like.

TODOs:

[ ] More thorough unit tests.
[ ] An example program.

PR #603, opened 2020-08-05
More Better Worktree Support

foriequal0:
#404 has been draft for a while. Since @mkeeler said 'Feel free to take it over'([comment](https://github.com/rust-lang/git2-rs/pull/404#issuecomment-568784984)), I've appended some commits after it.
I've intentionally not cleaned commits to preserve history and intensions. I'll clean up the history as it get reviewed.

### DIFF
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -153,7 +153,7 @@ impl<'repo> Drop for Branches<'repo> {
 mod tests {
     use crate::BranchType;
 
-    repo_test!(smoke, Typical, TypicalWorktree, BareWorktree {
+    repo_test!(smoke, (Typical, TypicalWorktree, BareWorktree), |_| {
         let (_td, repo) = crate::test::repo_init();
         let head = repo.head().unwrap();
         let target = head.target().unwrap();
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -153,8 +153,7 @@ impl<'repo> Drop for Branches<'repo> {
 mod tests {
     use crate::BranchType;
 
-    #[test]
-    fn smoke() {
+    repo_test!(smoke, Typical, TypicalWorktree, BareWorktree {
         let (_td, repo) = crate::test::repo_init();
         let head = repo.head().unwrap();
         let target = head.target().unwrap();
@@ -174,5 +173,5 @@ mod tests {
         b1.set_upstream(None).unwrap();
 
         b1.delete().unwrap();
-    }
+    });
 }
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -153,7 +153,8 @@ impl<'repo> Drop for Branches<'repo> {
 mod tests {
     use crate::BranchType;
 
-    repo_test!(smoke, (Typical, TypicalWorktree, BareWorktree), |_| {
+    #[test]
+    fn smoke() {
         let (_td, repo) = crate::test::repo_init();
         let head = repo.head().unwrap();
         let target = head.target().unwrap();
@@ -173,5 +174,5 @@ mod tests {
         b1.set_upstream(None).unwrap();
 
         b1.delete().unwrap();
-    });
+    }
 }
--- a/src/call.rs
+++ b/src/call.rs
@@ -93,12 +93,6 @@ mod impls {
         }
     }
 
-    impl Convert<*mut libc::c_char> for CString {
-        fn convert(&self) -> *mut libc::c_char {
-            self.as_ptr() as *mut libc::c_char
-        }
-    }
-
     impl<T, U: Convert<*const T>> Convert<*const T> for Option<U> {
         fn convert(&self) -> *const T {
             self.as_ref().map(|s| s.convert()).unwrap_or(ptr::null())
--- a/src/call.rs
+++ b/src/call.rs
@@ -93,6 +93,12 @@ mod impls {
         }
     }
 
+    impl Convert<*mut libc::c_char> for CString {
+        fn convert(&self) -> *mut libc::c_char {
+            self.as_ptr() as *mut libc::c_char
+        }
+    }
+
     impl<T, U: Convert<*const T>> Convert<*const T> for Option<U> {
         fn convert(&self) -> *const T {
             self.as_ref().map(|s| s.convert()).unwrap_or(ptr::null())
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2829,15 +2829,6 @@ impl Repository {
         }
     }
 
-    /// Open a worktree of a the repository
-    ///
-    /// If a repository is not the main tree but a worktree, this
-    /// function will look up the worktree inside the parent
-    /// repository and create a new `git_worktree` structure.
-    pub fn to_worktree(&self) -> Result<Worktree, Error> {
-        Worktree::open_from_repository(self)
-    }
-
     /// Creates a new worktree for the repository
     pub fn worktree(
         &self,
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -163,6 +163,19 @@ impl Repository {
         }
     }
 
+    /// Attempt to open an already-existing repository from a worktree.
+    pub fn open_from_worktree(worktree: &Worktree) -> Result<Repository, Error> {
+        crate::init();
+        let mut ret = ptr::null_mut();
+        unsafe {
+            try_call!(raw::git_repository_open_from_worktree(
+                &mut ret,
+                worktree.raw()
+            ));
+            Ok(Binding::from_raw(ret))
+        }
+    }
+
     /// Attempt to open an already-existing repository at or above `path`
     ///
     /// This starts at `path` and looks up the filesystem hierarchy
@@ -2822,12 +2835,8 @@ impl Repository {
     /// If a repository is not the main tree but a worktree, this
     /// function will look up the worktree inside the parent
     /// repository and create a new `git_worktree` structure.
-    pub fn worktree_open_from_repository(&self) -> Result<Worktree, Error> {
-        let mut raw = ptr::null_mut();
-        unsafe {
-            try_call!(raw::git_worktree_open_from_repository(&mut raw, self.raw));
-            Ok(Binding::from_raw(raw))
-        }
+    pub fn to_worktree(&self) -> Result<Worktree, Error> {
+        Worktree::open_from_repository(self)
     }
 
     /// Creates a new worktree for the repository
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2808,7 +2808,7 @@ impl Repository {
     /// Opens a worktree by name for the given repository
     ///
     /// This can open any worktree that the worktrees method returns.
-    pub fn worktree_lookup(&self, name: &str) -> Result<Worktree, Error> {
+    pub fn find_worktree(&self, name: &str) -> Result<Worktree, Error> {
         let mut raw = ptr::null_mut();
         let raw_name = CString::new(name)?;
         unsafe {
@@ -2831,7 +2831,7 @@ impl Repository {
     }
 
     /// Creates a new worktree for the repository
-    pub fn worktree_add(
+    pub fn worktree(
         &self,
         name: &str,
         path: &Path,
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -16,6 +16,7 @@ use crate::stash::{stash_cb, StashApplyOptions, StashCbData};
 use crate::string_array::StringArray;
 use crate::tagforeach::{tag_foreach_cb, TagForeachCB, TagForeachData};
 use crate::util::{self, path_to_repo_path, Binding};
+use crate::worktree::{Worktree, WorktreeAddOptions};
 use crate::CherrypickOptions;
 use crate::RevertOptions;
 use crate::{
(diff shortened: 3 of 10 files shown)


src/worktree.rs
```diff
@@ -0,0 +1,331 @@
...
+//
+pub struct Worktree {
+    raw: *mut raw::git_worktree,
+}
+
+// It is the current belief that a `Worktree` can be sent among threads, or
+// even shared among threads in a mutex
+unsafe impl Send for Worktree {}
+
+/// Options which can be used to configure how a repository is initialized
```
Aetf, line 24:
```suggestion
/// Options which can be used to configure how a worktree is initialized
```

src/worktree.rs
```diff
@@ -0,0 +1,331 @@
...
+        let mut raw = ptr::null_mut();
+        unsafe {
+            try_call!(raw::git_worktree_open_from_repository(&mut raw, repo.raw()));
+            Ok(Binding::from_raw(raw))
+        }
+    }
+
+    /// Retrieves the name of the worktree
+    ///
+    /// This is the name that can be passed to repo::Repository::worktree_lookup
```
Aetf, line 60:
```suggestion
    /// This is the name that can be passed to repo::Repository::find_worktree
```

src/worktree.rs
```diff
@@ -0,0 +1,331 @@
...
+                0 => Ok(WorktreeLockStatus::Unlocked),
+                _ => {
+                    println!("Buf: {}", buf.as_str().unwrap());
+                    let v = buf.to_vec();
+                    println!("Length of v: {}", v.len());
+                    Ok(WorktreeLockStatus::Locked(match v.len() {
+                        0 => None,
+                        _ => String::from_utf8(v).ok(),
+                    }))
+                }
```
Aetf, line 123: `git_worktree_is_locked` can return <0 when error and >0 for the locked state, which I think should be handled separately. Also, the `println`s are still there ;)

src/worktree.rs
```diff
@@ -0,0 +1,331 @@
...
+            try_call!(raw::git_worktree_prune(self.raw, opts.as_mut()));
+        }
+        Ok(())
+    }
+
+    /// Checks if the worktree is prunable
+    pub fn is_prunable(&self, opts: Option<&WorktreePruneOptions>) -> bool {
+        unsafe {
+            let mut opts = opts.map(|o| o.raw());
+            if call!(raw::git_worktree_is_prunable(self.raw, opts.as_mut())) <= 0 {
```
Aetf, line 143: `git_worktree_is_prunable` can return <0 for error. Maybe the return type should be `Result<bool, Error>`?

src/call.rs
```diff
@@ -93,6 +93,12 @@ mod impls {
         }
     }
 
+    impl Convert<*mut libc::c_char> for CString {
+        fn convert(&self) -> *mut libc::c_char {
+            self.as_ptr() as *mut libc::c_char
+        }
+    }
+
```
Aetf, line 101: The pointer returned by `as_ptr` should never be written to. See https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr

src/worktree.rs
foriequal0: `try_call!` macro handles when the return value is `< 0` by calling `c_try`

kamalmarhubi:
@foriequal0 could you add a rough todo list of what needs to happen for this to be done? I might have some time to put in, and as a heavy worktree user I'd love this to make its way into git2-rs based tools. :-)

foriequal0:
I rebased it to resolve merge conflict.

@kamalmarhubi I think it is done. I forgot this while waiting for a review.

kamalmarhubi:
ping @joshtriplett or @alexcrichton as you've both recently merged prs

src/repo.rs
```diff
@@ -162,6 +163,19 @@ impl Repository {
         }
     }
 
+    /// Attempt to open an already-existing repository from a worktree.
+    pub fn open_from_worktree(worktree: &Worktree) -> Result<Repository, Error> {
+        crate::init();
```
alexcrichton, line 168: I think this can be skipped since the presence of `Worktree` implies it's already been called.

systest/build.rs
```diff
@@ -41,5 +43,6 @@ fn main() {
     cfg.skip_roundtrip(|t| t == "git_clone_options" || t == "git_submodule_update_options");
 
     cfg.skip_type(|t| t == "__enum_ty");
+    cfg.skip_type(|t| t == "git_transfer_progress");
```
alexcrichton, line 46: How come this is skipped?

systest/build.rs
```diff
@@ -22,7 +22,9 @@ fn main() {
     });
     cfg.skip_field(|struct_, f| {
         // this field is marked as const which ctest complains about
-        struct_ == "git_rebase_operation" && f == "id"
+        (struct_ == "git_rebase_operation" && f == "id") ||
+        // the real name of this field is ref but that is a reserved keyword
+        (struct_ == "git_worktree_add_options" && f == "reference")
```
alexcrichton, line 27: Could `r#ref` be used as a field name instead?

src/test.rs
```diff
@@ -17,6 +18,118 @@ macro_rules! t {
     };
 }
 
+// `repo_test! will
+macro_rules! repo_test {
```
alexcrichton, line 22: I may be missing something, but how come a macro for this was added? Could the tests be written just against the right function?

src/test.rs
```diff
@@ -17,6 +18,118 @@ macro_rules! t {
...
+    let worktree_repo = worktree.to_repository().unwrap();
+
+    let tds = TempDirs {
+        main: worktree_td,
+        _rest: vec![td],
+    };
+    (tds, worktree_repo)
+}
+
+pub fn repo_init_typical_worktree() -> (TempDirs, Repository) {
```
alexcrichton, line 106: Since these functions seem related to worktree-testing, perhaps they could be scoped to the worktree tests submodule?

src/worktree.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+
+    /// Validates the worktree
+    ///
+    /// This checks that it still exists on the
+    /// filesystem and that the metadata is correct
+    pub fn validate(&self) -> Result<(), Error> {
+        unsafe {
+            call::c_try(raw::git_worktree_validate(call::convert(&self.raw)))?;
```
alexcrichton, line 87: Could this use the `try_call!` macro instead of the `call` module directly?

alexcrichton: (same below)

src/worktree.rs
```diff
@@ -0,0 +1,333 @@
...
+    pub fn is_locked(&self) -> Result<WorktreeLockStatus, Error> {
+        let buf = Buf::new();
+        unsafe {
+            match try_call!(raw::git_worktree_is_locked(buf.raw(), self.raw)) {
+                0 => Ok(WorktreeLockStatus::Unlocked),
+                _ => {
+                    let v = buf.to_vec();
+                    Ok(WorktreeLockStatus::Locked(match v.len() {
+                        0 => None,
+                        _ => String::from_utf8(v).ok(),
```
alexcrichton, line 119: Could this use `.unwrap()` since it's expected to always be utf-8?

src/worktree.rs
```diff
@@ -0,0 +1,333 @@
...
+        unsafe {
+            let mut opts = opts.map(|o| o.raw());
+            let rv = try_call!(raw::git_worktree_is_prunable(self.raw, opts.as_mut()));
+            Ok(rv != 0)
+        }
+    }
+
+    /// Opens the repository from the worktree
+    pub fn to_repository(&self) -> Result<Repository, Error> {
+        Repository::open_from_worktree(self)
```
alexcrichton, line 148: I personally prefer to stick close to the original `git2` API where possible, so could this and the other convenience alias be removed?

src/worktree.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// interior of this structure
+    pub unsafe fn raw(&self) -> raw::git_worktree_add_options {
+        let mut opts = mem::zeroed();
+        assert_eq!(
+            raw::git_worktree_add_options_init(&mut opts, raw::GIT_WORKTREE_ADD_OPTIONS_VERSION),
+            0
+        );
+
+        opts.lock = self.lock as c_int;
+        opts.reference = crate::call::convert(&self.reference.as_ref().map(|o| o.raw()));
```
alexcrichton, line 187: Instead of the `convert` function could this do the conversion manually?

Also I think it might be good to store the `git_worktree_add_options` directly in this structure.

src/worktree.rs
```diff
@@ -0,0 +1,333 @@
...
+            self.flags &= !(flag as u32);
+        }
+        self
+    }
+    /// Creates a set of raw prune options to be used with `git_worktree_prune`
+    ///
+    /// This method is unsafe as the returned value may have pointers to the
+    /// interior of this structure
+    pub unsafe fn raw(&self) -> raw::git_worktree_prune_options {
+        let mut opts = mem::zeroed();
```
alexcrichton, line 237: Could this `opts` be stored directly in the structure?

alexcrichton (commented):
Looking good, thanks for picking this up!

systest/build.rs
foriequal0: I didn't write this commit. I think @mkeeler wanted to turn off deprecation warning according to the commit message.

systest/build.rs
foriequal0: ctest depends on old syntex_syntax2, which doesn't know about raw identifier. It complaints that "found invalid character; only `#` is allowed in raw string delimitation"

src/test.rs
foriequal0: I asume that @mjkeeler wanted to make sure that same code should have same functionalities even if underlying repositories are different (normal repo, normal repo's worktree, bare repo, bare repo's worktree). I've finished his implementation, but I still think it is the libgit2's responsibility since git2-rs is just a binding. Should I remove this?

src/branch.rs
```diff
@@ -153,8 +153,7 @@ impl<'repo> Drop for Branches<'repo> {
 mod tests {
     use crate::BranchType;
 
-    #[test]
-    fn smoke() {
+    repo_test!(smoke, (Typical, TypicalWorktree, BareWorktree), |_| {
```
alexcrichton, line 156: I think we probably don't need to update old tests like this

src/test.rs
alexcrichton: I think it's good to keep some tests in this repository, but whether or not it's done with a macro is up to you. I personally prefer to avoid macros where possible.

src/worktree.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    /// If enabled, this will cause the newly added worktree to be locked
+    pub fn lock(&mut self, enabled: bool) -> &mut WorktreeAddOptions<'a> {
+        self.raw.lock = enabled as c_int;
+        self
+    }
+
+    /// reference to use for the new worktree HEAD
+    pub fn reference(&mut self, reference: Option<&Reference<'a>>) -> &mut WorktreeAddOptions<'a> {
+        self.raw.reference = if let Some(reference) = reference {
```
alexcrichton, line 174: I'm not sure how safe this is in the sense that the `Reference` can be dropped before these options are used. I think that there needs to be a lifetime connection or this structure needs to take ownership of the `Reference`

systest/build.rs
alexcrichton: Ah, that's ok then.

src/branch.rs
foriequal0: I'll revert old tests, but keep worktree tests. I'll move `repo_test` macro to `worktree.rs`

src/worktree.rs
foriequal0: I've made sure that `Reference` will outlive `WorktreeAddOptions` by changing it to `pub fn reference(&mut self, reference: Option<&'a Reference<'_>>) -> &mut WorktreeAddOptions<'a>`

kamalmarhubi:
Looks like review comments were all addressed, is that right @foriequal0?

src/worktree.rs
```diff
@@ -0,0 +1,335 @@
...
+///
+/// This structure corresponds to a `git_worktree` in libgit2.
+//
+pub struct Worktree {
+    raw: *mut raw::git_worktree,
+}
+
+// It is the current belief that a `Worktree` can be sent among threads, or
+// even shared among threads in a mutex
+unsafe impl Send for Worktree {}
```
alexcrichton, line 22: Oh sorry I think I missed this earlier, but are you sure of this? Is this documented in libgit2?

src/worktree.rs
alexcrichton: Looks good to me, thanks!

systest/build.rs
alexcrichton: I think this may still want to get removed?

alexcrichton (commented):
Just a few more small comments!

src/worktree.rs
kamalmarhubi: This is the best I could find:
> Unless otherwise specified, libgit2 objects cannot be safely accessed by multiple threads simultaneously.
> [...]
> Use an object from a single thread at a time. Most data structures do not guard against concurrent access themselves. This is because they are rarely used in isolation and it makes more sense to synchronize access via a larger lock or similar mechanism.

—https://github.com/libgit2/libgit2/blob/master/docs/threading.md

I think this can be read as they are Send but not Sync?

kamalmarhubi: The original version of that file was clearer on this point:
> You may safely use any libgit2 object from any thread, though there may be issues depending on the cryptographic libraries libgit2 or its dependencies link to (more on this later).

—https://github.com/libgit2/libgit2/commit/15bea02c04ebf3172d495f14219b30eba262eb6a#diff-bb59cfb2bcd0e444b33e25dc3f1e127e095d72379150cc82d408afa63b9903bf

joshtriplett:
On Mon, Nov 16, 2020 at 10:30:06AM -0800, Kamal Marhubi wrote:
> @kamalmarhubi commented on this pull request.
> 
> 
> 
> > +use std::path::Path;
> +use std::ptr;
> +use std::str;
> +use std::{marker, mem};
> +
> +/// An owned git worktree
> +///
> +/// This structure corresponds to a `git_worktree` in libgit2.
> +//
> +pub struct Worktree {
> +    raw: *mut raw::git_worktree,
> +}
> +
> +// It is the current belief that a `Worktree` can be sent among threads, or
> +// even shared among threads in a mutex
> +unsafe impl Send for Worktree {}
> 
> This is the best I could find:
> > Unless otherwise specified, libgit2 objects cannot be safely accessed by multiple threads simultaneously.
> > [...]
> > Use an object from a single thread at a time. Most data structures do not guard against concurrent access themselves. This is because they are rarely used in isolation and it makes more sense to synchronize access via a larger lock or similar mechanism.
> 
> —https://github.com/libgit2/libgit2/blob/master/docs/threading.md
> 
> I think this can be read as they are Send but not Sync?

Not necessarily. `Send` would mean you can share a read-only reference
to them across threads. That would in turn mean you can call functions
that only need a read-only reference. But in some cases, libgit2's
structures contain internal mutability (on the C side), which means even
"read-only" methods are not safe to call from multiple threads.

Marking any libgit2 structure `Send` means carefully checking for any
such internal mutability, and having high confidence that no new
internal mutability will be introduced in the future.

src/worktree.rs
alexcrichton: Yeah as @joshtriplett [pointed out below](https://github.com/rust-lang/git2-rs/pull/603#issuecomment-728262191) libgit2's policy isn't enough to dictate whether things are `Send` or `Sync` in Rust. You'll need to devle into the internals and audit what's happening to see if it's safe. I think I did that for `git_repository` a long time ago, but if this wants to stay then it should happen for `git_worktree` as well. It's always ok to drop this `unsafe impl` for now though and add it back later if it works.

foriequal0: You're right. I didn't notice that. I'll remove it.

foriequal0:
Thanks for the review! I think you can squash merge if it is okay or I should rebase commits before merging them.

alexcrichton:
OK everything looks great to me, thanks again! And no worries, I use the squash-and-merge button so that happens automatically.

joshtriplett:
@foriequal0 To the best of your knowledge, does this fully supersede https://github.com/rust-lang/git2-rs/pull/404 , or is there additional work in that PR that needs updating and merging?

foriequal0:
@joshtriplett I think there is no additional work for this :)

PR #604, opened 2020-08-05
Make the calls to `crate::init()` consistent

lhchavez:
This change makes the calls to `crate::init()` in `src/repo.rs` all use
the full namespace for consistency.

This makes no difference, but makes it easier to find (e.g. through a
simple `grep`) which functions eventually invoke `git_libgit2_init()`.


alexcrichton:
Thanks!

Issue #605, opened 2020-08-15
Provide binding to git_libgit2_opts


joshtriplett:
I'd like to be able to call `git_libgit2_opts` using `git2-rs`.

By default, when reading in objects, libgit2 computes their sha1 hash and confirms that it matches the expected hash. That's a reasonable default, but it costs a *lot* of performance. Calling `git_libgit2_opts(GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION, 0)` can substantially improve the performance of reading objects.

`git_libgit2_opts` is a varargs function, and the arguments depend on the option. Thus, I'd suggest type-safe bindings for individual options.

taralx:
I was just looking for exactly the same thing... Consider this my 👍 .

extrawurst:
This will not be possible on stable rust until https://github.com/rust-lang/rust/issues/44930 is stabilised, right?

taralx:
At least for the existing options, we could write C wrappers that provide non-variadic versions?

taralx:
Also `GIT_OPT_ENABLE_STRICT_HASH_VERIFICATION` is definitely the big one, so perhaps just supporting that for now? Maybe I'll put together a PR real quick.

joshtriplett:
@extrawurst That feature is for defining C-compatible variadic functions in Rust. Stable Rust supports *calling* C variadic functions, without any wrapper needed.

@taralx Thanks for the PR! Could you please adapt it to call the variadic function directly, without a C wrapper?

DanielEScherzer:
It looks like the PR was manually merged as 20b17131062dcd311f4b1c01458d59c95fc35d46 - I think this can be closed?

PR #607, opened 2020-08-15
Add git2::opts::strict_hash_verification

taralx:
Addresses #605 for the immediate use case.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -3759,10 +3759,7 @@ extern "C" {
         given_opts: *const git_revert_options,
     ) -> c_int;
 
-    pub fn git_libgit2_opts(
-        option: c_int,
-        ...,
-    ) -> c_int;
+    pub fn git_libgit2_opts(option: c_int, ...) -> c_int;
 }
 
 pub fn init() {
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -1778,6 +1778,42 @@ git_enum! {
     }
 }
 
+git_enum! {
+    pub enum git_libgit2_opt_t {
+        GIT_OPT_GET_MWINDOW_SIZE = 0,
+        GIT_OPT_SET_MWINDOW_SIZE,
+        GIT_OPT_GET_MWINDOW_MAPPED_LIMIT,
+        GIT_OPT_SET_MWINDOW_MAPPED_LIMIT,
+        GIT_OPT_GET_SEARCH_PATH,
+        GIT_OPT_SET_SEARCH_PATH,
+        GIT_OPT_SET_CACHE_OBJECT_LIMIT,
+        GIT_OPT_SET_CACHE_MAX_SIZE,
+        GIT_OPT_ENABLE_CACHING,
+        GIT_OPT_GET_CACHED_MEMORY,
+        GIT_OPT_GET_TEMPLATE_PATH,
+        GIT_OPT_SET_TEMPLATE_PATH,
+        GIT_OPT_SET_SSL_CERT_LOCATIONS,
+        GIT_OPT_SET_USER_AGENT,
+        GIT_OPT_ENABLE_STRICT_OBJECT_CREATION,
+        GIT_OPT_ENABLE_STRICT_SYMBOLIC_REF_CREATION,
+        GIT_OPT_SET_SSL_CIPHERS,
+        GIT_OPT_GET_USER_AGENT,
+        GIT_OPT_ENABLE_OFS_DELTA,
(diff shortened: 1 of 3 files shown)


taralx:
(I have no idea what Travis is complaining about.)

joshtriplett:
You don't need a C wrapper; you can call a C variadic function directly from stable Rust.

Also, please don't just assert; please check the error and return a `git2::Error`, as other functions do.

Thanks!

taralx:
> You don't need a C wrapper; you can call a C variadic function directly from stable Rust.

Not according to https://github.com/rust-lang/rust/issues/44930?

taralx:
Ah, never mind, I see that it's the other way around. Okay, I'll adjust this.

I used a debug assert because this function actually cannot fail. It only returns an error because the interface says it has to.

taralx:
That CI failure is entirely bizarre. It passes locally for me (windows, rust 1.45.2).

taralx:
I'd really rather not return an error here -- it makes the ergonomics of using the thing significantly worse.

taralx:
Also, any idea why the enums are typed as unsigned int and not regular int?

alexcrichton:
I think that this function may need to be `unsafe` because when I last glanced at libgit2 I don't think it did any internal synchronization when these global options are set?

taralx:
I thought about that. We could put a global mutex in if you really want, but there isn't any way to trigger unsafe behavior with this particular option.

I don't think marking it unsafe is the right answer though.

Thoughts?

joshtriplett:
@taralx Since we're providing bindings to particular options, I think we can determine the right way to handle each option on a case-by-case basis. And I agree with your analysis that this particular function doesn't need to be `unsafe`.

alexcrichton:
I'm not sure why I understand why this wouldn't need to be `unsafe`. This is a data race if it's not synchronized with anything else. I don't think a mutex is a great solution since we'd basically need an rwlock for the entire library.

The write [here](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/settings.c#L276) is not atomic nor is the [read here](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/odb.c#L1079-L1087) synchronized at all. If we want this to be safe then I think we should send a patch to libgit2 to make these atomic reads and writes. Otherwise we can't paper over the fact that this is a possible data race.

To be clear I agree this *shouldn't* be unsafe, but my point is that PR exactly as-is is unsafe.

joshtriplett:
@alexcrichton There's no synchronization around the flag, so if you write it from one thread while another thread reads it, it might get the old value or the new value. So if you turn strict hash verification on or off while doing operations that might reference it, you may or may not get strict hash verification for that particular operation. That's easily solved by calling the function before you do anything else if you want deterministic behavior, but if you don't, it still doesn't seem like it needs `unsafe`. No safety issue will occur if you call it at an arbitrary time.

Global state seems like a bad way to handle this, and it *should* be part of some git2 object that encapsulates such state, but the net effect here for this particular option does not seem problematic.

(By comparison, setting an option like `GIT_OPT_SET_ALLOCATOR` concurrently seems like an incredibly bad idea and would absolutely need `unsafe`.)

alexcrichton:
Sorry so to be clear again, I don't *want* this to be `unsafe`. I don't think it matters what value a thread gets so long as it has reasonable behavior. The actual literal assembly probably works just fine today as well. My point is that this matches the definition of a data race. Two unsynchronized memory acceses to the same location where at least one is a write. 

My comments are about the literal implementation of libgit2 as-is today. There's no escaping that this is exposing a possible data race in safe Rust code. Until that's fixed I think this should be `unsafe`, or it should be a feature request in libgit2 to make it atomic and properly thread safe.

taralx:
I would argue that this matches the definition of a *benign* data race.

I suppose we could try patching the underlying library to use relaxed atomics -- but are those even available in C?

alexcrichton:
AFAIK data races are unconditional undefined behavior? While it may be ok today there's no guarantee it's ok tomorrow?

joshtriplett:
I discussed this further with @alexcrichton on Zulip. This doesn't introduce a race in Rust; it calls into libgit2, which is responsible for handling this. libgit2's current approach to this particular class of data races is the common C approach, namely "is there an actual system on which this wouldn't work" (to which the answer is "no"). libgit2 could potentially switch to C11 relaxed atomics, except that libgit2 tries to be C90, and such a change would involve substantial additional compiler feature detection. It's also worth noting that libgit2 currently has *many* internal instances of such code.

joshtriplett:
Squashed and merged.

PR #608, opened 2020-08-16
Use cc 1.0.43 or later

lo48576:
cc-1.0.42 should not be used as minimal dependency verison, because it could hang on some environment during the build.

I use `cargo +nightly update -Z minimal-versions` on CI to generate Cargo.lock for minimal depnedency versions and check if Cargo.toml is OK.
However, I confirmed that compilation using cc 1.0.42 hangs on rust:latest docker (maybe Debian?).
I think this is https://github.com/alexcrichton/cc-rs/issues/440, and this version should be avoided.


alexcrichton:
👍

PR #610, opened 2020-08-17
feature: enable building git2-curl with vendored openssl

corbinu:
the main crate can be build with vendored ssl git2-curl can not. This adds that feature

### DIFF
--- a/git2-curl/Cargo.toml
+++ b/git2-curl/Cargo.toml
@@ -25,7 +25,7 @@ conduit-git-http-backend = "0.8"
 tempfile = "3.0"
 
 [features]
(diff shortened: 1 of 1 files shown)


ehuss:
Can you explain more why you would need this?  In what scenario do you depend on git2-curl without also depending on git2?

corbinu:
@ehuss Sorry about that you are right my fault. I am trying to update cargo-outdated to compile with vendored SSL. Something I recently did with with cargo-update. cargo-outdated depends on git2 directly where as cargo-outdated only depends on git2-curl. I was thinking I needed this but in reality it was just needed to build cargo itself with vendored-openssl as that is where the git2 dep is coming from

PR #611, opened 2020-08-17
feat: add git_submodule_clone

### LINKED ISSUE #609: Correctly working with submodules
I have some trouble when trying to clone a submodule. The documentation mentions that to fully emulate `git submodule add` I'd have to manually open the repo and run clone but I couldn't figure out how. According to the docs calling `Repository#submodule` does already initialize a repository at the given submodule folder but after opening the submodule repository, the `Repository` instance does sadly not provide a clone method that I can use.

I run the following code:
```rust
// Open the root git repository to allow for submodule registration.
let repo = Repository::open(&workspace_path).unwrap();
let submodule_rel_path = xtools_crate_dir.join(repo_name_from_git_url(crate_name));
let submodule_path = Path::new(submodule_rel_path.as_str());
let mut submodule = repo.submodule(crate_name, &submodule_path, false).expect("failed to add submodule");

submodule.sync().expect("failed to sync");
//submodule.add_to_index(true).expect("failed to write to index");
submodule.update(true, None).expect("failed to init submodule"); // 1.
```

The last line failed with an error message `could not get ID of submodule` so that's why I thought I might have to add it to the index to make it work (the line above). But that resulted in another error message: `cannot add submodule without HEAD to index`. Which makes probably sense as the module hasn't been checked out.

I am probably looking for the equivalent of libgit2's [git_submodule_clone](https://libgit2.org/libgit2/#HEAD/group/submodule/git_submodule_clone). From my intuition I'd say that [Submodule#update](https://docs.rs/git2/0.13.8/git2/struct.Submodule.html#method.update) is indeed the function that I should call here. But it sadly doesn't work.

SirWindfield:
Closes #609

### DIFF
--- a/src/submodule.rs
+++ b/src/submodule.rs
@@ -33,6 +33,23 @@ impl<'repo> Submodule<'repo> {
         unsafe { crate::opt_bytes(self, raw::git_submodule_branch(self.raw)) }
     }
 
+    ///
+    pub fn clone(
+        &mut self,
+        repo: Option<&mut &mut Repository>,
+        opts: &mut SubmoduleUpdateOptions<'_>,
+    ) -> Result<(), Error> {
+        unsafe {
+            let mut raw_repo = repo.map(|o| o.raw());
+            try_call!(raw::git_submodule_clone(
+                raw_repo.as_mut().map_or(ptr::null_mut(), |o| o),
+                self.raw().as_mut(),
+                &opts.raw() as *const _
+            ));
+        }
+        Ok(())
+    }
+
     /// Get the submodule's url.
     ///
     /// Returns `None` if the url is not valid utf-8 or if the URL isn't present
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2288,6 +2288,11 @@ extern "C" {
     ) -> c_int;
     pub fn git_submodule_add_to_index(submodule: *mut git_submodule, write_index: c_int) -> c_int;
     pub fn git_submodule_branch(submodule: *mut git_submodule) -> *const c_char;
+    pub fn git_submodule_clone(
+        repo: *mut *mut git_repository,
+        submodule: *mut git_submodule,
+        opts: *const git_submodule_update_options,
+    ) -> c_int;
     pub fn git_submodule_foreach(
         repo: *mut git_repository,
         callback: git_submodule_cb,


src/submodule.rs
```diff
@@ -33,6 +33,23 @@ impl<'repo> Submodule<'repo> {
         unsafe { crate::opt_bytes(self, raw::git_submodule_branch(self.raw)) }
     }
 
+    ///
+    pub fn clone(
+        &mut self,
+        repo: Option<&mut &mut Repository>,
+        opts: &mut SubmoduleUpdateOptions<'_>,
```
SirWindfield, line 40: There is probably a nicer way to create the public API but I have zero experience with raw bindings :(

SirWindfield:
Not sure if this is a faulty implementation or not, but calling `submodule.open()` actually failed after I called this clone implementation. No error message, just `()` was displayed inside the stacktrace.

src/submodule.rs
ehuss: Instead of passing the `repo` parameter, it should return a `Repository` as a return value.  It should look roughly the same as [`Submodule::open`](https://github.com/rust-lang/git2-rs/blob/c2a26133731d152f67a4fb11b35c4e139441e60c/src/submodule.rs#L113-L123), plus the `opts` parameter.

You can grab the docs from the [header file](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/include/git2/submodule.h#L289-L299), and just adjust them similar to how the other methods do.

It would also be nice to have a test for it. There are a few at the bottom of the file, and you can probably just copy one of those.

SirWindfield: That's what I wasn't sure off. I haven't really touched any C code in my life so it's quite hard to get to know the calling conventions. I thought that it was an optional parameter that was needed to function.

https://libgit2.org/libgit2/#HEAD/group/submodule/git_submodule_clone

But after re-reading it seems that it is actually the output value.

src/submodule.rs
```diff
@@ -33,6 +33,21 @@ impl<'repo> Submodule<'repo> {
         unsafe { crate::opt_bytes(self, raw::git_submodule_branch(self.raw)) }
     }
 
+    /// Perform the clone step for a newly created submodule.
+    ///
+    /// This performs the necessary `git_clone` to setup a newly-created submodule.
+    pub fn clone(&mut self, opts: &mut SubmoduleUpdateOptions<'_>) -> Result<Repository, Error> {
```
ehuss, line 39: I think `opts` should probably be an `Option`, similar to how [`update`](https://github.com/rust-lang/git2-rs/blob/c2a26133731d152f67a4fb11b35c4e139441e60c/src/submodule.rs#L186-L201) is done.

src/submodule.rs
```diff
@@ -33,6 +33,21 @@ impl<'repo> Submodule<'repo> {
...
+        unsafe {
+            let mut raw_repo = ptr::null_mut();
+            try_call!(raw::git_submodule_clone(
+                &mut raw_repo,
+                self.raw().as_mut(),
```
ehuss, line 44: I think this can be accessed directly.

```suggestion
                self.raw,
```

src/submodule.rs
SirWindfield: According to the docs the values isn't optional though: https://libgit2.org/libgit2/#v1.0.1/group/submodule/git_submodule_clone.

What we could do is create a default instance if the Option value is `None`. If that is Ok with you.

src/submodule.rs
SirWindfield: Tested locally and it works.

src/submodule.rs
ehuss: The libgit2 docs aren't consistent or necessarily complete, and in this case I think it is fine to pass NULL.  The code [handles NULL](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/submodule.c#L845), and I think it would be unlikely that they would remove that.

Another example, [`git_rebase_open`](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/include/git2/rebase.h#L210) doesn't mention NULL is supported, but it is, and the rust bindings [can pass NULL](https://github.com/rust-lang/git2-rs/blob/c2a26133731d152f67a4fb11b35c4e139441e60c/src/repo.rs#L1987).

SirWindfield: I did the change but that's where I hit my knowledge limitations. I can't get it to compile properly. The whole pointer stuff doesn't work out. I do know why (because `*mut git_submodule_opts` is not the same as the type of `ptr::null()`) but I don't know how to fix this :(

ehuss: The issue here is that `SubmoduleUpdateOptions::raw` returns a *value* and not a pointer.  `ptr::null` is a pointer, so the types aren't the same.  The call to `git_submodule_clone` needs a pointer, so the solution is to place the value in a (temporary) local variable, and pass a pointer to it, something like this:

```rust
        unsafe {
            let mut raw_repo = ptr::null_mut();
            let raw_opts = opts.map(|o| o.raw());
            try_call!(raw::git_submodule_clone(
                &mut raw_repo,
                self.raw,
                raw_opts.as_ref()
            ));
            Ok(Binding::from_raw(raw_repo))
        }
```

The call to `as_ref` turns it to an `Option<&git_submodule_update_options>`, which is a nullable pointer. `Option<T>` has a guarantee that None will translate to NULL for ffi bindings.

I would suggest to also update the unit test to use None instead of default, just to double check that. 😄

SirWindfield: > The call to `as_ref` turns it to an `Option<&git_submodule_update_options>`, which is a nullable pointer. `Option<T>` (for certain types of T) has a guarantee that None will translate to NULL for ffi bindings.

Does this hold true for all ffi bindings or is that something that has been setup in git2?

SirWindfield:
Thanks for taking the time to help me out @ehuss! :)

src/submodule.rs
ehuss: > Does this hold true for all ffi bindings or is that something that has been setup in git2?

This is mostly macro magic in the git2 library.  The [`try_call`](https://github.com/rust-lang/git2-rs/blob/c2a26133731d152f67a4fb11b35c4e139441e60c/src/call.rs#L12-L19) does a bunch of implicit conversions (in this case converting `Option<&git_submodule_update_options>` to `*const git_submodule_update_options`), which handles the conversion of None to null (around [here](https://github.com/rust-lang/git2-rs/blob/c2a26133731d152f67a4fb11b35c4e139441e60c/src/call.rs#L96-L100) I think).

There is a guarantee in some cases that Option can represent a null (see [nullable pointer optimization](https://doc.rust-lang.org/nomicon/ffi.html#the-nullable-pointer-optimization) for more detail), but I believe this particular scenario is not leveraging that.

PR #612, opened 2020-08-18
Add support for zlib-ng

joshtriplett:
Depend on libz-sys 1.1.0, and provide "zlib-ng-compat" features to opt
into zlib-ng. Use feature dependencies to ensure that either all C
libraries opt in or none do.

This provides a substantial performance boost when compressing or
decompressing objects in a git2 repository.

(This is blocked on https://github.com/alexcrichton/curl-rust/pull/351
and https://github.com/alexcrichton/ssh2-rs/pull/197 , and subsequent
publishes of both. CI won't pass until then.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ url = "2.0"
 bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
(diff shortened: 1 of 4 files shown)


joshtriplett:
Updated git2-curl to a newer curl, with the fix for https://github.com/alexcrichton/curl-rust/issues/352 .

PR #614, opened 2020-08-19
Fix dangling pointer in format_email.

### LINKED ISSUE #613: test diff::tests::format_email_simple fails, only in release mode
While testing git2, I noticed that `cargo test --release` fails while `cargo test` succeeds. The test `diff::tests::format_email_simple` fails, because the generated email `From ` line starts with:
```
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
It has an all-zero hash instead of the actual commit hash. It's especially disturbing that this only happens in an optimized build.

This failure happens with either `rustc 1.45.2 (d3fb005a3 2020-07-31)` or `rustc 1.47.0-nightly (8e21bd063 2020-08-14)`.

ehuss:
The code was creating an Oid object, taking a pointer to it, and then immediately dropping it.  The fix is to use a local variable to hold the value.

Fixes #613

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -266,7 +266,8 @@ impl<'repo> Diff<'repo> {
         message = &message[summary.len()..];
         raw_opts.patch_no = patch_no;
         raw_opts.total_patches = total_patches;
-        raw_opts.id = commit.id().raw();
+        let id = commit.id();
+        raw_opts.id = id.raw();
         raw_opts.summary = summary.as_ptr() as *const _;
         raw_opts.body = message.as_ptr() as *const _;
         raw_opts.author = commit.author().raw();


joshtriplett:
@ehuss Thanks for the fix! The same issue would apply to `commit.author().raw()` a few lines later, right?

ehuss:
> The same issue would apply to `commit.author().raw()` a few lines later, right?

I don't think so.  The `Signature` returned from `commit.author()` is not owned (see [`from_raw_const`](https://github.com/rust-lang/git2-rs/blob/f6e90a79f6521427a5067c1ca7a1c519bcbaebb6/src/signature.rs#L123-L129)).  Thus when the `Signature` is dropped, the raw pointer is still alive.

joshtriplett:
You're right. I was trying to see if there were other instances of this pattern in the codebase, and searched for `().raw` in the process.

I wonder if there's some reasonable way we could get the compiler to identify issues like this, even though they use raw pointers.

PR #619, opened 2020-09-11
Compile libgit2 deterministically by sorting read_dir

### LINKED ISSUE #618: filesystem-related nondeterminism from libgit2-sys
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
when building below packages, there were slight differences between each build:

* bat
* cargo-c
* exa
* onefetch

from libgit2-sys
e.g. around `git_attr_get_many_with_session.cold`

My analysis showed that it is from filesystem-ordering.

I see ordering issues in the build env
/home/abuild/rpmbuild/BUILD/bat-0.15.4/target/release/build/libgit2-sys-c2cd739238a5d529/output
where .h files change order.

kpcyrd:
It seems the order that C files are passed to `build.file(&path);` has impact on the binary. Right now build.rs is using a simple `read_dir`, so the final binary varies depending on the order the files are returned by the filesystem of the build machine. This patch collects into a Vec first and then sorts it to ensure the order of `build.file(&path);` calls is always deterministic.

Fixes #618

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::fs;
+use std::io;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #620, opened 2020-09-15
Add bindings for `git_reference_normalize_name`

kim:
Occasionally useful to be able to specify the `git_reference_format_t` flags.

The effect of the flags is not always intuitive -- let me know if the examples
given are too much, or too less.


kim:
Hm, not sure why Windows CI is complaining.

ehuss:
enums don't necessarily have a specific integer type in C.  The `flags` parameter in `git_reference_normalize_name` needs to match the type in the C definition (`unsigned int`).

Also, run `cargo run --manifest-path systest/Cargo.toml` to verify the types are defined correctly.

kim:
Thanks @ehuss!

alexcrichton:
Looks good to me, thanks!

PR #621, opened 2020-09-17
Init before creating Odb with `new`

stoically:
This allows calling `Odb::new()` as first thing in a code base


alexcrichton:
Thanks!

Issue #623, opened 2020-09-20
Rustls transport?


jsha:
Hi,

I've been looking at the security of the cargo / crates.io ecosystem, and I've noticed that it depends heavily on the security of TLS. The index of crates is downloaded over HTTPS from GitHub, and serves as the root of trust for subsequent crate downloads, because it contains the sha256 hashes of the crates.

Given that, I think it makes sense to use the strongest TLS implementation. Fortunately, rustls is a very high quality implementation, and is in Rust! Besides the memory safety benefits, it has a couple of other benefits: It does not implement older TLS versions (1.0 and 1.1), which have some vulnerabilities. Similarly it does not implement weaker ciphersuites. Also, it ships by default with [webpki-certs](https://crates.io/crates/webpki-roots), which is beneficial because it tends to be more up-to-date than the native root store on Debian-derived operating systems. When there's a CA that's been removed from the Mozilla root program (and others), but not removed yet from a root store, that's a big risk. In that situation there aren't any audits or controls happening any longer to ensure the CA is doing the right thing.

So I was thinking it would be useful to implement a git transport for HTTPS that uses Rustls. Probably the most straightforward way to do that would be to incorporate a Rust HTTP client. I see git2-rs already has a small HTTP client in git2-curl, but that it's considered less battle-tested (and has an issue with pulling a whole repository into memory), so isn't used as the default in cargo, but is only used when there are special HTTP setting, like proxies or custom root stores.

Is that something you think would be useful? Since my goal would be for cargo to (eventually) adopt this as the default instead of libgit's built-in HTTP transport, what quality and testing bar would it need to meet?

Thanks,
Jacob

alexcrichton:
Seems reasonable to me! There's the `git2-curl` crate which implements all HTTP crate fetches through the `curl` crate, so presumably there could be a `git2-hyper` crate (or something like that) which disables all of libgit2's internal networking and provides it via hyper+rustls?

gilescope:
Being able to use a pure rust ssl makes using a lot of rust tools a lot easier where finding openssl or building ssl from source is tricky. Building cargo, cargo-audit etc. It would make a big chunk of the ecosystem easier for people to use.

weihanglo:
Hi. I am wondering about how to integrate an **async** HTTP backend like hyper into git2. And it seems that spawning threads for hyper in the `Read`/`Write` trait is the most feasible solution which I can come up with. If there is other possible approach, please share it here. I am happy to do some experiments.

[...]

PR #624, opened 2020-09-25
Add git2_rustls transport

jsha:
This is one approach to solving #623. It takes a "minimum dependencies" approach and just relies on rustls, plus a tiny internal HTTP client. Advantages: Very few added dependencies. Disadvantages: No proxy support (yet). Naive handling of headers (e.g. case sensitive handling). Another possible approach would be to incorporate an external HTTP crate. I might try that in a separate PR using https://github.com/algesten/ureq/ for comparison. Curious what you think about the tradeoffs.

Updates examples/clone.rs to take an option "transport" argument, to use it with git2-curl or git2-rustls.

Relative to git2-curl, this does not have the problem that all data from a response has to be buffered into memory. Its read method reads directly off of a rustls stream.


alexcrichton:
Thanks for this! I think though that a solution like #625 is better where an HTTP client with TLS backed by rustls is better to do. Hard-coding a bunch of HTTP things here feels pretty wonky, and it seems like it'll eventually grow to have more HTTP features over time where it's probably best to just use a HTTP client crate.

jsha:
Yep, makes sense to me!

PR #625, opened 2020-09-25
Add git2_ureq transport

jsha:
This is an alternate approach to #624 that uses an actual HTTP client (ureq) on top of rustls. In the end this came out significantly cleaner, and by using default_features = false I was able to keep the dependencies very minimal.

Note that this branch is built on top of the branch for #624 for convenience. If we decide to go this route I'll excise the git2_rustls files.


jsha:
Also note this actually won't compile with ureq 1.4.1. I needed to make some local changes to add Send + Sync to the return value of into_reader, which I'll upstream now.

alexcrichton:
Nice! This seems pretty reasonable to me to add! I  haven't looked at git2-curl in a very long time though so some things could probably be improved there, but this seems like a great start at least!

jsha:
Updated to refer to a released version of ureq with the needed Send marker. I also refactored significantly to better reflect my understanding of how things work. Definitely interested in feedback on code structure and style.

jsha:
I'm actually going to close this for now. When I went deeper on integrating this with cargo, I realized the other fetches in cargo (the crate downloads) rely on HTTP/2 for a significant speedup. Since `ureq` doesn't currently support HTTP/2, it wouldn't be suitable as a general replacement for curl in cargo. I'm going to take a look at whether it's feasible to implement a lightweight subset of HTTP/2 in ureq for concurrent / pipelined fetches. But I may wind up coming back around to this with a different HTTP client.

PR #627, opened 2020-10-13
Update to libgit2 v1.1 and use `main` branch for tests.

de-vri-es:
This PR updates the `libgit2` submodule to the `v1.1.0` tag. It also updates the unit tests to use a fixed name for the HEAD of created test repositories.

Since libgit2 1.1.0, the gitconfig of the environment can change the name of the default branch. If the test do not specify a fixed name for the initial HEAD, they will fail on some systems with different configuration.

To make sure I didn't miss anything, I ran the tests this with `init.defaultBranch` in my gitconfig set to a value different from both `master` and `main`.

As a bonus, this PR also replaces `master` with `main` in documentation and examples of the `git2` crate.

This should automatically address rust-lang/cargo/issues/8588 once cargo updates the dependency on this crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ version of Rust known to pass tests.
 
 ## Version of libgit2
 
-Currently this library requires libgit2 1.0.0. The source for libgit2 is
+Currently this library requires libgit2 1.1.0. The source for libgit2 is
 included in the libgit2-sys crate so there's no need to pre-install the libgit2
 library, the libgit2-sys crate will figure that and/or build that for you.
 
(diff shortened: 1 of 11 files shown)


ehuss:
Thanks!  I think the version will also need to be bumped (similar to https://github.com/rust-lang/git2-rs/commit/218a2b380a6eae17ec8a7499683a5be2232cb659).

alexcrichton:
I'm gonna go ahead and merge this and I'll do the version bumps as a follow-up commit for a release.

PR #635, opened 2020-11-22
Expose git_merge_file function from libgit2 to repo.rs of git2-rs

kvzn:


libgit2-sys/lib.rs
```diff
@@ -3854,6 +3855,82 @@ extern "C" {
...
+    pub mode: c_uint,
+
+    /// The contents of the merge.
+    pub ptr: *const c_char,
+
+    /// The length of the merge contents.
+    pub len: size_t,
+}
+
+extern "C" {
```
alexcrichton, line 3917: Could this get merged (hah!) with the general block of function definitions elsewhere in this file?

src/lib.rs
```diff
@@ -1049,6 +1051,34 @@ pub enum FileMode {
...
+impl From<u32> for FileMode {
+    fn from(mode: u32) -> Self {
+        match mode {
+            raw::GIT_FILEMODE_UNREADABLE => FileMode::Unreadable,
+            raw::GIT_FILEMODE_TREE => FileMode::Tree,
+            raw::GIT_FILEMODE_BLOB => FileMode::Blob,
+            raw::GIT_FILEMODE_BLOB_EXECUTABLE => FileMode::BlobExecutable,
+            raw::GIT_FILEMODE_LINK => FileMode::Link,
+            raw::GIT_FILEMODE_COMMIT => FileMode::Commit,
+            mode => panic!("unknown file mode: {}", mode),
```
alexcrichton, line 1063: Since this is a fallible conversion I think it would either be best to avoid the `From` trait implementation or move this conversion to a helper function perhaps?

src/lib.rs
```diff
@@ -1049,6 +1051,34 @@ pub enum FileMode {
...
+        }
+    }
+}
+
+impl Into<u32> for FileMode {
```
alexcrichton, line 1068: FWIW this is typically phrase as `From<FileMode> for u32`, but if `FileMode` is an `enum` I think it would be better to set the discriminants to the `raw::*` values and that way `enum_value as u32` should work for this purpose.

src/merge.rs
```diff
@@ -192,3 +194,295 @@ impl<'repo> Drop for AnnotatedCommit<'repo> {
...
+    /// The mode that the resultant merge file should use.
+    pub mode: FileMode,
+
+    /// The contents of the merge.
+    pub content: Option<Vec<u8>>,
+}
+
+impl MergeFileResult {
+    /// Create MergeFileResult from C
+    pub fn from_raw(raw: raw::git_merge_file_result) -> MergeFileResult {
```
alexcrichton, line 445: I think this'll want to be `unsafe`?

src/repo.rs
```diff
@@ -1840,6 +1841,80 @@ impl Repository {
...
+                ancestor_raw,
+                ours_raw,
+                theirs_raw,
+                options.map(|o| o.raw())
+            ));
+
+            let result = MergeFileResult::from_raw(ret);
+
+            // FIXME: need to free????
+            raw::git_merge_file_result_free(&mut ret as *mut _);
```
alexcrichton, line 1912: Typically this is modeled by `MergeFileResult` taking ownership of the `git_merge_file_result` and is then responsible for any deallocation necessary, would that be possible here?

alexcrichton (commented):
Thanks for this!

src/merge.rs
someniceone: I have used `unsafe` at the specified lines, is that ok?
```rust
impl MergeFileResult {
    /// Create MergeFileResult from C
    pub fn from_raw(raw: raw::git_merge_file_result) -> MergeFileResult {
        let c_str: &CStr = unsafe { CStr::from_ptr(raw.path) };  // unsafe
        let str_slice: &str = c_str.to_str().unwrap_or("unknown path");
        let path: String = str_slice.to_owned();

        let content =
            unsafe { slice::from_raw_parts(raw.ptr as *const u8, raw.len as usize).to_vec() }; // unsafe

        MergeFileResult {
            automergeable: raw.automergeable > 0,
            path: Some(path),
            mode: raw.mode.into(),
            content: Some(content),
        }
    }
}
```

src/repo.rs
someniceone: So `raw::git_merge_file_result_free(&mut ret as *mut _);` is not needed?

src/lib.rs
someniceone: Is this way ok? The `From<>` `trait` is removed
```rust
impl FileMode {
    fn from(mode: u32) -> Self {
        match mode {
            raw::GIT_FILEMODE_UNREADABLE => FileMode::Unreadable,
            raw::GIT_FILEMODE_TREE => FileMode::Tree,
            raw::GIT_FILEMODE_BLOB => FileMode::Blob,
            raw::GIT_FILEMODE_BLOB_EXECUTABLE => FileMode::BlobExecutable,
            raw::GIT_FILEMODE_LINK => FileMode::Link,
            raw::GIT_FILEMODE_COMMIT => FileMode::Commit,
            mode => panic!("unknown file mode: {}", mode),
        }
    }
}
```

src/lib.rs
someniceone: So we don't need `Into<u32>` trait implementation but just use `file_mode as u32` when needed?

libgit2-sys/lib.rs
someniceone: I was worrying to make it mess at the beginning so they were put together, now they've been moved to the right places.

someniceone:
@alexcrichton BTW, could you point out how I can fix the issue in `CI / Test (windows)`? Tried but have no idea.

src/lib.rs
alexcrichton: Yeah so long as this is private this should be ok 👍

src/merge.rs
```diff
@@ -192,3 +194,295 @@ impl<'repo> Drop for AnnotatedCommit<'repo> {
...
+
+    /// Get the raw struct in C
+    pub fn raw(&self) -> *const raw::git_merge_file_input {
+        &self.raw as *const _
+    }
+}
+
+impl MergeFileInput {
+    /// Create from Repository and IndexEntry
+    pub fn from(repo: &Repository, index_entry: &IndexEntry) -> MergeFileInput {
```
alexcrichton, line 382: FWIW I've personally shied away from helper functions like this in the past for this crate. This crate tries to stick pretty close to libgit2's own API without adding too many Rust helpers and such. Is there a particular reason though that we should have this when libgit2 doesn't?

src/merge.rs
alexcrichton: Ah the `unsafe` part is that this is taking a raw C structure where the pointers aren't guaranteeed to be valid, so the `unsafe` signals "the caller needs to take care of that validity"

src/merge.rs
```diff
@@ -192,3 +194,295 @@ impl<'repo> Drop for AnnotatedCommit<'repo> {
...
+
+    /// The path that the resultant merge file should use, or NULL if a
+    /// filename conflict would occur.
+    pub path: Option<String>,
+
+    /// The mode that the resultant merge file should use.
+    pub mode: FileMode,
+
+    /// The contents of the merge.
+    pub content: Option<Vec<u8>>,
```
alexcrichton, line 440: FWIW I think this would probably be best done by (privately) containing `raw::git_merge_file_result` and providing accessors for each field. That way Rust doesn't need to allocate anything unnecessarily.

src/merge.rs
```diff
@@ -192,3 +194,295 @@ impl<'repo> Drop for AnnotatedCommit<'repo> {
...
+
+        if let Some(mode) = self.mode {
+            self.raw.mode = mode as u32;
+        }
+
+        self
+    }
+
+    /// File content, text or binary
+    pub fn content(&mut self, content: Option<Vec<u8>>) -> &mut MergeFileInput {
```
alexcrichton, line 361: I think it might actually be best to take `&[u8]` here and tie the lifetime to `MergeFileInput` so that way users won't have to copy data around.

src/merge.rs
```diff
@@ -192,3 +194,295 @@ impl<'repo> Drop for AnnotatedCommit<'repo> {
...
+    /// File name of the conflicted file, or `NULL` to not merge the path.
+    ///
+    /// You can turn this value into a `std::ffi::CString` with
+    /// `CString::new(&entry.path[..]).unwrap()`. To turn a reference into a
+    /// `&std::path::Path`, see the `bytes2path()` function in the private,
+    /// internal `util` module in this crate’s source code.
+    path: Option<CString>,
+
+    /// File mode of the conflicted file, or `0` to not merge the mode.
+    pub mode: Option<FileMode>,
```
alexcrichton, line 308: These fields should be private so alterations are forced to go through public functions.

alexcrichton: Also I think it's ok to eschew storing `mode` here, we can just store it directly in the `raw` field.

src/repo.rs
```diff
@@ -1840,6 +1841,77 @@ impl Repository {
...
+            ancestor_raw = ptr::null();
+        }
+        if let Some(ours) = ours {
+            ours_input = MergeFileInput::from(&self, ours);
+            ours_raw = ours_input.raw();
+        } else {
+            ours_raw = ptr::null();
+        }
+        if let Some(theirs) = theirs {
+            theirs_input = MergeFileInput::from(&self, theirs);
```
alexcrichton, line 1886: FWIW I think it would be best to take `theirs: Option<&MergeFileInput>` in this function instead of doing this computation internally.

alexcrichton (commented):
Oh the windows error I think has to do with the fact that enums in C have a different signededness than other platforms. I think you can probably fix things with some casts?

src/merge.rs
someniceone: Understood, than you!

src/merge.rs
someniceone: Understood, it's close now:
```rust
impl MergeFileInput {
    /// Create MergeFileInput
    pub fn from(path: &str, mode: FileMode, content: Vec<u8>) -> MergeFileInput {
        let mut input = MergeFileInput::new();
        input.path(path);
        input.mode(Some(mode));
        input.content(Some(content));
        input
    }
}
```

src/merge.rs
someniceone: I have made it using `lifetime`, but the `merge_file()` in `repo.rs`(line 1868) becomes ugly, could you tell me how to improve it?

src/merge.rs
someniceone: Yes, fixed that

src/repo.rs
someniceone: Understood, now it accepts `Option<&MergeFileInput>`

someniceone:
> Oh the windows error I think has to do with the fact that enums in C have a different signededness than other platforms. I think you can probably fix things with some casts?

I tried this, but doesn't work, any idea:
```rust
trait FileModeConvert {
    type ValueType;
    fn from(mode: Self::ValueType) -> Self;
}

impl FileModeConvert for FileMode {
    #[cfg(target_env = "msvc")]
    type ValueType = i32;

    #[cfg(not(target_env = "msvc"))]
    type ValueType = u32;

    fn from(mode: Self::ValueType) -> Self {
        match mode {
            raw::GIT_FILEMODE_UNREADABLE => FileMode::Unreadable,
            raw::GIT_FILEMODE_TREE => FileMode::Tree,
            raw::GIT_FILEMODE_BLOB => FileMode::Blob,
            raw::GIT_FILEMODE_BLOB_EXECUTABLE => FileMode::BlobExecutable,
            raw::GIT_FILEMODE_LINK => FileMode::Link,
            raw::GIT_FILEMODE_COMMIT => FileMode::Commit,
            mode => panic!("unknown file mode: {}", mode),
        }
    }
}
```

src/merge.rs
someniceone: Great, improved.

kvzn:
Can we merge this PR now? Thank you!

PR #640, opened 2020-12-11
Add rerun-if-changed to libgit2-sys build script.

ehuss:
pkg-config 0.3.18 changed the default of `env_metadata` to `true` (see https://github.com/rust-lang/pkg-config-rs/pull/105), which means it is now emitting `rerun-if-env-changed` directives, which means the default Cargo behavior of scanning the entire directory is no longer working. This means that changing anything in libgit2 will not trigger a build.

This fixes it by adding rerun-if-changed directives for libgit2.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -182,6 +182,10 @@ fn main() {
         println!("cargo:rustc-link-lib=framework=Security");
         println!("cargo:rustc-link-lib=framework=CoreFoundation");
     }
+
+    rerun_if(Path::new("libgit2/include"));
+    rerun_if(Path::new("libgit2/src"));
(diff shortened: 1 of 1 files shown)


ehuss:
@alexcrichton I've always wondered why if you pass a directory to `cargo:rerun-if-changed`, why it doesn't interpret that as "anything within this directory"?  Would there be any reason not to change that behavior?  It would allow avoiding needing the recursive function which I added here.

alexcrichton:
We debated it a bit when it was first implemented in Cargo but figured it was conservative not to do it. I don't think there's any reasonable portable interpretation though other than "everything in here", so I think it would be fine to implement nowadays in Cargo.

Thanks for taking the time to do this!

Issue #461, opened 2019-08-27
Support creating detached remotes


keirlawson:
It would be great to be able to create detached remotes with git2-rs, to allow for doing the equivalent of a `git ls-remote` when outside of a git repository.  It looks like this is supported in libgit2 via `git_remote_create_detached`.

PR #641, opened 2020-12-16
Allow creating detached remotes

Alovchin91:
This PR introduces an associated function on `Remote` type to allow creating detached remotes.

With this change, it's simple to implement `git ls-remote` doing something like this:

```rust
fn ls_remote(url: &str) -> Result<(), git2::Error> {
    let mut remote = Remote::create_detached(url)?;

    remote.connect(Direction::Fetch)?;

    for remote_ref in remote.list()? {
        println!("{}", remote_ref.name());
    }

    remote.disconnect()?;

    Ok(())
}
```

This also addresses #461.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2030,10 +2030,7 @@ extern "C" {
         repo: *mut git_repository,
         url: *const c_char,
     ) -> c_int;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks!

PR #643, opened 2020-12-16
Sync config docs

ehuss:
Sync the config documentation with the docs from libgit2.  Primarily, I wanted to emphasize that `Config::get_str` can return an error for a non-snapshot config.


alexcrichton:
👍

PR #648, opened 2020-12-21
Add TreeUpdateBuilder to support git_tree_create_updated

joshtriplett:
git_tree_create_updated requires a length and a contiguous array of
git_tree_update structures, each of which has a pointer to a C string.
Create a builder, `TreeUpdateBuilder`, to wrap that safely without
excessive copies.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -7,8 +7,8 @@ use std::path::Path;
 use std::ptr;
 
 use crate::util::{self, Binding};
-use crate::{panic, raw, Error, FetchOptions, IntoCString, Repository};
-use crate::{CheckoutNotificationType, DiffFile, Remote};
+use crate::{panic, raw, Error, FetchOptions, IntoCString, Oid, Repository, Tree};
+use crate::{CheckoutNotificationType, DiffFile, FileMode, Remote};
 
 /// A builder struct which is used to build configuration for cloning a new git
 /// repository.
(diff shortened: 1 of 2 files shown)


src/build.rs
```diff
@@ -674,9 +680,82 @@ extern "C" fn notify_cb(
...
+
+    /// Add an update setting the specified `path` to a specific Oid, whether it currently exists
+    /// or not.
+    ///
+    /// Note that libgit2 does not support an upsert of a previously removed path, or an upsert
+    /// that changes the type of an object (such as from tree to blob or vice versa).
+    ///
+    /// The mode given must be one of 0o040000, 0o100644, 0o100755, 0o120000 or
+    /// 0o160000 currently.
+    pub fn upsert<T: IntoCString>(&mut self, path: T, id: Oid, filemode: i32) -> &mut Self {
```
ehuss, line 722: The use of i32 here seems a little strange to me.  It seems consistent with the rest of this crate which uses i32 for filemode, but that seems like a strange choice considering *most* unix platforms use u32.  

Since the value is constrained to the git_filemode_t enum, wouldn't it be better to pass in a `FileMode`?  It seems like that would be safer and a clearer.

joshtriplett: I was going for consistency, but I can very easily change this to `FileMode`.

joshtriplett:
@ehuss Fixed.

ehuss (approved):

PR #650, opened 2020-12-22
Exclude more files from the published package.

ehuss:
This cuts about 3.5MB from the uncompressed package (11MB down to 7.5).


joshtriplett:
Nice improvement, thank you.

PR #651, opened 2020-12-23
test for the Repository::config() function

alexanderkjall:
I was debugging a thing, and thought that I could just as well push the test I wrote for debugging upstream.

This test verifies that the Repository::config() function returns a config object that have the local repository as highest prio.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -3624,16 +3624,13 @@ mod tests {
 
     #[test]
     fn smoke_config_write_and_read() {
-        let td = TempDir::new().unwrap();
-        let path = td.path();
-
-        let repo = Repository::init(path).unwrap();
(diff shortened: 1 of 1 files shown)


src/repo.rs
```diff
@@ -3621,4 +3621,25 @@ mod tests {
         // reverting twice restores `foo` file
         assert!(foo_file.exists());
     }
+
+    #[test]
+    fn smoke_config_write_and_read() {
+        let td = TempDir::new().unwrap();
```
joshtriplett, line 3627: There's a helper function used in other tests that creates a temporary directory and repository for you, in one call.

alexanderkjall: Thanks for the quick feedback, I have updated the PR to use that.

alexcrichton:
Thanks!

PR #655, opened 2020-12-30
Add bindings for git_merge_analysis_for_ref

kamalmarhubi:
This is a slightly generalised version of git_merge_analysis that allows
specifying the ref to analyse against instead of using the repository's
current HEAD.


alexcrichton:
Thanks! I think this needs a rebase now though?

kamalmarhubi:
Oops there was also an `#[allow(warnings)]` in there that should have been removed before I pushed. Fixed it up and rebased.

alexcrichton:
👍

PR #656, opened 2021-01-01
Add git_libgit2_opts binding for get/set search path

weihanglo:
Prepare these bindings in order to fix https://github.com/rust-lang/cargo/issues/8863.

Due to the same reason mentioned https://github.com/rust-lang/git2-rs/pull/126#discussion_r57835890 , the bindings do not split strings to paths and leave the work for callers to call `std::env::join_paths` themselves.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
 libgit2-sys = { path = "libgit2-sys", version = "0.12.17" }
+once_cell = "1.5"
 
 [target."cfg(all(unix, not(target_os = \"macos\")))".dependencies]
 openssl-sys = { version = "0.9.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
 libgit2-sys = { path = "libgit2-sys", version = "0.12.17" }
-once_cell = "1.5"
 
 [target."cfg(all(unix, not(target_os = \"macos\")))".dependencies]
 openssl-sys = { version = "0.9.0", optional = true }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -363,7 +363,7 @@ pub enum BranchType {
 #[derive(PartialEq, Eq, Debug, Copy, Clone)]
 pub enum ConfigLevel {
     /// System-wide on Windows, for compatibility with portable git
-    ProgramData,
+    ProgramData = 1,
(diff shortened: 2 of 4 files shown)


weihanglo:
Not sure how to handle path on Windows. Should it returns `OsString` directly? Or just throw an invalid UTF-8 error like `util::byte2path`?

https://github.com/rust-lang/git2-rs/blob/68805637017da2fb59bc7dc600e89614db016c32/src/util.rs#L91-L95

weihanglo:
Also, 0075b12 makes it thread-safe to access global state of search path, but have no idea is that necessary.

weihanglo:
> Not sure how to handle path on Windows. Should `get_search_path` returns `CString` directly? Or just throw an invalid UTF-8 error like `util::byte2path`?
> 

e5d8643 _fix: get_search_path will returns CString_

I think returning a `CString` is more straight foward since `set_search_search` also accept an `IntoCString` type.

alexcrichton:
This seems generally fine but the global-state-ness is worrisome with respect to how this is being used in https://github.com/rust-lang/cargo/pull/9035. If this doesn't work for that PR would you still like to land this?

weihanglo:
> This seems generally fine but the global-state-ness is worrisome with respect to how this is being used in [rust-lang/cargo#9035](https://github.com/rust-lang/cargo/pull/9035). If this doesn't work for that PR would you still like to land this?

If i didn't get it wrong, from the view of git2-rs itself, the global state in git2-rs just reflects the global state in libgit2. To make git2-rs as safe as possible ([that's one of it's goal](https://github.com/rust-lang/git2-rs/blob/c7492dfbba9d38a2b8897ee8ca81ba73b05685b3/src/lib.rs#L9-L11) 😂), I would choose an approach like this one no matter how callers use it.

alexcrichton:
Ok, in that case though I think it may be better to flag these functions as `unsafe`. They're not safe to mix with other configuration functions in multiple threads and this probably has to happen near the front of libgit2's usage rather than later. I think the synchronization can only really be left up to the caller since we don't know about all the calls to libgit2's configuration functions.

weihanglo:
Got it. I'll remove the synchronization part. Thanks!

alexcrichton:
Looks good! As one final thing though, I think the test may wantt to move to an integration test in something like `tests/foo.rs`? Otherwise I don't think we can guarantee that it's synchronized with all other tests.

weihanglo:
Thanks for your advice. The test has been moved to a integration test at adc0ee8.  

I have a dumb question to task. If I understand it correctly, the reason moving it to a integration test is that if people what to mutate the same global state, it can edit the "integration test" directly without touching "unit test" part, right?

alexcrichton:
👍

PR #661, opened 2021-01-16
fix: call init() before setting global state

weihanglo:
Must inititialize global state for libgit2 before setting.


alexcrichton:
👍

PR #663, opened 2021-01-20
Add useHttpPath support in CredentialHelper

arlosi:
Respect the `useHttpPath` git configuration parameter when calling the git credential helper.

### DIFF
--- a/src/cred.rs
+++ b/src/cred.rs
@@ -22,6 +22,7 @@ pub struct CredentialHelper {
     pub username: Option<String>,
     protocol: Option<String>,
     host: Option<String>,
+    path: Option<String>,
     url: String,
     commands: Vec<String>,
 }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Looks good to me, thanks!

sharkdp:
This changeset uses [`str::strip_prefix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) which is only available in Rust 1.45. The `git2` crate previously built fine with 1.42.

I realize that you don't have a strict policy on this (*"git2-rs works with stable Rust, and typically works with the most recent prior stable release as well"*), but I thought it would be worth mentioning. Bumping the min. supported Rust version in a patch release like this possibly breaks downstream builds that simply use
``` toml
git2 = "0.13"
```

The README also says *"Check the MSRV job of the CI script to see the oldest version of Rust known to pass tests."*, but I don't see an explicit requirement in https://github.com/rust-lang/git2-rs/blob/455098a8a1ce24305faf690264fc5b636d210739/.github/workflows/main.yml

Also, thank you for creating and maintaining this library - it has been a pleasure to work with!

PR #665, opened 2021-01-24
DiffOptions fixes - Fix DiffOptions::force_binary to pass the correct flag

joshtriplett:
force_binary was passing GIT_DIFF_FORCE_TEXT (like force_text); change
it to pass GIT_DIFF_FORCE_BINARY.

Also fix some documentation issues.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -685,7 +685,7 @@ impl DiffOptions {
         self.flag(raw::GIT_DIFF_INCLUDE_UNTRACKED, include)
     }
 
-    /// Flag indicating whether untracked directories are deeply traversed or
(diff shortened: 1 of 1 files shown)


src/diff.rs
```diff
@@ -685,7 +685,7 @@ impl DiffOptions {
         self.flag(raw::GIT_DIFF_INCLUDE_UNTRACKED, include)
     }
 
-    /// Flag indicating whether untracked directories are deeply traversed or
+    /// Flag indicating whether untracked directories are traversed deeply or
```
joshtriplett, line 688: This change matches the comment for `recurse_ignored_dirs`.

Issue #662, opened 2021-01-20
Expose git_revwalk_add_hide_cb


initcrash:
Hi,
I have a use case that would greatly benefit from using that functionality.
Is it already planed (work in progress?) to add this or should I look into adding it myself?

Thanks :)

initcrash:
fixed

PR #667, opened 2021-01-26
Expose git_revwalk_hide_cb

initcrash:


src/revwalk.rs
```diff
@@ -10,6 +10,19 @@ use crate::{raw, Error, Oid, Repository, Sort};
 pub struct Revwalk<'repo> {
     raw: *mut raw::git_revwalk,
     _marker: marker::PhantomData<&'repo Repository>,
+    hide_cb: Option<Box<dyn FnMut(Oid) -> bool>>,
+}
+
+extern "C" fn revwalk_hide_cb(commit_id: *const raw::git_oid, payload: *mut c_void) -> c_int {
+    unsafe {
```
alexcrichton, line 17: This'll want to get wrapped in `panic::call` to avoid aborting the process on panics

src/revwalk.rs
```diff
@@ -119,6 +133,20 @@ impl<'repo> Revwalk<'repo> {
         Ok(())
     }
 
+    /// Hide all commits for which the callback returns true from
+    /// the walk.
+    pub fn hide_callback(&mut self, callback: Box<dyn FnMut(Oid) -> bool>) -> Result<(), Error> {
```
alexcrichton, line 138: Ideally this would take `impl FnMut(...)` rather than a `Box` (but we'd probably box it up intternally). Additionally this would ideally not have a `'static` bound similar to other callbacks in these bindings. Adding more lifetime arguments would be a breaking change to `Revwalk`, but we could at least just bound the callback by `'repo` I believe.

src/revwalk.rs
```diff
@@ -21,6 +34,7 @@ impl<'repo> Revwalk<'repo> {
         unsafe {
             try_call!(raw::git_revwalk_reset(self.raw()));
         }
+        self.hide_cb = None;
```
alexcrichton, line 37: To confirm, `git_revwalk_reset` clears the callback internally in libgit2, right? So this is just reflecting how the Rust callback will never get called anyway?

src/revwalk.rs
alexcrichton: Additionally `panic::check` will need to be called after C functions which may trigger the initial call of the callback

alexcrichton (commented):
Thanks for this!

src/revwalk.rs
initcrash: Ok, I saw this in other places but did not really get it :)

src/revwalk.rs
initcrash: I tried to do do it that way (boxing internally) but could not get it to compile. It suggests me to add `'static` bound which is not desirable. Not sure `'repo` is right either because the lifetime of both the `Revwalk` as well as the callback can be much shorter than `'repo`.

src/revwalk.rs
initcrash: Yes that's right. I'll add a comment.

src/revwalk.rs
initcrash: I tried: 
```suggestion
        pub fn hide_callback<C: 'repo>(&mut self, callback: C) -> Result<(), Error> 
    where
        C: FnMut(Oid) -> bool,

```
But that does not compile :(

alexcrichton: Hm ok, well in any case relaxing the lifetime bound here will require adjusting the parameters on `Revwalk`. For now it's fine to just require `'static` and we can fix this in a later breaking change if necessary. Can this still switch to the generic version with a `'static` bound though rather than taking a trait object here?

src/revwalk.rs
```diff
@@ -186,7 +217,11 @@ impl<'repo> Iterator for Revwalk<'repo> {
         };
         unsafe {
             try_call_iter!(raw::git_revwalk_next(&mut out, self.raw()));
-            Some(Ok(Binding::from_raw(&out as *const _)))
+            Some(if crate::panic::panicked() {
```
alexcrichton, line 220: I think here this will want to propagate the panic rather than convert that to an error

src/revwalk.rs
initcrash: So I came up with a different way to introduce this API that is quite I bit better I believe. (See new commit)
Comment and docs will still need to be updated if you agree with the approach taken.

extrawurst:
related to #662

PR #669, opened 2021-02-09
Expose the git_commit_header_field function.

Databean:
This accesses individual header fields in the object database representation of commit objects. These header fields can be seen by running the following command in a git repository:

    git cat-file -p $(git rev-parse HEAD)


alexcrichton:
Thanks!

PR #671, opened 2021-02-20
Add bindings for `git_attr_value()`

### LINKED ISSUE #670: No `git_attr_value()` binding
Currently the value of a git attribute can be obtained via [`Repository::get_attr()`](https://docs.rs/git2/0.13.17/git2/struct.Repository.html#method.get_attr) or [`Repository::get_attr_bytes()`](https://docs.rs/git2/0.13.17/git2/struct.Repository.html#method.get_attr_bytes), but they return `Option<&str>` or `Option<&[u8]>`, which do not tell the attribute is set to [true](https://docs.rs/libgit2-sys/0.12.18+1.1.0/libgit2_sys/fn.git_attr_value.html?search=GIT_ATTR_VALUE_TRUE), [false](https://docs.rs/libgit2-sys/0.12.18+1.1.0/libgit2_sys/constant.GIT_ATTR_VALUE_FALSE.html), [a string](https://docs.rs/libgit2-sys/0.12.18+1.1.0/libgit2_sys/constant.GIT_ATTR_VALUE_STRING.html) or [unspecified](https://docs.rs/libgit2-sys/0.12.18+1.1.0/libgit2_sys/constant.GIT_ATTR_VALUE_UNSPECIFIED.html).

`git_attr_get()` returns some [special *pointers*](https://github.com/libgit2/libgit2/blob/27e34f9b9843f7bcc33a4ccfe3e395fe303cba63/src/attr.c#L18-L34) (note that the bindings return an optional slice, but fortunately that preserves the address) to differentiate those situations, and [`git_attr_value()`](https://docs.rs/libgit2-sys/0.12.18+1.1.0/libgit2_sys/fn.git_attr_value.html) is used to obtain that information, but there's no binding to that function (https://github.com/rust-lang/git2-rs/pull/389#discussion_r266569002 looks relevant, but unfortunately it turns out that the current bindings lose informations, and the they [do not check whether the value is really a string](https://github.com/rust-lang/git2-rs/blob/455098a8a1ce24305faf690264fc5b636d210739/src/repo.rs#L984) anyway).

This is what I currently do to extract that information (feel free to copy-and-paste if it helps):
<details>

```rust
#[cfg_attr(test, derive(Debug, PartialEq))]
enum AttrValue<'a> {
    True,
    False,
    String(&'a [u8]),
    Unspecified,
}

fn attr_value(value: Option<&[u8]>) -> AttrValue<'_> {
    match unsafe { git_attr_value(value.map_or(ptr::null(), |v| v.as_ptr().cast())) } {
        GIT_ATTR_VALUE_TRUE => AttrValue::True,
        GIT_ATTR_VALUE_FALSE => AttrValue::False,
        GIT_ATTR_VALUE_STRING => AttrValue::String(value.unwrap()),
        GIT_ATTR_VALUE_UNSPECIFIED => AttrValue::Unspecified,
        _ => unreachable!(),
    }
}

#[test]
fn test_attr_value() {
    extern "C" {
        static mut git_attr__true: *const c_char;
        static mut git_attr__false: *const c_char;
        static mut git_attr__unset: *const c_char;
    }
    let attr_true = unsafe { CStr::from_ptr(git_attr__true) };
    let attr_false = unsafe { CStr::from_ptr(git_attr__false) };
    let attr_unset = unsafe { CStr::from_ptr(git_attr__unset) };
    assert_eq!(
        super::attr_value(attr_true.to_owned().as_bytes().into()),
        AttrValue::String(attr_true.to_bytes())
    );
    assert_eq!(
        super::attr_value(attr_false.to_owned().as_bytes().into()),
        AttrValue::String(attr_false.to_bytes())
    );
    assert_eq!(
        super::attr_value(attr_unset.to_owned().as_bytes().into()),
        AttrValue::String(attr_unset.to_bytes())
    );
    assert_eq!(super::attr_value(Some(b"foo")), AttrValue::String(b"foo"));
    assert_eq!(super::attr_value(attr_true.to_bytes().into()), AttrValue::True);
    assert_eq!(super::attr_value(attr_false.to_bytes().into()), AttrValue::False);
    assert_eq!(super::attr_value(attr_unset.to_bytes().into()), AttrValue::Unspecified);
    assert_eq!(super::attr_value(None), AttrValue::Unspecified);
}
```
</details>

I'm also willing to submit a pull request.

hyd-dev:
#### New public APIs
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum AttrValue<'string> {
    True,
    False,
    String(&'string str),
    Bytes(&'string [u8]),
    Unspecified,
}

impl<'string> AttrValue<'string> {
    pub fn from_string(_: Option<&'string str>) -> Self;
    pub fn from_bytes(_: Option<&'string [u8]>) -> Self;
}
```

#### Summary
- Adds some new public APIs as described above.
- States that the return types of `Repository::get_attr` and `Repository::get_attr_bytes` will probably be changed in the next major version in their documentations.

Wrapping the return value of `Repository::get_attr(_bytes)` in a new type is probably better, but that's a breaking change, so it's probably not desirable unless it's ready to bump to 0.14.

Resolves #670.

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -1,0 +1,174 @@
+use crate::raw;
+use std::convert::TryFrom;
+use std::ptr;
+use std::str::{self, Utf8Error};
+
+macro_rules! define_inspector {
+    ($enum_name:ident, $enum_doc:literal, $fn_name:ident, $fn_doc:literal, $s:ty) => {
+        #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+        #[doc = $enum_doc]
+        pub enum $enum_name<'string> {
+            /// The attribute is set to true.
+            True,
+            /// The attribute is unset (set to false).
+            False,
+            /// The attribute is set to a string.
+            String(&'string $s),
+            /// The attribute is not specified.
+            Unspecified,
+        }
+
+        #[doc = $fn_doc]
+        pub fn $fn_name(attr_value: Option<&$s>) -> $enum_name<'_> {
+            match unsafe {
+                raw::git_attr_value(attr_value.map_or(ptr::null(), |v| v.as_ptr().cast()))
+            } {
+                raw::GIT_ATTR_VALUE_TRUE => $enum_name::True,
+                raw::GIT_ATTR_VALUE_FALSE => $enum_name::False,
+                raw::GIT_ATTR_VALUE_STRING => $enum_name::String(attr_value.unwrap()),
+                raw::GIT_ATTR_VALUE_UNSPECIFIED => $enum_name::Unspecified,
+                _ => unreachable!(),
+            }
+        }
+    };
+}
+
+define_inspector!(
+    AttrValue,
+    "All possible states of an attribute, using [`prim@str`] to represent the string.\n\n\
+     This enum is returned by [`attr_value`].",
+    attr_value,
+    "Returns the state of an attribute by inspecting its [value](crate::Repository::get_attr) \
+     by a [string](prim@str).",
+    str
+);
+
+/// Converts [`AttrValueBytes`] to [`AttrValue`]. If the attribute is [set to a string](`AttrValueBytes::String`),
+/// this implementation will use [`str::from_utf8`] to perform the conversion.
+impl<'string> TryFrom<AttrValueBytes<'string>> for AttrValue<'string> {
+    type Error = Utf8Error;
+
+    fn try_from(value: AttrValueBytes<'string>) -> Result<Self, Self::Error> {
+        match value {
+            AttrValueBytes::True => Ok(Self::True),
+            AttrValueBytes::False => Ok(Self::False),
+            AttrValueBytes::String(s) => Ok(Self::String(str::from_utf8(s)?)),
+            AttrValueBytes::Unspecified => Ok(Self::Unspecified),
+        }
+    }
+}
+
+define_inspector!(
+    AttrValueBytes,
+    "All possible states of an attribute, using a [byte](u8) [slice] to represent the string.\n\n\
+     This enum is returned by [`attr_value_bytes`].",
+    attr_value_bytes,
+    "Returns the state of an attribute by inspecting its [value](crate::Repository::get_attr_bytes) \
+     by a [byte](u8) [slice].",
+    [u8]
+);
+
+/// Converts [`AttrValue`] to [`AttrValueBytes`]. This implementation will convert the
+/// [string slice](prim@str) to a [byte](u8) [slice] when the attribute is
+/// [set to a string](`AttrValue::String`).
+impl<'string> From<AttrValue<'string>> for AttrValueBytes<'string> {
+    fn from(value: AttrValue<'string>) -> Self {
+        match value {
+            AttrValue::True => Self::True,
+            AttrValue::False => Self::False,
+            AttrValue::String(s) => Self::String(s.as_ref()),
+            AttrValue::Unspecified => Self::Unspecified,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{AttrValue, AttrValueBytes};
+    use std::convert::{TryFrom, TryInto};
+    use std::ffi::CStr;
+    use std::os::raw::c_char;
+
+    extern "C" {
+        // libgit2 defines them as mutable, so they are also declared mutable here.
+        // However, libgit2 never mutates them, thus it's always safe to access them in Rust.
+        static mut git_attr__true: *const c_char;
+        static mut git_attr__false: *const c_char;
+        static mut git_attr__unset: *const c_char;
+    }
+
+    macro_rules! define_test {
+        ($enum_name:ident, $fn_name:ident) => {
+            #[test]
+            fn $fn_name() {
+                let attr_true = unsafe { CStr::from_ptr(git_attr__true) }.to_str().unwrap();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1390,6 +1392,85 @@ impl Default for StashFlags {
...
+            /// The attribute is unset (set to false).
+            False,
+            /// The attribute is set to a string.
+            String(&'string $s),
+            /// The attribute is not specified.
+            Unspecified,
+        }
+
+        #[doc = $fn_doc]
+        pub fn $fn_name(attr_value: Option<&$s>) -> $enum_name<'_> {
```
alexcrichton, line 1411: Perhaps this should be an inherent method on `$enum_name`, for example `AttrValue::new(Some("bar"))`?

src/lib.rs
```diff
@@ -1390,6 +1392,85 @@ impl Default for StashFlags {
     }
 }
 
+macro_rules! define_attr_inspector {
```
alexcrichton, line 1395: Could this be located in a submodule for organization, perhaps `src/attr.rs`?

(still reexported from the root, however)

src/lib.rs
```diff
@@ -1390,6 +1392,85 @@ impl Default for StashFlags {
...
+     This enum is returned by [`attr_value`].",
+    attr_value,
+    "Returns the state of an attribute by inspecting its [value](Repository::get_attr) \
+     by a [string](prim@str).",
+    str
+);
+
+/// Converts [`AttrValueBytes`] to [`AttrValue`]. If the attribute is [set to a string](`AttrValueBytes::String`),
+/// this implementation will use [`str::from_utf8`] to perform the conversion.
+impl<'string> TryFrom<AttrValueBytes<'string>> for AttrValue<'string> {
```
alexcrichton, line 1437: Instead of having two enums, could there perhaps be one enum with a string/bytes variant where the bytes are guaranteed to always be invalid utf-8?

hyd-dev: > where the bytes are guaranteed to always be invalid utf-8?

Does that mean `AttrValue::new()` will always perform UTF-8 validation? Could I add another `new` that skips UTF-8 validation?

alexcrichton: Seems reasonable to me!

alexcrichton:
Thanks!

src/attr.rs
```diff
@@ -0,0 +1,165 @@
...
+#[cfg(test)]
+mod tests {
+    use super::AttrValue;
+    use std::ffi::CStr;
+    use std::os::raw::c_char;
+
+    extern "C" {
+        // libgit2 defines them as mutable, so they are also declared mutable here.
+        // However, libgit2 never mutates them, thus it's always safe to access them in Rust.
+        static mut git_attr__true: *const c_char;
```
kim, line 88: I'm getting linker errors for these since 65c0dccdf6640d2a832395f0c5d2db763f36628d when trying to `cargo test`:

```
  = note: ld.lld: error: undefined symbol: git_attr__true
          >>> referenced by attr.rs:135 (src/attr.rs:135)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_string::hba9c77ae386af5fe)
          >>> referenced by attr.rs:140 (src/attr.rs:140)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_bytes::h12449f8f751186bc)

          ld.lld: error: undefined symbol: git_attr__false
          >>> referenced by attr.rs:0 (src/attr.rs:0)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_string::hba9c77ae386af5fe)
          >>> referenced by attr.rs:0 (src/attr.rs:0)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_bytes::h12449f8f751186bc)

          ld.lld: error: undefined symbol: git_attr__unset
          >>> referenced by attr.rs:0 (src/attr.rs:0)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_string::hba9c77ae386af5fe)
          >>> referenced by attr.rs:0 (src/attr.rs:0)
          >>>               /tmp/git2-rs/target/debug/deps/git2-33fc7bd91ae7f3bc.tgtslt60c5cmi6e.rcgu.o:(git2::attr::tests::attr_value_from_bytes::h12449f8f751186bc)
          collect2: error: ld returned 1 exit status
```

What am I doing wrong?

hyd-dev: @kim I'm not sure... I am not able to reproduce with `RUSTFLAGS=-Clink-args=-fuse-ld=lld cargo test`.

hyd-dev: @kim Could you share what linker flags/`rustc` flags/Cargo profile settings you use and maybe a verbose build output? I'm sorry for the breakage and willing to fix the problem.

PR #673, opened 2021-02-24
Add an additional binding for `git_attr_value()` that skips UTF-8 validation

hyd-dev:
This PR adds an addition binding that skips UTF-8 validation as described in https://github.com/rust-lang/git2-rs/pull/671#discussion_r581157198 as `AttrValue::always_bytes()`:
```rust
impl<'string> AttrValue<'string> {
    pub fn always_bytes(value: Option<&'string [u8]>) -> Self;
}
```

### DIFF
--- a/src/attr.rs
+++ b/src/attr.rs
@@ -49,14 +49,13 @@ impl<'string> AttrValue<'string> {
     /// This function will perform UTF-8 validation when the attribute is set to a string, returns
     /// [`AttrValue::String`] if it's valid UTF-8 and [`AttrValue::Bytes`] otherwise.
     pub fn from_bytes(value: Option<&'string [u8]>) -> Self {
-        from_value!(value => {
-            let value = value.unwrap();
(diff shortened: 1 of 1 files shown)


src/attr.rs
```diff
@@ -15,51 +15,60 @@ pub enum AttrValue<'string> {
...
+            let value = value.unwrap();
+            if let Ok(string) = str::from_utf8(value) {
+                Self::String(string)
+            } else {
+                Self::Bytes(value)
             }
-            raw::GIT_ATTR_VALUE_UNSPECIFIED => Self::Unspecified,
-            _ => unreachable!(),
-        }
+        })
```
alexcrichton, line 59: Could this method perhaps call `always_bytes` instead of reusing a macro to call the same C function?

alexcrichton (commented):
Seems reasonable to me!

src/attr.rs
hyd-dev: Good idea! :+1:

PR #678, opened 2021-03-07
Implement git_diff_patchid

foriequal0:


libgit2-sys/lib.rs
```diff
@@ -3435,6 +3442,16 @@ extern "C" {
         version: c_uint,
     ) -> c_int;
 
+    pub fn git_diff_patchid(
+        out: *mut git_oid,
+        diff: *mut git_diff,
+        opts: *mut git_diff_patchid_options,
```
foriequal0, line 3448: `libgit2` doesn't seem to have a clear rule on arguments whether they should be `const *` or not. I would've liked to make `diff` and `opts` `*const`, but `libgit2` didn't made them 'const *' although they doesn't seem to be mutated.

src/diff.rs
```diff
@@ -278,6 +283,21 @@ impl<'repo> Diff<'repo> {
         Ok(buf)
     }
 
+    /// Create an patchid from a diff.
+    pub fn patchid(&self, opts: Option<&mut DiffPatchidOptions>) -> Result<Oid, Error> {
+        let mut raw = raw::git_oid {
+            id: [0; raw::GIT_OID_RAWSZ],
+        };
```
foriequal0, line 290: I prefer this
```rust
let mut oid = Oid::zeros();
unsafe { try_call!(raw::git_diff_patchid(oid.raw(), ...) };
Ok(oid)
```
but all oids are constructed in raw type and wrapped before the return throughout the code.

src/diff.rs
```diff
@@ -278,6 +283,21 @@ impl<'repo> Diff<'repo> {
         Ok(buf)
     }
 
+    /// Create an patchid from a diff.
+    pub fn patchid(&self, opts: Option<&mut DiffPatchidOptions>) -> Result<Oid, Error> {
```
foriequal0, line 287: Underlying `self.raw` is `*mut`, and `raw::git_diff_patchid` takes `*mut`.
`Diff::format_email` choose `&mut self` in similar situation, while `Diff::stats` choose `&self`.
I choose `&self` since it is semantically more natural (`patchid` doesn't seem to modify the diff).
I also wanted to make `opts` to be `Option<&DiffPatchidOptions>` by similar reason (who expects an option to be mutated?), but many of existing other `opts` are `Option<&mut ...>` and it doesn't seem to be easy since the underlying raw value is not a pointer.

foriequal0:
@alexcrichton Can I ask you a review for this PR? Also I've left some comments.

libgit2-sys/lib.rs
alexcrichton: We just match `libgit2`, so this should be fine.

src/diff.rs
alexcrichton: I think this is fine for now. It largely depends on whether callbacks are involved which they aren't here yet, and this can always be updated later if necessary.

alexcrichton:
Looks good to me, thanks for this!

PR #679, opened 2021-03-08
expose diffline origin value (as enum)

extrawurst:
so i do not need to hardcode char's but can use actual typed enum values.

Just one question: do I have to write a wrapper enum for these things instead of exposing the one from `raw`?

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -951,6 +951,61 @@ impl<'diff> DoubleEndedIterator for Deltas<'diff> {
 }
 impl<'diff> ExactSizeIterator for Deltas<'diff> {}
 
+/// Line origin constants.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum DiffLineType {
+    /// These values will be sent to `git_diff_line_cb` along with the line
+    Context,
+    ///
+    Addition,
+    ///
+    Deletion,
+    /// Both files have no LF at end
+    ContextEOFNL,
+    /// Old has no LF at end, new does
+    AddEOFNL,
+    /// Old has LF at end, new does not
+    DeleteEOFNL,
+    /// The following values will only be sent to a `git_diff_line_cb` when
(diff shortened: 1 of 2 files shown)


weihanglo:
I think we should not expose raw values from libgit2 to git2 consumer. `raw::git_diff_line_t` is just an `u32` not even a `char`, and exposing it won't gain supports from Rust's strong type system. 

Your enum wrapper idea sounds more suitable to me for this case. 😄

alexcrichton:
Yeah if possible I think it'd be good to have a wrapper type for the return value.

extrawurst:
ok updated and wrapped in rust idiomatic enum `DiffLineType`. Just one thing: how do you think about the acronyms? right now I left `ContextEOFNL` pretty much like in libgit2. Alternative would be `ContextEndOfFileNewLine` 🤔

src/diff.rs
```diff
@@ -931,6 +931,61 @@ impl<'diff> DoubleEndedIterator for Deltas<'diff> {
 }
 impl<'diff> ExactSizeIterator for Deltas<'diff> {}
 
+/// Line origin constants.
+#[derive(Copy, Clone, Debug)]
+pub enum DiffLineType {
+    /// These values will be sent to `git_diff_line_cb` along with the line
+    Context,
+    /// ditto
```
weihanglo, line 939: It would be great to add more context instead of **ditto**. 😀

image

extrawurst: the problem is, that the official libgit2 docs do not document these either. so better remove the `ditto`?

extrawurst: changed

alexcrichton:
Looks reasonable to me! Could you add a test exercising this as well?

extrawurst:
@alexcrichton good call, the unittest uncovered the lack of `pub use` the new type in `lib.rs`

alexcrichton:
Thanks!

PR #680, opened 2021-03-09
Parse credential.helper command line using shell-words

arlosi:
Since `sh` is not available by default on Windows, git2-rs currently splits the credential
helper string by whitespace, which attempts to execute an invalid command line in the case
of the default credential helper on Windows.

The default git credential helper on Windows (Git Credential Manager Core) is set up as
 `C:/Program\ Files\ \(x86\)/Git\ Credential\ Manager\ Core/git-credential-manager-core.exe`

This change resolves the issue by parsing the command line using `shell_words`,
which follows the unix-style parsing rules used by sh. `shell-words` is a small crate that
does not have any additional dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ url = "2.0"
 bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
+shell-words = "1.0"
 libgit2-sys = { path = "libgit2-sys", version = "0.12.18" }
 
 [target."cfg(all(unix, not(target_os = \"macos\")))".dependencies]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for this! I'm not sure if this is the right implementation for Windows, though. Do you know what `git` itself does in situations like this? Could/should we perhaps delegate to `cmd` instead of `sh` if it fails?

arlosi:
I think it's correct to always interpret it as a Unix shell command. `git` for Windows includes `sh`, and `git` always runs the credential helper command through `sh`.

Here's a screenshot from Process Explorer when the credential helper is prompting for credentials:
image

Running the command through `cmd` will fail, since the backslashes will be interpreted as directory separators instead of escapes. Using `cmd` would work if the user actually configured their `credential.helper` to use backslash separated paths, but that would break `git` - so it seems unlikely.

Also, the current defaults situation isn't quite as bad as I originally thought. When GCMC is installed bundled with git, its set up as `manager-core`, rather than the full path. Only users who installed GCMC separately will be broken (like me).

PR #682, opened 2021-03-15
Fix handling of URLs with port in credential.helper

sckfranke:
Credentials stored by Git for the URL `https://example.com:3000/foo/bar` use
`host=example.com:3000`. A CredentialHelper object for the same URL currently
queries the credential helper with `host=example.com` and thus can't access the
stored credentials.


alexcrichton:
Thanks!

PR #685, opened 2021-03-21
git2-hyper backend

henry40408:
First step of #623. The crate introduces rustls and hyper and tokio as async runtime.


SirWindfield:
Wouldn't it be better to have two new features? One that uses hyper with native tls and one that uses hyper with rustls?

git2-hyper/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        io::Error::new(io::ErrorKind::Other, err)
+    }
+
+    fn execute(&mut self, data: &[u8]) -> io::Result<()> {
+        if self.sent_request {
+            return Err(self.err("already sent HTTP request"));
+        }
+
+        // FIXME: wrap a runtime here is definitely NOT a good idea
+        let rt = tokio::runtime::Runtime::new().unwrap();
```
kim, line 140: I think that you might want to take a [Handle](https://docs.rs/tokio/1.4.0/tokio/runtime/struct.Handle.html) for `register`, which you then clone and pass around.

git2-hyper/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+                &format!(
+                    "expected a Content-Type header \
+                         with `{}` but didn't find one",
+                    expected
+                )[..],
+            ));
+        }
+
+        // Ok, time to read off some data.
+        let body = rt.block_on(res.body_mut().data());
```
kim, line 229: I’m not really familiar with hyper, but iiuc this will read the entire response body into memory, which is not so desirable. From the docs, I think you’ll want to execute the request here, and then store the response such that you can poll the body stream for the next chunk of data in the `Read` impl.

alexcrichton:
Thanks for the PR! This seems fine to me but I personally don't really know much about all the pieces involved here. For example I wouldn't be able to help out with bug reports, fix CI, etc. For those reasons could something like this live external to this repository? I'd be fine linking to a `git2-hyper` repository of course!

git2-hyper/src/lib.rs
henry40408: fixed in 165de0b

git2-hyper/src/lib.rs
henry40408: fixed in 8854f95

mitchmindtree:
@henry40408 exciting stuff - any follow-up on this? Quite interested for a downstream project that's looking to move away from openssl.

henry40408:
hi @mitchmindtree, I have opened a [repository](https://github.com/henry40408/git2-hyper) and prepare to publish on https://crates.io. stay tuned.

PR #686, opened 2021-03-23
Add bindings for the transaction API

kim:
Defined in `transaction.h`, this is essentially a way to place file locks on a set of refs, and update them in one (non-atomic) operation.

---

~~**Note** that this is currently based on 455098a8a1ce24305faf690264fc5b636d210739, as the commit after that causes some strange [linker errors](https://github.com/rust-lang/git2-rs/pull/671#discussion_r599524358) on my machine.~~

Rebased onto `master`


alexcrichton:
Looks great to me, thanks!

PR #691, opened 2021-03-28
Add bindings for mailmap

commonquail:
I was playing around with git2-rs and thought it might be fun to extend it, so I tried my hand at adding [mailmap](https://libgit2.org/libgit2/#HEAD/group/mailmap) support. FFI is new territory for me and everything I've done is based on existing code -- keep that in mind if you decide to review it.

The only reason I didn't add [`git_mailmap_resolve`](https://libgit2.org/libgit2/#HEAD/group/mailmap/git_mailmap_resolve) is that I couldn't come up with a signature for multiple return values.

I split the work in 2 commits because half the API strikes me as a bit pointless: there seems to be no way to write the mailmap representation to disk directly, or traverse it and write its contents by hand, so what's the point of being able to manipulate purely in-memory objects? Oh, well.


src/commit.rs
```diff
@@ -183,6 +183,20 @@ impl<'repo> Commit<'repo> {
         }
     }
 
+    /// Get the author of this commit, using the mailmap to map names and email
+    /// addresses to canonical real names and email addresses.
+    pub fn author_with_mailmap(&self, mailmap: &Mailmap<'_>) -> Result<Signature<'static>, Error> {
```
commonquail, line 188: `Signature<'static>` because `git_commit_author_with_mailmap` (`git_commit_committer_with_mailmap`) allocates a new signature the caller must free. `Signature::new` has the same return type.

src/mailmap.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::marker;
+use std::ptr;
+
+use crate::util::Binding;
+use crate::{raw, Error, Repository, Signature};
+
+/// A structure to represent a repository's .mailmap file.
+///
+/// The representation cannot be written to disk.
+pub struct Mailmap<'repo> {
```
commonquail, line 11: I'm not sure `'repo` makes sense. I implemented `git_mailmap_from_repository` first and copied this from somewhere else (`Commit`?), however, a `Mailmap` instance can exist independently of a `Repository` as long as `git_mailmap_new` and `git_mailmap_from_buffer` are also implemented.

src/commit.rs
alexcrichton: Yeah that should be fine I think!

src/mailmap.rs
alexcrichton: For this you'll probably want to check the implementation on the libgit2 side of things. If `git_mailmap_t` stores pointers to the `git_repository_t` then we'll need the lifetime, otherwise it's safe to remove.

commonquail: [It doesn't](https://github.com/libgit2/libgit2/blob/v1.1.0/src/mailmap.h#L20-L22), I've removed it.

alexcrichton:
Ok looks good to me! Want to rebase and I can merge?

PR #699, opened 2021-04-21
Use EUSER instead of generic error to stop foreach

### LINKED ISSUE #698: foreach callbacks use GIT_EUSER instead of GIT_ERROR to stop iteration
Hi, this is a minor nitpick about an otherwise very nice library. The callbacks to `Diff::foreach` get to return false to stop iterating. It's converted to `GIT_ERROR` in `diff::binary_cb_c` and friends. This is slightly ambiguous, as libgit2 could return `GIT_ERROR` for another reason. According to [libgit2's docs](https://github.com/libgit2/libgit2/blob/508361401fbb5d87118045eaeae3356a729131aa/include/git2/errors.h#L35), `GIT_EUSER` is reserved for this sort of use case.

Would you mind a PR for this?

danielzfranklin:
Previously foreach returned a generic generic error to signal the user asked to stop iterating. If a user got that error, they can't know for certain if it's because they asked to stop iterating, or if a legitimate error occurred.

libgit2 provides EUSER as a reserved error code for this case, so we use it.

Closes #698

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -338,9 +338,9 @@ pub extern "C" fn print_cb(
             (*data)(delta, hunk, line)
         });
         if r == Some(true) {
-            0
+            raw::GIT_OK
         } else {
-            -1
+            raw::GIT_EUSER
         }
     }
 }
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #701, opened 2021-04-25
Add a missing initialization in Diff::from_buffer

zertosh:
Without this, `Diff::from_buffers` fails with "an unknown git error occurred",
when run without having first called some other function that happens to
have initialized the git global allocator.


alexcrichton:
👍

PR #711, opened 2021-05-21
Add support for specifying custom headers on git push and fetch

arlosi:
Exposes the ability to set custom headers on push and fetch operations that are available from libgit2.

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -1,11 +1,12 @@
 use libc;
-use std::ffi::CString;
+use raw::git_strarray;
 use std::marker;
 use std::mem;
 use std::ops::Range;
 use std::ptr;
 use std::slice;
 use std::str;
+use std::{ffi::CString, os::raw::c_char};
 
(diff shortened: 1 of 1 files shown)


src/remote.rs
```diff
@@ -43,13 +44,17 @@ pub struct FetchOptions<'cb> {
     prune: FetchPrune,
     update_fetchhead: bool,
     download_tags: AutotagOption,
+    custom_headers: Vec<CString>,
+    custom_headers_ptrs: Vec<*const c_char>,
```
extrawurst, line 48: why do we need the duplication?
also I think the best practice is to keep stuff in `Option` by default.

alexcrichton: Nah I think this is ok, empty vecs don't need allocation and one here is the owned vector of strings and the other is what we pass to libgit2 itself

alexcrichton:
Thanks!

Issue #714, opened 2021-05-29
*name_is_valid() functions


extrawurst:
missing bindings for :

* [x] `git_branch_name_is_valid` (#715)
* [ ] `git_reference_name_is_valid`	
* [ ] `git_remote_name_is_valid`	
* [ ] ` git_tag_name_is_valid`

funny enough `git_remote_is_valid_name` and `git_reference_is_valid_name` is here :)

franciscod:
trying to set these up!

franciscod:
I'll leave the rest for somebody else to try. They're a good first PR!

leleliu008:
is there somebody who plan to fix the rest missing bindings?

extrawurst:
> is there somebody who plan to fix the rest missing bindings?

I think you can pick this one up - no one works on this right now

leleliu008:
OK. let me study some rust language syntax first. then I will try to fix it.

extrawurst:
after https://github.com/rust-lang/git2-rs/issues/734 we can at least go ahead with the branch name validity check

sanjayts:
I was interested in fixing this so have created a PR https://github.com/rust-lang/git2-rs/pull/882 ; would appreciate some feedback on it!

PR #715, opened 2021-05-29
add bindings for `git_branch_name_is_valid`

franciscod:
Hello! This is my first PR for Rust code. Feedback is greatly appreciated!

related: #714

### DIFF
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -28,6 +28,16 @@ impl<'repo> Branch<'repo> {
         Branch { inner: reference }
     }
 
+    /// Ensure the branch name is well-formed.
+    pub fn name_is_valid(name: &str) -> Result<bool, Error> {
+        let name = CString::new(name)?;
+        let mut valid: libc::c_int = 0;
+        unsafe {
+            try_call!(raw::git_branch_name_is_valid(&mut valid, name.as_ptr()));
+        }
+        Ok(valid == 1)
+    }
+
     /// Gain access to the reference that is this branch
     pub fn get(&self) -> &Reference<'repo> {
         &self.inner
@@ -151,7 +161,7 @@ impl<'repo> Drop for Branches<'repo> {
 
 #[cfg(test)]
 mod tests {
-    use crate::BranchType;
+    use crate::{Branch, BranchType};
 
     #[test]
     fn smoke() {
@@ -174,5 +184,13 @@ mod tests {
         b1.set_upstream(None).unwrap();
 
         b1.delete().unwrap();
+    }
+
+    #[test]
+    fn name_is_valid() {
+        assert!(Branch::name_is_valid("foo").unwrap());
+        assert!(!Branch::name_is_valid("").unwrap());
+        assert!(!Branch::name_is_valid("with spaces").unwrap());
+        assert!(!Branch::name_is_valid("~tilde").unwrap());
     }
 }
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2740,6 +2740,7 @@ extern "C" {
         force: c_int,
     ) -> c_int;
     pub fn git_branch_name(out: *mut *const c_char, branch: *const git_reference) -> c_int;
+    pub fn git_branch_name_is_valid(valid: *mut c_int, name: *const c_char) -> c_int;
     pub fn git_branch_remote_name(
         out: *mut git_buf,
         repo: *mut git_repository,


src/branch.rs
```diff
@@ -28,6 +28,16 @@ impl<'repo> Branch<'repo> {
         Branch { inner: reference }
     }
 
+    /// Ensure the branch name is well-formed.
+    pub fn name_is_valid(name: &str) -> bool {
+        let name = CString::new(name).unwrap();
+        unsafe { let mut valid: libc::c_int = 0;
```
extrawurst, line 34: I am pretty sure the formatting is off. please run `cargo fmt`

src/branch.rs
```diff
@@ -28,6 +28,16 @@ impl<'repo> Branch<'repo> {
         Branch { inner: reference }
     }
 
+    /// Ensure the branch name is well-formed.
+    pub fn name_is_valid(name: &str) -> bool {
```
extrawurst, line 32: it should be easy to write a unittest for this

extrawurst (changes requested):

src/branch.rs
franciscod: thanks!

src/branch.rs
franciscod: done, thanks!

src/branch.rs
```diff
@@ -28,6 +28,18 @@ impl<'repo> Branch<'repo> {
...
+        let name = CString::new(name).unwrap();
+        unsafe {
+            let mut valid: libc::c_int = 0;
+            // TODO:check that git_branch_name_is_valid returns 0?
+            // what should we do if it doesn't?
```
extrawurst, line 37: check out `delete` and the use of `try_call` to see how to handle the return values. this allows us to return a `Result`

franciscod: thanks for the pointers! returning `Result<bool>` now.

src/branch.rs
```diff
@@ -151,6 +161,7 @@ impl<'repo> Drop for Branches<'repo> {
 
 #[cfg(test)]
 mod tests {
+    use crate::Branch;
     use crate::BranchType;
```
neithernut, line 165: Tipp (since you implied you'd be somewhat new to rust): you can combine imports into a single line. In this case, for example, you could write `use crate::{Branch, BranchType}`. But I don't know what's the maintainers' preference.

src/branch.rs
```diff
@@ -28,6 +28,16 @@ impl<'repo> Branch<'repo> {
         Branch { inner: reference }
     }
 
+    /// Ensure the branch name is well-formed.
+    pub fn name_is_valid(name: &str) -> Result<bool, Error> {
+        let name = CString::new(name).unwrap();
```
extrawurst, line 33: Unwrap will panic at runtime. You can check other such CString usages in here but you can essentially propagate the error up by using `?` instead

src/branch.rs
franciscod: yes, I'm still unsure about these style decisions. thanks for the feedback anyway :)

src/branch.rs
franciscod: ah, didn't know about `?`... that seems like a really nice thing to use! also didn't know that unwrap panics. thanks for the feedback!

alexcrichton:
Thanks! Is this an API though where the initialization function needs to be called first? If so this function may need to have a call to `crate::init`?

franciscod:
Hmm I'm new to libgit2 as well. But let me try to answer anyway:

By looking at the function source code, it doesn't look like it needs any initialization:
https://github.com/libgit2/libgit2/blob/508361401fbb5d87118045eaeae3356a729131aa/src/branch.c#L740. After digging down some function calls, I've not seen any mention of an initialization function call being required.
Also I don't see any function in `src/branch.rs` or the related `src/reference.rs` which calls `crate::init`.

I'd say there's no need to call `crate::init`.

alexcrichton:
Ok!

decathorpe:
Is this function supposed to be present in libgit2 1.1.0? Because I'm getting linker errors while building the latest libgit2-sys / git2 crates (0.12.21+1.1.0 / 0.13.21) on Fedora (building against system libgit2):

```
git2-0.13.21/src/branch.rs:36: undefined reference to `git_branch_name_is_valid'
          collect2: error: ld returned 1 exit status
```

ehuss:
No, libgit2-sys now needs latest master.  That issue is tracked in #721.

PR #719, opened 2021-06-07
Fix vendored-openssl feature on macOS

### LINKED ISSUE #718: vendored-openssl broken with new resolver
over at `gitui` a user noted that we start to see `openssl` as a dylib dependency on macos:
```
otool -L target/debug/gitui
target/debug/gitui:
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59754.100.106)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1775.118.101)
	/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
```

no the interesting part is: we use git2 via:
```
git2 = { version = "0.13", features = ["vendored-openssl"] }
```

now this worked up until now but started to stop actually vendoring (and therefore not having a dependency on openssl shared objects) since we use `resolver = "2"`.

Now I am wondering is that something about how this feature is implemented or actually a bug in the new resolver ? 😱

see https://github.com/extrawurst/gitui/issues/772#issuecomment-855476599

alexcrichton:
Turns out the git2 dependency on openssl-sys was conditional to happen
everywhere but macOS, which means that vendored-openssl didn't work. The
fix here is to forward the feature to the `libgit2-sys` crate which
depends on `openssl-sys` even on macOS.

Closes #718

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ unstable = []
 default = ["ssh", "https", "ssh_key_from_memory"]
 ssh = ["libgit2-sys/ssh"]
 https = ["libgit2-sys/https", "openssl-sys", "openssl-probe"]
-vendored-openssl = ["openssl-sys/vendored"]
+vendored-openssl = ["openssl-sys/vendored", "libgit2-sys/vendored-openssl"]
 ssh_key_from_memory = ["libgit2-sys/ssh_key_from_memory"]
 zlib-ng-compat = ["libgit2-sys/zlib-ng-compat"]
 
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -36,6 +36,7 @@ openssl-sys = { version = "0.9", optional = true }
 ssh = ["libssh2-sys"]
 https = ["openssl-sys"]
 ssh_key_from_memory = []
+vendored-openssl = ["openssl-sys/vendored"]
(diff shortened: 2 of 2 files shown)


ehuss:
Is there a particular reason that openssl-sys is excluded from macos here?

https://github.com/rust-lang/git2-rs/blob/b055a28f0eebe491f4e2047be6dd3190e3c31060/Cargo.toml#L25

libgit2-sys does not have the same macos exclusion, so I believe it has no effect to have it.

alexcrichton:
Hm at this point it's probably historical for one reason or another, I don't recall the original reason but I would agree that it probably doesn't need to be present any more. (that being said it seemed fine to go ahead and add a `vendored-openssl` feature to `libgit2-sys` as well)

PR #722, opened 2021-06-16
Fix revparsing with windows paths in blame example

enizor:
`revparse_single` uses a git object name and not a file path - that means backslashes must be translated to slashes on windows.
Otherwise the example fails:
```
E:\dev\git2-rs>.\target\debug\blame.exe examples\blame.rs
error: the path 'examples\blame.rs' does not exist in the given tree; class=Tree (14); code=NotFound (-3)
```
I reused the `cfg!` conditional from `remote.rs`, but there might be a better way to do that.

### DIFF
--- a/examples/blame.rs
+++ b/examples/blame.rs
@@ -73,7 +73,15 @@ fn run(args: &Args) -> Result<(), git2::Error> {
         }
     }
 
-    let spec = format!("{}:{}", commit_id, path.display());
+    let spec = if cfg!(unix) {
+        format!("{}:{}", commit_id, path.display())
+    } else {
+        format!(
+            "{}:{}",
+            commit_id,
+            path.display().to_string().replace("\\", "/")
+        )
+    };
     let blame = repo.blame_file(path, Some(&mut opts))?;
     let object = repo.revparse_single(&spec[..])?;
     let blob = repo.find_blob(object.id())?;


examples/blame.rs
```diff
@@ -73,7 +73,15 @@ fn run(args: &Args) -> Result<(), git2::Error> {
...
     }
 
-    let spec = format!("{}:{}", commit_id, path.display());
+    let spec = if cfg!(unix) {
+        format!("{}:{}", commit_id, path.display())
+    } else {
+        format!(
+            "{}:{}",
+            commit_id,
+            path.display().to_string().replace("\\", "/")
```
extrawurst, line 82: like I mentioned in https://github.com/extrawurst/gitui/pull/791 I was under the impression that preventing the need for such manual hacks is the reason we have a type like `Path` in the first place - is `display()` the problem here? I am using `Path::to_str` in gitui to get the correct type of path seperators

enizor: As I understand it, using `revparse_single` with the syntax `<ref>:<path>` implies using `/`, as it refers to the path of a git object. 

`blame_file` uses a disk path and should (according to libgit2's conventions) accept `\` on Windows (in any case `git2-rs` normalizes with a call to `path_to_repo_path`).

Normalization is required for this example to accept Windows-style paths.

What I can see as a possible fix:
- normalize before the `revparse` call as in this PR (ugly)
- normalize inside the `revparse_single` function (weird IMO)
- remove revparse from this example
- accept that windows-style paths are rejected in this example (with a comment/help message for the `path` argument to clear things up)

PR #726, opened 2021-06-28
Improve docs and xrefs re tree building

ijackson:
I had to discover some of the facts here empirically so I thought I would write them up for other folks :-).

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -66,10 +66,10 @@ pub type RemoteCreate<'cb> =
 
 /// A builder struct for git tree updates.
 ///
-/// Paths passed to `remove` and `upsert` can be multi-component paths, ie they
+/// Paths passed to `remove` and `upsert` can be multi-component paths, i.e. they
 /// may contain slashes.
 ///
-/// This is the more-cooked tree update facility.  There is also [`TreeBuilder`]
+/// This is a higher-level tree update facility.  There is also [`TreeBuilder`]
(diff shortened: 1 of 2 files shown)


src/build.rs
```diff
@@ -64,7 +64,15 @@ pub struct RepoBuilder<'cb> {
 pub type RemoteCreate<'cb> =
     dyn for<'a> FnMut(&'a Repository, &str, &str) -> Result<Remote<'a>, Error> + 'cb;
 
-/// A builder struct for git tree updates, for use with `git_tree_create_updated`.
+/// A builder struct for git tree updates.
+///
+/// Paths passed to `remove` and `upsert` can be multi-component paths, ie they
+/// may contain slashes.
+///
+/// This is the more-cooked tree update facility.  There is also [`TreeBuilder`]
```
joshtriplett, line 72:
```suggestion
/// This is a higher-level tree update facility.  There is also [`TreeBuilder`]
```

(Parallel construction with "lower-level".)

joshtriplett:
I posted one very minor wording tweak, but otherwise this seems reasonable.

Would you consider adding coverage of the main *limitation* of the high-level API as well? Quoting the libgit2 documentation:

>  * Deleting and adding the same entry is undefined behaviour, changing
>  * a tree to a blob or viceversa is not supported.

ehuss (approved):
I went ahead and applied Josh's suggestion. Thanks!

PR #730, opened 2021-07-23
enable tracing and add tracing types and func

### LINKED ISSUE #696: bindings for git_trace_set
allowing to hook into libgit2 internal tracing might help diagnosing problems.
see https://libgit2.org/libgit2/#HEAD/group/trace/git_trace_set

PR follows

extrawurst:
* this enables tracing define (default in their CMake since 0.99)
* adds the callback type and sys function for this

in terms of adding a safe interface for this I will followup in a separate issue/PR since this probably requires adding `lazy static` as a dependency to safely store a safe callback globally.

closes #696


alexcrichton:
Thanks!

PR #732, opened 2021-07-24
Safe tracing

### LINKED ISSUE #731: safe interface for tracing
@alexcrichton like mentioned in #696 I would like to add a safe wrapper for `git_trace_set`. since it does not take a custom payload I can only imagine doing this via a global variable to store the user callback. this requires Mutex protection which in turn requires `lazy_static`

a) is that ok for you to add as a dependency?
b) do we have an alternative way to do this?
c) do we say such stuff belongs into user code or a utility crate?

extrawurst:
closes #731

@alexcrichton not sure if this is the approach you had in mind. if so I will clean it up, add documentation and fix the leak that happens if someone calls `trace_set` multiple times

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -1,0 +1,94 @@
+use lazy_static::lazy_static;
+use std::sync::Mutex;
+
+use libc::c_char;
+
+use crate::{panic, raw, util::Binding};
+
+/// Available tracing levels.  When tracing is set to a particular level,
+/// callers will be provided tracing at the given level and all lower levels.
+#[derive(Copy, Clone, Debug)]
+pub enum TraceLevel {
+    /// No tracing will be performed.
+    None,
+
+    /// Severe errors that may impact the program's execution
+    Fatal,
+
+    /// Errors that do not impact the program's execution
+    Error,
+
+    /// Warnings that suggest abnormal data
+    Warn,
+
+    /// Informational messages about program execution
+    Info,
+
+    /// Detailed data that allows for debugging
+    Debug,
+
+    /// Exceptionally detailed debugging data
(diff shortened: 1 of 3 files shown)


src/tracing.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+pub type TracingCb = fn(TraceLevel, &str);
+
+static CALLBACK: AtomicUsize = AtomicUsize::new(0);
+
+///
+pub fn trace_set(level: TraceLevel, cb: TracingCb) -> bool {
+    let boxed_cb = Box::new(cb);
+    CALLBACK.store(Box::into_raw(boxed_cb) as *mut _ as usize, Ordering::SeqCst);
```
alexcrichton, line 70: You can elide the `Box` here and cast the `TracingCb` directly to a `usize`

src/tracing.rs
```diff
@@ -0,0 +1,87 @@
...
+        raw::git_trace_set(level.raw(), Some(tracing_cb_c));
+    }
+
+    return true;
+}
+
+extern "C" fn tracing_cb_c(level: raw::git_trace_level_t, msg: *const c_char) {
+    panic::wrap(|| unsafe {
+        let cb = CALLBACK.load(Ordering::SeqCst);
+        let cb: Box<TracingCb> = Box::from_raw(cb as *mut _);
```
alexcrichton, line 82: Here you'll need to use a `transmute` to convert the `usize` to a `fn(TraceLevel, &str)`

extrawurst:
@alexcrichton thanks for the feedback! this got us rid of the box 👍

alexcrichton:
👍

PR #738, opened 2021-08-21
Add binding for git_repository_fetchhead_foreach

andrewhickman:


src/repo.rs
```diff
@@ -58,6 +63,35 @@ extern "C" fn mergehead_foreach_cb(oid: *const raw::git_oid, payload: *mut c_voi
...
+    is_merge: c_uint,
+    payload: *mut c_void,
+) -> c_int {
+    panic::wrap(|| unsafe {
+        let data = &mut *(payload as *mut FetchheadForeachCbData<'_>);
+        let res = {
+            let callback = &mut data.callback;
+
+            let ref_name = str::from_utf8(CStr::from_ptr(ref_name).to_bytes()).unwrap();
+            let remote_url = str::from_utf8(CStr::from_ptr(remote_url).to_bytes()).ok();
```
alexcrichton, line 79: Could this add asserts that the input pointers aren't null?

alexcrichton (commented):
Looks reasonable to me, thanks!

src/repo.rs
andrewhickman: I've added asserts for `ref_name`, `remote_url` and `oid`

src/repo.rs
```diff
@@ -58,6 +63,39 @@ extern "C" fn mergehead_foreach_cb(oid: *const raw::git_oid, payload: *mut c_voi
...
+            assert!(!ref_name.is_null());
+            assert!(!remote_url.is_null());
+            assert!(!oid.is_null());
+
+            let ref_name = str::from_utf8(CStr::from_ptr(ref_name).to_bytes()).unwrap();
+            let remote_url = str::from_utf8(CStr::from_ptr(remote_url).to_bytes()).ok();
+            let oid = Binding::from_raw(oid);
+            let is_merge = is_merge == 1;
+
+            callback(&ref_name, remote_url.as_deref(), &oid, is_merge)
```
alexcrichton, line 87: Typically `Option<&str>` when used in these bindings represents a possibly-not-present `&str`, but here it's being used exclusively for not-valid-utf8 `&str`. Whenever that's handled in `git2` it's typically done with some form of both a bytes and a str accessor, although I don't think that's handled with callbacks today.

If this is always present, could this argument for `remote_url` be `&str`? We can deal with the actually-this-may-not-be-utf8 in the future.

andrewhickman: Are you suggesting just to unwrap it for now (like I did for `ref_name`)?

alexcrichton: I don't know, unfortunately. If this is an actually nullable pointer coming from libgit2 then this should be `Option<&str>`. If it's always present coming from libgit2 then it should be `&str`. I don't know which is the case, though.

andrewhickman: I've made that change now

alexcrichton:
Thanks!

Issue #721, opened 2021-06-13
Undefined symbols for architecture x86_64: "_git_branch_name_is_valid"


arxanas:
As of 0454efbaa0447a1d02eb11459a7db5e1c27d98e0, `cargo test` fails on macOS.

System details:

````
$ uname -a
Darwin waleedkhan-mbp 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:07:06 PST 2021; root:xnu-7195.81.3~1/RELEASE_X86_64 x86_64
$ rustup show
Default host: x86_64-apple-darwin
rustup home:  /Users/Waleed/.rustup

stable-x86_64-apple-darwin (default)
rustc 1.52.1 (9bc8c42bb 2021-05-09)
```

Error message:

```
   Compiling git2 v0.13.20 (/Users/Waleed/workspace/git2-rs)
error: linking with `cc` failed: exit code: 1
  |
[...]

PR #739, opened 2021-08-23
Add `vendored` feature

arxanas:
As per https://github.com/rust-lang/git2-rs/issues/721, some consumers may
prefer to build `libgit2` themselves to simplify distribution.

To test the vendored version, you can run:

```
cargo build --features vendored-libgit2`
```

That being said, the test `branch::tests::name_is_valid` fails on macOS. This
test is currently broken as per
https://github.com/rust-lang/git2-rs/issues/721#issuecomment-860256600


alexcrichton:
Personally I think this is a case where the burden of getting things working shouldn't be placed on the contributor but rather this crate itself. If libgit2 is stable enough that there's different versions in the wild that we're using, then the API of this crate should be adapted when using an older version or the system version should be rejected if it isn't new enough.

arxanas:
> the burden of getting things working shouldn't be placed on the contributor

Who do you mean by the "contributor" in this case? The prospective contributor to `git2-rs`, or to `libgit2`? I don't follow how adding a `vendored` feature affects either of those parties.

> the API of this crate should be adapted when using an older version or the system version should be rejected if it isn't new enough

Whose use-case is this helping? I guess you mean for people writing libraries on top of `git2-rs`, who would prefer to integrate with the system `libgit2` if available?

In the case of end-user applications like my https://github.com/arxanas/git-branchless or @extrawurst's https://github.com/extrawurst/gitui, I think we would like to have an option where we can significantly simplify the distribution/compilation for our users who aren't consuming our applications via package manager, such as when compiling from source or downloading an executable directly from Github Releases.

PR #748, opened 2021-09-27
Add way to get build version information.

ehuss:
This adds a `Version` struct which can be used to query the version information of libgit2.

My use case is to expose this information in Cargo (per https://github.com/rust-lang/cargo/issues/6275).


alexcrichton:
👍

PR #749, opened 2021-09-27
Add binding for `git_message_trailers`

LemmingAvalanche:


LemmingAvalanche:
I don’t know how to fix the current error from the CI.

The output on my machine:

```
$ if [[ "stable" != "nightly" ]]; then cargo run --manifest-path systest/Cargo.toml; fi
…
  running: "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "<path-to-repo>/target/debug/build/libgit2-sys-b48a2ccdd54f1249/out/include" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-unknown-warning-option" "-Wno-deprecated-declarations" "-o" "<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.o" "-c" "<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.c"
  cargo:warning=<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.c: In function ‘__test_field_type_git_message_trailer_array_trailers’:
  cargo:warning=<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.c:18438:28: error: return from incompatible pointer type [-Werror=incompatible-pointer-types]
  cargo:warning=                     return &b->trailers;
  cargo:warning=                            ^~~~~~~~~~~~
  cargo:warning=<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.c: At top level:
  cargo:warning=cc1: error: unrecognized command line option ‘-Wno-unknown-warning-option’ [-Werror]
  cargo:warning=cc1: error: unrecognized command line option ‘-Wno-address-of-packed-member’ [-Werror]
  cargo:warning=cc1: all warnings being treated as errors
  exit code: 1

  --- stderr
  rust version: 1.47.0


  error occurred: Command "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-I" "<path-to-repo>/target/debug/build/libgit2-sys-b48a2ccdd54f1249/out/include" "-Wall" "-Wextra" "-Wall" "-Wextra" "-Werror" "-Wno-unused-parameter" "-Wno-type-limits" "-Wno-address-of-packed-member" "-Wno-unknown-warning-option" "-Wno-deprecated-declarations" "-o" "<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.o" "-c" "<path-to-repo>/target/debug/build/systest-ae23871569b4b2ee/out/all.c" with args "cc" did not execute successfully (status code exit code: 1).
```

The C function at line 18438 in that file:

``` c
const git_message_trailer** __test_field_type_git_message_trailer_array_trailers(git_message_trailer_array* b) {
    return &b->trailers;
}
```

The error output from the CI is better than the one on my machine:

```
error: returning ‘git_message_trailer **’ {aka ‘struct <anonymous> **’} from a function with incompatible return type ‘const git_message_trailer **’ {aka ‘const struct <anonymous> **’} [-Werror=incompatible-pointer-types]
```

How can I provide a better type in `src/libgit2-sys/lib.rs`?

alexcrichton:
That looks like a difference of `*const` and `*mut` in Rust I think?

LemmingAvalanche:
Oh wow.  Yep, that was it.  Thanks!

src/message.rs
```diff
@@ -28,15 +31,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
     Ok(ret.as_str().unwrap().to_string())
 }
 
+/// Collection of trailer key–value pairs.
+///
+/// Use `iter()` to get access to the values.
+pub struct MessageTrailers<'pair> {
```
alexcrichton, line 37: I believe that this is a uniquely owned value, so I don't think that a `'pair` lifetime parameter here is necessary.

src/message.rs
```diff
@@ -28,15 +31,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+
+impl<'pair> Drop for MessageTrailers<'pair> {
+    fn drop(&mut self) {
+        unsafe {
+            raw::git_message_trailer_array_free(&mut self.raw);
+        }
+    }
+}
+
+impl<'pair> Binding for MessageTrailers<'pair> {
```
alexcrichton, line 74: I think this may not actually be necessary?

src/message.rs
```diff
@@ -28,15 +31,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+            let value = CStr::from_ptr(self.value).to_str().unwrap();
+            (key, value)
+        }
+    }
+}
+
+/// A borrowed iterator.
+pub struct MessageTrailersIterator<'pair> {
+    trailers: &'pair MessageTrailers<'pair>,
+    counter: usize,
```
alexcrichton, line 106: One thing I might recommend is to change this to `Range<usize>` so that way `next()` is easy to implement.

src/message.rs
```diff
@@ -28,15 +31,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+            raw: *raw,
+            _marker: marker::PhantomData,
+        }
+    }
+    fn raw(&self) -> *mut raw::git_message_trailer_array {
+        &self.raw as *const _ as *mut _
+    }
+}
+
+struct Trailer<'pair> {
```
alexcrichton, line 87: This is a one-off struct just for the method below so I think it's fine to remove this and just inline the method into the iterator implementation.

src/message.rs
```diff
@@ -28,15 +31,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+                        key: (*addr).key,
+                        value: (*addr).value,
+                        _marker: marker::PhantomData,
+                    }
+                    .to_str_tuple(),
+                )
+            }
+        }
+    }
+}
```
alexcrichton, line 130: Could you add implementations for `size_hint` and `ExactSizeIterator` here as well? (which would also necessitate `DoubleEndedIterator`, and the implementations should be relatively easy if the `counter` field is changed to a `Range`)

alexcrichton (commented):
Thanks again for this! I left some comments but otherwise this is looking pretty good!

src/message.rs
LemmingAvalanche: Never heard of, but I like it!

See rev. e94008a00fd

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
 
+/// Collection of trailer key–value pairs.
+///
+/// Use `iter()` to get access to the values.
+pub struct MessageTrailers {
+    raw: raw::git_message_trailer_array,
+    _marker: marker::PhantomData<c_char>,
+}
+
+impl<'pair> MessageTrailers {
```
alexcrichton, line 43: Since the type doesn't have a lifetime parameter any more this shoudl move down to the `iter` method itself, but even on the iter method it can be elided and it can return `MessageTrailersIterator<'_>`

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
     Ok(ret.as_str().unwrap().to_string())
 }
 
+/// Collection of trailer key–value pairs.
+///
+/// Use `iter()` to get access to the values.
+pub struct MessageTrailers {
+    raw: raw::git_message_trailer_array,
+    _marker: marker::PhantomData<c_char>,
```
alexcrichton, line 40: This I don't believe is necessary any longer.

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+}
+
+/// A borrowed iterator.
+pub struct MessageTrailersIterator<'a> {
+    trailers: &'a MessageTrailers,
+    range: Range<usize>,
+}
+
+impl<'pair> Iterator for MessageTrailersIterator<'pair> {
+    type Item = (&'pair str, &'pair str);
```
alexcrichton, line 98: Could this perhaps be an iterator of `&[u8]` instead of `&str` to allow for non-utf8 messages?

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+        let addr = trailers.raw.trailers.wrapping_add(index);
+        let key = CStr::from_ptr((*addr).key).to_str().unwrap();
+        let value = CStr::from_ptr((*addr).value).to_str().unwrap();
+        (key, value)
+    }
+}
+
+impl ExactSizeIterator for MessageTrailersIterator<'_> {
+    fn len(&self) -> usize {
+        self.range.end
```
alexcrichton, line 123: This'll want to be `self.range.len()` since the length changes over time.

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+}
+
+impl<'pair> Iterator for MessageTrailersIterator<'pair> {
+    type Item = (&'pair str, &'pair str);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.range
+            .next()
+            .map(|index| to_str_tuple(&self.trailers, index, marker::PhantomData))
+    }
```
alexcrichton, line 104: Could you add `size_hint` here too?

src/message.rs
```diff
@@ -28,15 +32,129 @@ fn _message_prettify(message: CString, comment_char: Option<u8>) -> Result<Strin
...
+}
+
+#[inline(always)]
```
alexcrichton, line 107: This annotation isn't necessary

src/message.rs
LemmingAvalanche: See rev a0bd02601

src/message.rs
LemmingAvalanche: See rev 34fa7b51ce

src/message.rs
LemmingAvalanche: See rev 023fb921

src/message.rs
LemmingAvalanche: Ah okay I think I understand now.  I confused myself thought that `size_hint()` was part of `ExactSizeIterator`.

See rev 16321f38702

src/message.rs
LemmingAvalanche: See rev 70b7ebd96

PR #763, opened 2021-11-11
Make `Odb` `Send` and `Sync`

joshtriplett:
As of libgit2 1.2.0, `git_odb` uses locking internally, and should be thread-safe. Mark it `Send` and `Sync` to allow access from multiple threads.


joshtriplett:
One day I will get the marker-trait impl syntax right on the first try. But it is not this day.

joshtriplett:
Also see https://github.com/libgit2/libgit2/pull/6109

alexcrichton (approved):
Seems good to me, but let's wait on merging until the upstream docs are merged?

joshtriplett:
@alexcrichton It has now been merged.

PR #770, opened 2021-11-20
we have to expect non-utf8 strings in tracing messages

### LINKED ISSUE #769: panic in tracing
```
06:45:16 [ERROR] panic: PanicInfo { payload: Any { .. }, message: Some(called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 17, error_len: Some(1) }), location: Location { file: "/Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/github.com-1ecc6299db9ec823/git2-0.13.21/src/tracing.rs", line: 79, col: 58 } }
trace:
   0: backtrace::capture::Backtrace::new
   1: gitui::set_panic_handlers::{{closure}}
   2: std::panicking::rust_panic_with_hook
   3: std::panicking::begin_panic_handler::{{closure}}
   4: std::sys_common::backtrace::__rust_end_short_backtrace
   5: _rust_begin_unwind
   6: core::panicking::panic_fmt
   7: core::result::unwrap_failed
   8: core::result::Result<T,E>::unwrap
   9: git2::tracing::tracing_cb_c
  10: _git_trace__write_fmt
  11: _stream_write
  12: _git_http_client_send_body
  13: _http_stream_write
  14: _stream_thunk
  15: _write_one
  16: _write_pack
  17: _git_smart__push
  18: _git_push_finish
  19: _git_remote_upload
  20: _git_remote_push
  21: git2::remote::Remote::push
  22: std::sys_common::backtrace::__rust_begin_short_backtrace
  23: core::ops::function::FnOnce::call_once{{vtable.shim}}
  24: std::sys::unix::thread::Thread::new::thread_start
  25: _pthread_jit_write_protect_np
```
this was reported here: https://github.com/extrawurst/gitui/issues/881#issuecomment-974563786

probably caused because our tracing implementation (the callback) did not expect libgit2 sending non-utf8 tracing. turns out it factually traces arbitrary buffers. so its no wonder out `to_str.unwrap` chokes on it 🙈

extrawurst:
fixes #769

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -57,6 +57,12 @@ impl Binding for TraceLevel {
     }
 }
 
+/// Callback type used to pass tracing events to the subscriber.
+/// see `trace_set` to register a scubscriber.
+///
+/// Note:
+///     libgit2 might pass non-utf8 strings therefore we
+///     pass the message as a byte slice
 pub type TracingCb = fn(TraceLevel, &[u8]);
 
 static CALLBACK: AtomicUsize = AtomicUsize::new(0);
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -76,7 +76,7 @@ extern "C" fn tracing_cb_c(level: raw::git_trace_level_t, msg: *const c_char) {
     let cb = CALLBACK.load(Ordering::SeqCst);
     panic::wrap(|| unsafe {
         let cb: TracingCb = std::mem::transmute(cb);
-        let msg = std::ffi::CStr::from_ptr(msg).to_str().unwrap();
-        cb(Binding::from_raw(level), msg);
+        let msg = std::ffi::CStr::from_ptr(msg).to_string_lossy();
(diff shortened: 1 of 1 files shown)


alexcrichton:
Should the API here actually take `&[u8]` instead of `&str` to reflect how it's not always utf-8?

extrawurst:
makes sense. will update that

extrawurst:
ok that will break the api though, still ok?

extrawurst:
@alexcrichton in case API breakage is fine then the PR should be good to go

extrawurst:
If we do not want to break the API we would need a duplication of the API with the changed callback signature. let me know if you want that, then I update the PR

extrawurst:
@ehuss maybe you can chime in here?

alexcrichton:
Hm right yeah that's true, let's keep the lossy string interpretation and this can be updated later during the next major version bump.

PR #773, opened 2021-11-20
support fetch_all

### LINKED ISSUE #772: support fetching all remotes heads
`git_remote_fetch` supports passing NULL for the specs array: https://libgit2.org/libgit2/#HEAD/group/remote/git_remote_fetch

we currently do not support that.

extrawurst:
closes #772

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -288,6 +288,25 @@ impl<'repo> Remote<'repo> {
         Ok(())
     }
 
+    /// like `fetch` but passing NULL for `refspecs` array effectively fetching all
+    pub fn fetch_all(
+        &mut self,
+        opts: Option<&mut FetchOptions<'_>>,
+        reflog_msg: Option<&str>,
+    ) -> Result<(), Error> {
+        let msg = crate::opt_cstr(reflog_msg)?;
+        let raw = opts.map(|o| o.raw());
+        unsafe {
+            try_call!(raw::git_remote_fetch(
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think it might be best to have `Option` around the refspecs on the `fetch` API perhaps? That should more closely mirror the underlying libgit2 operation I think.

Although in the spirit of not having breaking changes I'd also be fine having this but lodging a TODO or something somewhere to come back and tweak this API for the next major version bump.

extrawurst:
ok I will place a TODO, for future merging of both

src/remote.rs
```diff
@@ -288,6 +288,25 @@ impl<'repo> Remote<'repo> {
         Ok(())
     }
 
+    /// like `fetch` but passing NULL for `refspecs` array effectively fetching all
+    pub fn fetch_all(
```
ehuss, line 292: I feel like `all` may be a bit misleading, as it fetches the default remote refspecs, which may not be everything.

Is there a particular reason that using `fetch` with an empty refspec list isn't sufficient to handle this use case?

extrawurst:
@ehuss thanks for your comment. I don't know why I did not check that earlier but yeah: https://github.com/libgit2/libgit2/blob/e61abdce434e17104b40c09b34094e2d571ba9e9/src/remote.c#L1089

obviously empty array and NULL are treated the same. this PR is not needed, my bad

PR #775, opened 2021-11-28
chore: Bump thread-id to 4

ignatenkobrain:


extrawurst:
your PR made me wonder where we actually use it, we don't: https://github.com/rust-lang/git2-rs/pull/776

ignatenkobrain:
That's even better :)

PR #776, opened 2021-11-28
remove unused dependency

extrawurst:


ehuss (approved):
Thanks!  Looks like `-Z minimal-versions` works without it now. Presumably whatever issue since #336 has been resolved.

PR #779, opened 2021-11-29
update dependencies, do not pin patch release

extrawurst:


ehuss:
Can you say why this removes the patch numbers?  That sets the floor of the minimum version needed.

extrawurst:
its easier for code using this crate to update dependencies if they are not pinned hard like this.
how can I figure out the msrv by a certain dependency change?

ehuss:
Specifying all 3 components doesn't pin a dependency, it just sets the minimum (semver) compatible version.  Specifying x.y is the same as specifying x.y.0.

As it stands, these changes won't affect the versions typically picked up by users of this library. Cargo picks the newest semver compatible version when updating `Cargo.lock`.  Usually, the only reason to update a minor version in a dependency is if you want to force existing users to a newer version (for example, if a change to git2-rs uses a function was added in a newer version of a dependency).

As for determining msrv, you have to just build/test with older versions until it fails.

Issue #778, opened 2021-11-28
Doing in-memory merge conflict resolution


fin-ger:
Hi all,

first of all, I would like to thank you for providing this awesome libgit2 wrapper, including the searchable docs.rs documentation. This already helped me a lot in my project!

I am currently struggling to get an in-memory merge conflict resolution done with the libgit2. But let's first explain why I am seeking for in-memory conflict resolution:

 - I am currently building a rust-library for the [pass password store](https://www.passwordstore.org/). I would like to provide an easy-to-use API backed by git for decentralized synchronization (e.g. without a server) from device-to-device over e.g. the local network, ad-hoc wifi, etc. The original pass implementation (shell-script) just delegates the synchronization to git by mirroring the git cli. I therefore decided to base my synchronization approach on git as well, to make it (at least theoretically) compatible.
 - As from a user perspective a synchronization is either successful or not (maybe requiring some manual conflict resolution via e.g. a GUI popup), I would like to do the messing with git stages, etc. in-memory to prevent interfering with other tools handling a `pass` password store (e.g. the original cli, the qt intercace, the Firefox extension, etc.). Therefore, I thought that checking out a conflicted index to the repository is not a good idea. Also, if my lib crashes, is quit by the user, etc. my library would have to somehow continue a conflict resolution of a previous run. Furthermore, if multiple instances of my library or another pass implementation are running on the same repository I might run into race condition problems, etc. Those problems are ofc fixable with e.g. locks but it makes the implementation way more complex. That's why I settled with in-memory conflict resolution of a conflicted index.

So, what exactly is my problem? I have (somewhat) the following code:

```rust
let mut index = self.repo.merge_trees(&ancestor, &local_tree, &remote_tree, None)?;
// go over index.conflicts() and resolve them
// in pass passwords are gpg encrypted, not really relevant for this issue
index.add_frombuffer(new_index_entry, &encrypted_resolved_password_content)?;
```

However, this errors with `could not initialize index entry. Index is not backed up by an existing repository.` Which is true, I could not (yet) do e.g. a `index.write_tree_to(repo)` as my index still contains conflicts. So I tried another variant:

```rust
let mut index = self.repo.merge_trees(&ancestor, &local_tree, &remote_tree, None)?;
// go over index.conflicts() and resolve them
let oid = repo.blob(&encrypted_resolved_password_content)?;
index_entry.file_size = encrypted_resolved_password_content.len() as u32;
index_entry.id = oid;
[...]

PR #780, opened 2021-11-30
WIP: Add wrappers for git_index_conflict_{add,remove,get,cleanup}

fin-ger:
In response to #778 I added wrapper functions for

 - `git_index_conflict_add`
 - `git_index_conflict_remove`
 - `git_index_conflict_cleanup`
 - `git_index_conflict_get` 

However, I do not know if this trivial addition is actually safe, as conflicts are already exposed via an iterator. I do not know whether it is safe to modify the index while iterating over the conflicts.

Suggestions welcome!

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -85,6 +85,70 @@ pub struct IndexEntry {
     pub path: Vec<u8>,
 }
 
+// We cannot return raw::git_index_entry instances directly, as they rely on
+// a CString which is owned by the function. To make the pointer to the CString
+// valid during usage of raw::git_index_entry, we supply the index entry in a
+// callback where pointers to the CString are valid.
+fn try_raw_entries(
+    entries: &[Option<&IndexEntry>],
+    cb: impl FnOnce(&[*const raw::git_index_entry]) -> Result<(), Error>,
+) -> Result<(), Error> {
+    let paths = entries.iter()
+        .map(|entry| {
+            if let Some(entry) = entry {
+                CString::new(&entry.path[..]).map(|ok| Some(ok))
+            } else {
+                Ok(None)
+            }
+        })
+        .collect::<Result<Vec<Option<CString>>, std::ffi::NulError>>()?;
+
+    let raw_entries = entries.iter().zip(&paths).map(|(entry, path)| {
+        if let Some(entry) = entry {
+            // libgit2 encodes the length of the path in the lower bits of the
+            // `flags` entry, so mask those out and recalculate here to ensure we
+            // don't corrupt anything.
+            let mut flags = entry.flags & !raw::GIT_INDEX_ENTRY_NAMEMASK;
+
+            if entry.path.len() < raw::GIT_INDEX_ENTRY_NAMEMASK as usize {
+                flags |= entry.path.len() as u16;
+            } else {
+                flags |= raw::GIT_INDEX_ENTRY_NAMEMASK;
+            }
+
+            unsafe {
+                Some(raw::git_index_entry {
+                    dev: entry.dev,
+                    ino: entry.ino,
+                    mode: entry.mode,
+                    uid: entry.uid,
+                    gid: entry.gid,
+                    file_size: entry.file_size,
+                    id: *entry.id.raw(),
+                    flags,
+                    flags_extended: entry.flags_extended,
+                    path: path.as_ref().unwrap().as_ptr(),
+                    mtime: raw::git_index_time {
+                        seconds: entry.mtime.seconds(),
+                        nanoseconds: entry.mtime.nanoseconds(),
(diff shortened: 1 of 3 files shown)


src/index.rs
```diff
@@ -412,6 +412,191 @@ impl Index {
...
     }
 
+    /// Add or update index entries to represent a conflict. Any staged entries
+    /// that exist at the given paths will be removed.
+    ///
+    /// The entries are the entries from the tree included in the merge. Any entry
+    /// may be `None` to indicate that that file was not present in the trees during
+    /// the merge. For example, ancestor_entry may be `None` to indicate that a file
+    /// was added in both branches and must be resolved.
+    pub fn conflict_add(
```
alexcrichton, line 422: The general intention of this crate is to be a lightweight binding over libgit2 itself, but this is a pretty heavyweight method. Would it be possible to avoid the movement around of all the data internally?

fin-ger: I based this implementation on all the other methods taking a `IndexEntry` (e.g. [`Index::add()`](https://github.com/rust-lang/git2-rs/blob/master/src/index.rs#L147). This one is particularly long because it takes three `IndexEntry` parameters. I don't know if the handling of `IndexEntry`s can be more concise, my intention was more to handle them exactly the same as in the existing methods.

fin-ger: Maybe [`std::mem::transmute`](https://doc.rust-lang.org/stable/std/mem/fn.transmute.html) can be used here, but I have no clue if `git_index_entry` is even remotely binary compatible with an `IndexEntry` especially regarding the `path: Vec<u8>`. Also, this would be incredibly easy to mess up and hard to debug :see_no_evil: 

Maybe someone has a better idea :smile:

alexcrichton: Ah ok, well in that case can the conversion into a `git_index_entry` be refactored into a helper function instead of being duplicated?

fin-ger: Sure, will do so. Should I also refactor all conversions in other methods using `IndexEntry`?

alexcrichton: I guess? Sort of depends, I'm not intimately familiar with all this code.

fin-ger: Alright, my solution is to use a helper function which supplies pointers to `raw::git_index_entry` instances via a callback. This is needed as the construction of a `git_index_entry` requires the construction of a `CString` to which the `git_index_entry` points to. I went for a callback solution as it is (in my opinion) the easiest approach to make sure that the `CString` is available while using the raw `git_index_entry`. The procedure is implemented in [`try_raw_entries()@src/index.rs:92`](https://github.com/fin-ger/git2-rs/blob/add-index-conflict-fns/src/index.rs#L92). The implementation is a bit messy and requires the construction of three `Vec`s. Suggestions welcome!

src/index.rs
```diff
@@ -85,6 +85,76 @@ pub struct IndexEntry {
     pub path: Vec<u8>,
 }
 
+// We cannot return raw::git_index_entry instances directly, as they rely on
+// a CString which is owned by the function. To make the pointer to the CString
+// valid during usage of raw::git_index_entry, we supply the index entry in a
+// callback where pointers to the CString are valid.
+fn try_raw_entries(
```
alexcrichton, line 92: I think this function would be easier to use as:

```rust
impl IndexEntry {
    fn to_raw(&self, storage: &mut Vec<CString>) -> raw::git_index_entry {
        // ...
    }
}
```

which would avoid the vector, allocations internally, and callback style

fin-ger: In my approach the `Vec`s are necessary as I need multiple entries available in the callback, therefore `try_raw_entries` can handle multiple entries.

With your approach I think something like this would work:

```
impl IndexEntry {
    fn to_raw(&self, cpath: &CString) -> raw::git_index_entry {
        // ...
    }
}
```

Although, I think this would still not ensure that `cpath` lives longer than the returned value. The usage would look like this:

```
let cpath = CString::new(&self.path[..])?;
let raw = self.to_raw(&cpath);
```

My last commit replaces the `Vec`s with arrays but still contains the callback. If it is safe to have a pointer pointing to the `cpath` parameter in the return value, I would convert the implementation to that approach. But I'm unsure whether `cpath` is actually always freed after the return value as it has no lifetime attached to it. I am not that familiar with the internals of memory management in Rust.

fin-ger: I implemented your callback-less approach and checked if the tests are passing. Looks good so far, but I'm still not sure whether this API is safe: the return value references data of a parameter but does not carry the lifetime of the parameter to indicate that `c_path` must live longer than the returned value. Maybe this is not necessary as the `c_path` binding is always created before the return value in assigned to a variable, I don't know.

Issue #140, opened 2016-05-11
Repository::commit should accept parents as &[Commit], not just &[&Commit]


joshtriplett:
If you want to generate a list of commits to pass to `.commit` as parents, the type `&[&Commit]` proves inconvenient, because it requires the caller to maintain ownership of the Commit objects for long enough to pass them.  For instance, this could require generating a `Vec<Commit>`, and then an additional `Vec<&Commit>`.

Could Repository::commit accept a &[T] where T=AsRef<Commit>, instead?  That would allow both possibilities.  It does, however, have the disadvantage that passing `&[]` will result in an ambiguous type, though.  Another possibility would be always accepting `&[Commit]`, but that would break existing code that passes `&[&Commit]`.

I'd welcome other suggestions that would simplify code like this:
```rust
    let mut parents : Vec<git2::Commit> = Vec::new();
    parents.extend(tree.iter().filter_map(|e| {
        if e.kind() == Some(git2::ObjectType::Commit) {
            Some(repo.find_commit(e.id()).unwrap_or_else(die))
        } else {
            None
        }
    }));
    let new_commit_oid = repo.commit(Some("SHEAD"), &author, &committer, &msg, &tree, &parents).unwrap_or_else(die);
```

alexcrichton:
Yeah I wonder if this could actually look like:

```rust
pub fn commit<I>(&self, ..., parents: I)
    where I: IntoIterator,
          I::Item: AsRef<Commit>
```

I *think* that'd be backwards compatible? I'll probably do a new major release soon anyway due to `IndexEntry` so strict compat may not matter too much.

joshtriplett:
@alexcrichton That's not quite backward compatible, because it wouldn't know what type to use if the caller passes `&[]`.  It's really unfortunate to have to write `&[] as &[Commit]`.  Hopefully, when default type parameter fallback gets stabilized, that'll help simplify cases like this.

alexcrichton:
Gah right, that seems to keep biting this...

pwaller:
Just hit this. Rust newbie here. My use case was to create a commit based on another commit. Surprised to find I couldn't:

```
repo.commit(..., old_commit.parents())
```

I tried:
```
&*c.parents().collect::<Vec<_>>().into_boxed_slice(),
```

To discover:
```
   = note: expected reference `&[&git2::Commit<'_>]`
              found reference `&[git2::Commit<'_>]`
```

[...]

PR #789, opened 2022-01-06
Add Repository::commit_new method with permissive parent param

chetaldrich:
The existing method doesn't accept owned commits as parameters in
Repository::commit. This new method allows for both owned and unowned
versions of the slice, which will simplify passing in Vectors
containing owned values as well as retain backwards compatibility in
_most_ cases (unfortunately, an empty slice still needs a type specified if passed).

Fixes https://github.com/rust-lang/git2-rs/issues/140, which specifies the issue in more detail.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -1231,6 +1231,56 @@ impl Repository {
     /// current branch and make it point to this commit. If the reference
     /// doesn't exist yet, it will be created. If it does exist, the first
     /// parent must be the tip of this branch.
+    pub fn new_commit<'a, C: AsRef<Commit<'a>>>(
+        &self,
+        update_ref: Option<&str>,
+        author: &Signature<'_>,
+        committer: &Signature<'_>,
+        message: &str,
+        tree: &Tree<'_>,
+        parents: &[C],
+    ) -> Result<Oid, Error>
+    where
+        C: AsRef<Commit<'a>>,
+    {
+        let update_ref = crate::opt_cstr(update_ref)?;
+        let mut parent_ptrs = parents
+            .iter()
+            .map(|p| p.as_ref().raw() as *const raw::git_commit)
+            .collect::<Vec<_>>();
+        let message = CString::new(message)?;
+        let mut raw = raw::git_oid {
+            id: [0; raw::GIT_OID_RAWSZ],
+        };
+        unsafe {
+            try_call!(raw::git_commit_create(
+                &mut raw,
+                self.raw(),
+                update_ref,
+                author.raw(),
+                committer.raw(),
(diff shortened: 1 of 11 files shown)


src/repo.rs
```diff
@@ -1231,6 +1231,56 @@ impl Repository {
     /// current branch and make it point to this commit. If the reference
     /// doesn't exist yet, it will be created. If it does exist, the first
     /// parent must be the tip of this branch.
+    pub fn new_commit<'a, C: AsRef<Commit<'a>>>(
```
extrawurst, line 1234: can we add this `&[] as &[Commit<'_>]` for the empty list case to the documentation as this will be a little unintuitive 🙈

extrawurst:
I personally would prefer `commit_new` but thats a nitpicky

src/repo.rs
chetaldrich: Sure, I'll add that in. It's really too bad the default type parameter fallback feature didn't make it into the language since this issue was created.

chetaldrich:
I can switch over to commit_new since I'll be making that other small documentation change anyway.

src/repo.rs
alexcrichton: I think this may have actually been one of the issues which led me to the original signature, if this otherwise requires type annotations then I don't think that this is the best way to shape this API. Another option would be perhaps a builder-like construct but that's pretty wordy for just the type inference issue here.

chetaldrich: So there are two options here: 

1. We can stick with this, which allows existing calls that contain values to be the same (appears to be the majority of calls, based on some seen usage of this method in the wild and in the tests), but does have this edge case that Rust doesn't handle very gracefully, while supporting owned values in the slice/vector parameter. 
2. We could make a hard switch to using owned values, so the signature would be `&[Commit]` in that case. We lose compatibility with existing calls, but this edge case type inference issue goes away. 

I personally think that supporting owned values makes sense here, thus the reason for the PR, but would like to support the existing usage where possible. This makes the former seem like a good compromise to me since it only comes up in this one case, but I could go either way. 

Thoughts?

chetaldrich:
I've updated the PR to incorporate the suggestions of @extrawurst, but if we decide that the type inference issue makes us want to go the `&[Commit]` route I'm happy to make the needed changes.

alexcrichton:
Personally I still feel like `&[&Commit<'_>],` is the best for now since it doesn't require type annotations, nor does it require ownership of the commits themselves.

chetaldrich:
I suppose if you're committing to that then we should also close the issue, then. I'll close this.

PR #791, opened 2022-01-21
Add support for custom git extensions

knutwalker:
This would allow one to use git worktrees together with sparse-checkout.

Reference: https://github.com/libgit2/libgit2/pull/6031


knutwalker:
Thanks for the hint! I have moved the tests to the tests/ folder and rewritten them a bit. They now look like what I initially wanted them to look like, much better :)

PR #799, opened 2022-02-05
Harmonize commitish and committish in docstrings

sgasse:
The git glossary mentions both verions. However in-code, only the first
variant is used. This commit harmonizes the usages:
https://git-scm.com/docs/gitglossary#Documentation/
gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -811,7 +811,7 @@ impl Repository {
 
     /// Make the repository HEAD directly point to the commit.
     ///
-    /// If the provided committish cannot be found in the repository, the HEAD
+    /// If the provided commitish cannot be found in the repository, the HEAD
     /// is unaltered and an error is returned.
     ///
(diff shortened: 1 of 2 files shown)


src/repo.rs
```diff
@@ -831,7 +831,7 @@ impl Repository {
 
     /// Make the repository HEAD directly point to the commit.
     ///
-    /// If the provided committish cannot be found in the repository, the HEAD
+    /// If the provided commitish cannot be found in the repository, the HEAD
     /// is unaltered and an error is returned.
     /// If the provided commitish cannot be peeled into a commit, the HEAD is
```
sgasse, line 836: e.g. here we have both spellings directly one after another - so maybe it makes sense to harmonize it? Or does it not make any sense because it we might get both variants back anyway?

alexcrichton:
Seems reasonable!

PR #802, opened 2022-02-12
clarify license

dimbleby:
Per <https://doc.rust-lang.org/cargo/reference/manifest.html#slash>

> Previously multiple licenses could be separated with a /, but that usage is deprecated.

pretty sure from the README that your intention is OR rather than AND, I guess this form makes that clear

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "git2"
 version = "0.13.25"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0"
 readme = "README.md"
 keywords = ["git"]
 repository = "https://github.com/rust-lang/git2-rs"
--- a/git2-curl/Cargo.toml
+++ b/git2-curl/Cargo.toml
@@ -2,7 +2,7 @@
 name = "git2-curl"
 version = "0.14.1"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
-license = "MIT/Apache-2.0"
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the report, but `/` is used by quite a few crates so I don't think it makes sense to change it here in just this one crate. IIRC basically everything I published for a year or so used `/` and I think this may be one of the first that has had a request to change.

dimbleby:
What's the downside?  isn't the un-deprecated form just better?

We have tooling that parses licenses using [spdx](https://crates.io/crates/spdx).  That crate is happy with "MIT OR Apache-2.0" but not with "MIT/Apache-2.0": so that's my motivation.  Every now and then I go and submit a handful of MRs to projects that I've encountered using a string for which the parse fails.

dimbleby:
(in fact as it goes https://github.com/rust-lang/socket2/pull/288 looks like one where the `/` originated with you, so this isn't _quite_ the first to have had a request to change...)

atouchet:
There are a couple of other license fields in this repo with this outdated formatting as well:

https://github.com/rust-lang/git2-rs/blob/8c6a26bd74f2a16c49a8187127152ff3c3b53276/git2-curl/Cargo.toml#L5

https://github.com/rust-lang/git2-rs/blob/0443adb5388b2a529c042256e7716c36dc567220/libgit2-sys/Cargo.toml#L8

dimbleby:
Thanks.  Not sure whether the hint was that this would be more likely to be merged if I fixed all three files; but I've updated the MR so that it does that.

joshtriplett:
@dimbleby There are many, many crates out there that uses `/` instead of ` OR `. Any tool parsing SPDX license fields should likely handle `/` for compatibility.

I'm going to go ahead and merge this, but I would still recommend that whatever tool you're using needs to handle `/`.

PR #811, opened 2022-02-18
Update to 1.4.1 and bump versions

ehuss:
This updates libgit2 to 1.4.1, and bumps the Cargo versions as a SemVer breaking change.

This also includes a file with instructions on how to do a libgit2 update.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+# Contributing
+
+## Updating libgit2
+
+The following steps can be used to update libgit2:
+
+1. Update the submodule.
+   There are several ways to go about this.
+   One way is to go to the `libgit2-sys/libgit2` directory and run `git fetch origin` to download the latest updates, and then check out a specific tag (such as `git checkout v1.4.1`).
+2. Update all the references to the version:
+    * Update [`libgit2-sys/build.rs`](https://github.com/rust-lang/git2-rs/blob/master/libgit2-sys/build.rs).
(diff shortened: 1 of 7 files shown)


alexcrichton:
IIRC there's a fair bit of cruft along the lines of "remove this deprecated method when the next major version happens" or something like that. I unfortunately haven't been keeping great track of that over time so I don't exactly have a list off-hand. Would you be up for poking around and seeing how much is affected? If you're short on time then I think merging this is fine as well.

ehuss:
I can nuke all of the deprecated stuff.  There's a pretty long list in [deprecated.h](https://github.com/libgit2/libgit2/blob/main/include/git2/deprecated.h). Do you care about providing deprecated warnings?  That is:

* Remove anything already marked `#[deprecated]`
* Remove anything in libgit-sys that is deprecated in C. I personally don't feel like it is very important to provide deprecation warnings for these, since the number of active users for the raw API is probably close to zero. However, perhaps I am in a surly mood today.
* Remove anything in the git2 crate that is deprecated, even if not using `#[deprecated]`. There are a few things like the code for generating email-formatted diffs that has fundamentally changed the API.  I could maybe see adding deprecations instead of outright removal, but again I feel like the number of users is extremely small.

I can't think of any API breaking changes that are waiting in the wings so to speak.  I skimmed the open and closed PRs, but didn't see anything.

weihanglo:
I guess we need to publish a new version of git2-curl. PR is ready: https://github.com/rust-lang/git2-rs/pull/814

PR #814, opened 2022-02-26
Bump version of git2-curl

weihanglo:
It needs a SemVer breaking bump because the version of underlying git2/libgit2-sys changes.


ehuss (approved):
Thanks!  Not sure why I didn't do that.

Issue #813, opened 2022-02-25
Undefined Behavior/Segfault when using git2 < 0.14.0 with libgit2 >= 1.4.0


zRedShift:
A [new field](https://github.com/rust-lang/git2-rs/blame/64d849b8436a6b77a2636693a7523b5cac9cd993/libgit2-sys/lib.rs#L386) was added to `git_fetch_options` in v1.4.0 of libgit2, which makes [this call](https://github.com/rust-lang/git2-rs/blob/64d849b8436a6b77a2636693a7523b5cac9cd993/src/remote.rs#L303) unsound on the condition that any version of `git2` below v0.14.0 is used and `libgit2.so.1.4` is linked, which has happened already on bleeding edge distributions.

This has already happened to me and apparently a few others on killercup/cargo-edit#641, and in my investigation I found the issue to be due to a segfault in [`validate_custom_headers`](https://github.com/libgit2/libgit2/blob/v1.4.1/src/remote.c#L847), as `follow_redirects` is incorrectly "initialized" with `custom_headers`'s `git_strarray` while the actual `custom_headers` points to junk data past the end of the struct, causing `git_strarray`'s `count` to be uninitialized memory/random junk, which causes the code to dereference `strings` which also points to random junk.

I'm not entirely sure how to fix this gracefully. Maybe issue an advisory to all crates using `git2` to upgrade to 0.14.0? Wouldn't it cause the same issue in reverse to those who still use `libgit2.so.1.3`?

saethlin:
This is a bug in libgit2-sys, I tried to explain what's going on and what could be done in the issue that reported this problem when there was a ABI break with 1.3.0: https://github.com/rust-lang/git2-rs/issues/746#issuecomment-1012185316

Now that there has been another ABI break I think it's pretty clear that the version of libgit2 doesn't indicate ABI stability, and the build script should be changed to an exact version, and if possible this should be backported to past releases so there's a chance people stop getting segfaults from other crates that depend on older versions of libgit2-rs.

demurgos:
Is there a workaround available currently for this issue? I tried enabling the `vendored-libgit2` feature but still get the crash in my project.

saethlin:
If you run `ldd` on your executable, does it mention `libgit2`? If not, I would very much like to be able to compile this project for myself and check this out.

The `vendored-libgit2` feature ought to fix this.

demurgos:
I'll try to reduce the binary to a minimal reproduction.
Here is what I have:
[...]

Issue #746, opened 2021-09-17
Error running tests after upgrade to libgit2-sys 0.12.23+1.2.0


taurr:
In https://github.com/cargo-generate/cargo-generate, after upgrading `git2` to 0.13.22, tests no longer run due to

```
Git Error: invalid version 0 on git_proxy_options; class=Invalid (3)
```

Tracing this error leads to the upgrade of libgit2-sys, and seems to be an old issue: https://www.google.com/search?q=Git+Error%3A+invalid+version+0+on+git_proxy_options

alexcrichton:
Can you provide some more information about how to reproduce and how minimize perhaps? The only usage of `git_proxy_options` should be constructed with the version field set to 1, so I'm not sure where this would be coming from.

taurr:
Nor am I unfortunately, but it started failing immediately after upgrading to git2 version 0.13.21, and **only** on my linux setup (Manjoro/Arch) - not on windows 10.

After some searching and messing around I found that it ties down to the use of `libgit2-sys` version 0.12.23+1.2.0. A single revision back on that lib, and everything is ok again. Obviously it's a little difficult to downgrade once the offending version is on the system as the little `+` in the version allows it to pick the newer one :-)

taurr:
As for reproduction, I was running `cargo test` against https://github.com/cargo-generate/cargo-generate

srhb:
I did the mind-numbing bisect of libgit2 to find out what the culprit is, hoping that someone can pick up from here and realize what went wrong.

It appears the problem was introduced here:

https://github.com/libgit2/libgit2/pull/6012

Specifically in https://github.com/libgit2/libgit2/pull/6012/commits/72df17c659619707e4e5f27b2a51db60848a1d04

Doing a "slightly-sloppy-but-hey-it-builds"-revert appears to fix it: https://github.com/srhb/libgit2/commit/7d08801470ed377863c5b4b1a655b3f3ee980c22

kim:
It might be that an incompatible version of libgit2 from the system package manager is picked up. Maybe try building with the `vendored-libgit2` feature?

srhb:
You got me on the right track -- the problem was the opposite. I was pulling in a too-old version of git2-rs (and transitively the -sys component) which was incompatible with the newer libgit2 I was linking against. I had the problem backwards. Thank you! :smile:

saethlin:
[...]

PR #817, opened 2022-03-04
Respect libgit2's ABI instability

saethlin:
Notably this PR does _not_ close the mentioned issues. The problem there is that old versions of this library become broken when the system version of the library moves on. All I can suggest is that maybe old versions of this library might be better yanked or with this fix backported.

libgit2 does not have a stable ABI across minor versions, as has been
demonstrated in the last few minor releases, see #813 and #746. This
pain is primarily suffered by users of rolling release distros, because
they tend to get a new libgit2 version before the authors of libraries
release versions that pull in the new libgit2-sys version which works
with the new binary. (see also https://github.com/rust-lang/crater/pull/598)

This patch means that going forward, users don't need to rush to
upgrade their version of libgit2-sys or suffer errors/UB in the
meantime. If the system version changes, they will just start using
the vendored version which has been tested against the bindings.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -14,7 +14,7 @@ fn main() {
     let try_to_use_system_libgit2 = !vendored && !zlib_ng_compat;
     if try_to_use_system_libgit2 {
         let mut cfg = pkg_config::Config::new();
-        if let Ok(lib) = cfg.atleast_version("1.4.0").probe("libgit2") {
+        if let Ok(lib) = cfg.range_version("1.4.0".."1.5.0").probe("libgit2") {
             for include in &lib.include_paths {
                 println!("cargo:root={}", include.display());
             }


ehuss:
It's not clear to me how this will entirely fix the problem.  If a Rust tool is using the system libgit2 1.4, and then the system updates to an ABI breaking 1.5, wouldn't there still be a problem?  Or conversely, if 1.5 is already installed, then I think this will just switch to being a static build?

I would think the Rust side would need to do some kind of runtime version detection?  

Personally, I think using the system library is a bad idea.  I know others don't agree, but it generally seems like it is a recipe for problems.  I feel to some degree that shared linking should be opt-in instead of opt-out.

saethlin:
> If a Rust tool is using the system libgit2 1.4, and then the system updates to an ABI breaking 1.5, wouldn't there still be a problem?

Ah, right. There also needs to be an adjustment to what we ask the linker for. `-lgit2.so.1.4` or such. I think that's possible, I'll give it a shot in a few hours.

> Or conversely, if 1.5 is already installed, then I think this will just switch to being a static build?

Yup. Or we could produce a build-time error.

> Personally, I think using the system library is a bad idea.

I agree. It's a mistake to use components delivered by the system package manager without integrating with it. You may have a program on your system that requires libgit2 version 1.4.0 but if it isn't integrated with the package manager, the package manager has no way to know that it can't just upgrade to 1.5.0. There's no "just upgrade the shared library to upgrade all the apps" story here that people often use to argue for dynamic linkage, because git2 is shipping ABI breaks so often.

But even though we can't tell the package manager not to upgrade to a version of libgit2 that breaks the Rust program, we can at least not have UB. And also a better error. This change wouldn't completely fix the problem, but it would shave off the class of error which I and others have opened issues about. So I don't see a way that this sort of change (once I finish it, as you point out) would make things worse, but I would be in favor of removing dynamic linkage support entirely.

saethlin:
I looked into this and I'm not sure that it's possible, but I'm also not sure if it's necessary.

In short, it looks like Cargo doesn't support specifying a versioned native library. You can't pass `cargo:rustc-link-arg` because that doesn't apply to dependents. And `cargo:rustc-link-library` just forwards down to `-l`, so you can't provide the file extension through that flag. And you can't even use [this suggestion](https://stackoverflow.com/a/17294840) by saying `cargo:rustc-link-lib=:git2.so.1.4` because this invokes some piece of Cargo functionality that I'm currently unable to locate documentation for (I feel like it should be in the build script docs but I can't find it there):
```
   Compiling libgit2-sys v0.13.1+1.4.2 (/home/ben/git2-rs/libgit2-sys)
error: renaming of the library `` was specified, however this crate contains no `#[link(...)]` attributes referencing this library

error: could not compile `libgit2-sys` due to previous error
```

But also, I'm an Arch user so I've recently pulled down the latest `libgit2` and my `.cargo/bin` is in shambles:
```
╭ ➜ ben@archlinux:~/.cargo/bin
╰ ➤ ldd * | grep libgit
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
	libgit2.so.1.3 => not found
```
I don't know that we can _rely_ on this extension versioning to always kick in when linking, but this seems to be what it is for and it is doing its job. I suspect if we just prevent invalid linkage at build time, things will correctly break when the system is upgraded.

alexcrichton:
If libgit2's ABI is unstable, which it was assumed to not be with the 1.0 version and beyond, then I agree with @ehuss  that the option to dynamically link against a system-provided version should be opt-in, not opt-out.

alexcrichton:
I think that defaulting to the vendored copy probably shouldn't be done with features since it's typically pretty hard to turn off default features and this otherwise may want to be configured more regularly. Instead could an env var be invented for opting-in to using `pkg-config`? Maybe something like `GIT2RS_TRY_PKGCONFIG=1` or something like that. Build scripts often have random environment variables (in my experience at least) associated with them.

Alternatively we could have an off-by-default Cargo feature which enables pulling in `pkg-config` and actually uses it to try to find a system version.

joshtriplett:
I know that Linux distros will find it hard to work with a default of not using the system library, and will get annoyed at having to set one more environment variable or feature to use the system library.

In general, I wish we had a better unified mechanism for making this decision cargo-crate-graph-wide rather than unique mechanisms on a crate-by-crate basis.

But in the meantime, I agree that we shouldn't just set the vendoring feature by default.

saethlin:
The other alternative is that we could retain the status quo with features, but let users set `GIT2RS_TRY_PKGCONFIG=0`.

@joshtriplett They have set the SONAME correctly:
```
╭ ➜ ben@archlinux:~
╰ ➤ objdump -p /usr/lib/libgit2.so | grep SONAME
  SONAME               libgit2.so.1.4
  ```
If I understand correctly, that's why I had to reinstall tools like `cargo-edit` when I updated my system a few days ago.

**Comment by joshtriplett** [comment]
> @joshtriplett They have set the SONAME correctly:
> 
> ```
> ╭ ➜ ben@archlinux:~
> ╰ ➤ objdump -p /usr/lib/libgit2.so | grep SONAME
>   SONAME               libgit2.so.1.4
> ```
> 
> If I understand correctly, that's why I had to reinstall tools like `cargo-edit` when I updated my system a few days ago.

Oh, huh. My apologies, I misread the library on my system because I didn't realize my system copy of libgit2 was 1.1. I think it hasn't been updated in Debian in a while *because* it changes SONAME and thus requires a transition.

OK, if the SONAME *is* actually changing with each minor version, then I think it *does* suffice to just change our pkg-config search to look for the correct range of versions. Ideally, the pkg-config file itself would reflect the SONAME (e.g. `libgit2-1-4.pc`), but we can at least work around that.

I think we should just change our pkg-config search, and not change our defaults; that should be sufficient to prevent using an incompatible system library.

File: Cargo.toml [hunk 1]
```diff
@@ -34,7 +34,7 @@ paste = "1"
 
 [features]
 unstable = []
-default = ["ssh", "https", "ssh_key_from_memory"]
+default = ["ssh", "https", "ssh_key_from_memory", "vendored-libgit2"]
```
joshtriplett, line 37:
```suggestion
default = ["ssh", "https", "ssh_key_from_memory"]
```

saethlin: :+1: I dropped the commit

joshtriplett:
LGTM.

@alexcrichton @ehuss Thoughts on merging this in its current form?

alexcrichton:
I'm confused, this is still trying to use the system library by default? (but with more strict version constraints, which I thought is what this PR originally looked like and at the time I thought it would be better to not use pkg-config by default)

Personally though this has already taken more of my time than I'm willing to give, so I'll leave this up to y'all.

saethlin:
My biggest concern is that we are getting reports of projects that are built in a broken state, because the version constraint in here is _wrong_. This fixes that.

I think that defaulting to something other than pkg-config or introducing and environment variable, even if we do it, is going to stir up bikeshedding and I'd rather get a fix to users ASAP.

ethomson:
👋 Hello from https://github.com/libgit2/libgit2, where I've been seeing some reports of rust projects having a bad day with libgit2 releases. This is obviously disappointing. So I just wanted to mention that we would welcome a detailed and polite conversation about how we can ease your pain. As is noted upthread, we do try to be mindful of these issues by updating our SOVERSION when there are ABI changes, but there's certainly a lot more that we could do.

_This_ probably isn't the right place to have that chat, but feel free to drop by https://github.com/libgit2/libgit2/discussions if you'd like to discuss ABIs.

joshtriplett:
@ethomson Thank you, and sorry for the initial misunderstanding about libgit2's SONAME. (I've seen libraries that do the breakage-in-minor-versions thing *without* setting the SONAME accordingly, and I misread the state of libgit2 on my system and thought that it had done so.)

Going forward, I think we'll be able to address this breakage by having git2-rs check pkg-config more precisely for only one supported minor version number. We may also want to issue an updated 0.13 release with an upper bound on the pkg-config check.

That said, I think the Rust git2-rs bindings and any other binding that depends on a compiled binary would benefit from doing more traditional breakage-on-major-version-only library versioning, if only to match semantic versioning expectations from other libraries and make transitions easier. It's not that big a deal if libgit2 ends up having a major version number of 4 or 5.

saethlin:
I think that what this library should do going forward is a great topic for discussion, but the diff in this PR is small and I think uncontroversial? So it would be nice if it were not blocked by that discussion.

PR #822, opened 2022-03-16
Support for building refactored libgit2 sources for `vendored-libgit2` feature

bierbaum:
Since https://github.com/libgit2/libgit2/commit/3a3ab065f0685202c854e13708ddfd2a93d75e2c libgit2 is refactored into a library and CLI. This change allows git2-rs to build those sources when the  copy is used.

This should not be integrated until a libgit2 release is made incorporating those sources. The libgit2 submodule has been updated at the current commit on libgit2's `main` branch.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -41,19 +41,21 @@ fn main() {
     cp_r("libgit2/include", &include);
 
     cfg.include(&include)
-        .include("libgit2/src")
+        .include("libgit2/src/libgit2")
+        .include("libgit2/src/util")
         .out_dir(dst.join("build"))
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

ehuss:
@alexcrichton Do we want to keep libgit2-sys tied to specific tagged releases?  If so, this may make it difficult to make releases with changes until the next upstream release.

alexcrichton:
In general since we use pkg-config to find a system version I think we need to follow the upstream, but I figured that this was ok since there wasn't any ABI changes so whether or not the system version is used the `libgit2-sys` crate should still have the right bindings I think?

PR #829, opened 2022-04-03
remote: remove unnecessary mutable borrow of FetchOptions

alevinval:
Fetch and download are not mutating the FetchOptions, so the mutable
borrow is unnecessary.

This also makes using the crate confusing, because some places like
RepoBuilder accept owned versions of the options. This means users
of the library have to deal with providing both mutable and immutable
versions of their FetchOptions.

With this patch, it will no longer be the case.


alevinval:
Hello, this is my first contribution in rust, ever. I think inconsistency in the API was a small overlook, because I see from other examples that the usual pattern is mutably owning the fetch options (eg `let mut fo = FetchOptions::new()`, then the user chains as many builder calls as necessary. However, this mutable borrow slipped as one of the examples was trying to directly feed the result of the builder, which returns `&mut`. Let me know if this patch is good.

alevinval:
I see, closing then. Thanks.

PR #839, opened 2022-05-09
update libgit2 to newer snapshot

### LINKED ISSUE #795: Support for zdiff3 conflict-style from Git 2.35
Setting the new zdiff3 conflictStyle with
```
[merge]
    conflictStyle = zdiff3
```
in my global .gitignore leads to errors in e.g. cargo-deny like
```
2022-01-26 10:53:03 [ERROR] failed to fetch advisory database: git operation failed: unknown style 'zdiff3' given for 'merge.conflictstyle'; class=Checkout (20)
```

Since their source-code does not mention diff3 (the old value of that setting) but it uses git2-rs as a dependency and your source code does I assume it needs to be fixed on your end.

This is particularly annoying since that also means I can't use it with git itself.

davidkna:
[Fixes CVE 2022-24765](https://github.com/libgit2/libgit2/releases/tag/v1.4.3)

@justinsteven has a [writeup of this issue](https://github.com/justinsteven/advisories/blob/main/2022_git_buried_bare_repos_and_fsmonitor_various_abuses.md?rgh-link-date=2022-05-09T04%3A16%3A51Z) (see also starship/starship#3974)

I think this also fixes #795.

I needed to make some changes to the libgit2-sys `build.rs` to get this to build and the owner validation needed to be disabled for  the tests on Windows, I've not investigated the cause in detail.

Furthermore, I've increased the minor version and treated this as a breaking change, since this change could prevent opening a repository if it is not owned by the current user.

I am disclosing this security issue publicly because this class of vulnerability has already been reported publicly in several other similar projects.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git2"
-version = "0.14.4"
+version = "0.15.0"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT OR Apache-2.0"
 readme = "README.md"
@@ -20,7 +20,7 @@ url = "2.0"
 bitflags = "1.1.0"
 libc = "0.2"
 log = "0.4.8"
-libgit2-sys = { path = "libgit2-sys", version = "0.13.4" }
+libgit2-sys = { path = "libgit2-sys", version = "0.14.0" }
 
 [target."cfg(all(unix, not(target_os = \"macos\")))".dependencies]
 openssl-sys = { version = "0.9.0", optional = true }
--- a/git2-curl/Cargo.toml
+++ b/git2-curl/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2018"
 curl = "0.4.33"
 url = "2.0"
 log = "0.4"
-git2 = { path = "..", version = "0.14", default-features = false }
+git2 = { path = "..", version = "0.15", default-features = false }
 
 [dev-dependencies]
 civet = "0.11"
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libgit2-sys"
-version = "0.13.4+1.4.2"
+version = "0.14.0+1.4.4"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 links = "git2"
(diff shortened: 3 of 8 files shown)


justinsteven:
> @justinsteven has a [writeup of this issue](https://github.com/justinsteven/advisories/blob/main/2022_git_buried_bare_repos_and_fsmonitor_various_abuses.md?rgh-link-date=2022-05-09T04%3A16%3A51Z) (see also https://github.com/starship/starship/issues/3974)

For absolute clarity, the writeup linked to does not discuss CVE-2022-24765. The writeup is regarding similar but separate problems.

ehuss:
Thanks for the PR!  I should have opened an issue about this, as we have been working on it.

There are some issues here:

* It looks like this updated to the branch.  However, we have moved on to the `main` branch in #822.  This reverts all those changes, so I think it should stay on main.
* I'd like to delay updating until https://github.com/libgit2/libgit2/issues/6279 is fixed.
* GIT_EOWNER should be added
* There are some tricky sha changes added recently, I had a partial fix at: https://github.com/ehuss/git2-rs/commit/54435556dce47db19036287699148744b9cf366d
* It looks like you defined GIT_OPT_GET_OWNER_VALIDATION twice

davidkna:
@ehuss So should I rather close this for now, if you're already working on this? It looks like https://github.com/libgit2/libgit2/issues/6279 is the cause of the ownership issue in the tests on Windows I fixed.

> It looks like you defined GIT_OPT_GET_OWNER_VALIDATION twice.

It has GET and SET variants.

ehuss:
No, it's fine to keep this open, if you're willing to make the updates mentioned above.

davidkna:
Okay, I am fine with updating this a bit more. What exactly is missing in https://github.com/ehuss/git2-rs/commit/54435556dce47db19036287699148744b9cf366d? It seems to mostly match what's in the libgit2 cmake files. Having to include sha256-support, despite it not having a stable release, is a bit unfortunate.

ehuss:
I think the only things missing on my branch were GIT_EOWNER and GIT_OPT_GET_OWNER_VALIDATION and GIT_OPT_SET_OWNER_VALIDATION.

davidkna:
@ehuss [libgit2 v1.4.4](https://github.com/libgit2/libgit2/releases/tag/v1.4.4) was just released, which is supposed to fix the Windows behavior. Updated to latest `main` libgit2.

ehuss (approved):
Thanks! Hopefully the fallout from this won't be too severe.

PR #841, opened 2022-05-10
Add `Tree::get_name_bytes` to handle non-UTF-8 entry names

joshtriplett:


alexcrichton:
Could `get_name` be reimplemented in terms of this method?

joshtriplett:
@alexcrichton Good idea, will do.

PR #843, opened 2022-05-14
chore(deps): bump openssl-sys to 0.9.73

pkarakal:
This bumps the openssl-sys crate version to the latest version (v0.9.73). Since 0.9.0, which was released a while ago, openssl 3.0 support has been added to that crate. Also Fedora 36 ships openssl 3.0.2 , so building the git crate breaks.

Edit: Same thing happens in Ubuntu 22.04 which also has openssl 3.0.2)


alexcrichton:
Thanks for the PR but this isn't necessary. The dependency here on 0.9.0 is actually "anything semver compatible with 0.9.0 that's at least 0.9.0", so there's no need to upgrade this dependency here. Instead you should do so locally in your `Cargo.lock`.

PR #847, opened 2022-06-01
Add bindings for git email create

f0rget-the-sad:
Add bindings for `git_email_create_from_diff()` and
`git_email_create_from_commit()`. Deprecate `git_diff_format_email()` to
reflect upstream changes.

git_diff_format_email was refactored in [upsrtream](https://github.com/libgit2/libgit2/pull/6061).

It's also possible to impl git_email_create_from_diff/commit directly for Diff/Commit structs, and do not introduce `Email` struct, but I think current approach is better for future extension(e.g. there is plan to intorduce `git_email_create_from_commits` that produces a series of emails).

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -1992,6 +1992,28 @@ pub struct git_message_trailer_array {
     pub _trailer_block: *mut c_char,
 }
 
+#[repr(C)]
+pub struct git_email_create_options {
+    pub version: c_uint,
+    pub flags: u32,
+    pub diff_opts: git_diff_options,
+    pub diff_find_opts: git_diff_find_options,
+    pub subject_prefix: *const c_char,
+    pub start_number: usize,
+    pub reroll_number: usize,
+}
+
+pub const GIT_EMAIL_CREATE_OPTIONS_VERSION: c_uint = 1;
+
+git_enum! {
+    pub enum git_email_create_flags_t {
+        GIT_EMAIL_CREATE_DEFAULT = 0,
+        GIT_EMAIL_CREATE_OMIT_NUMBERS = 1 << 0,
+        GIT_EMAIL_CREATE_ALWAYS_NUMBER = 1 << 1,
+        GIT_EMAIL_CREATE_NO_RENAMES = 1 << 2,
+    }
+}
+
 extern "C" {
     // threads
     pub fn git_libgit2_init() -> c_int;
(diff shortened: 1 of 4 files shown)


src/diff.rs
```diff
@@ -254,6 +254,8 @@ impl<'repo> Diff<'repo> {
     /// Create an e-mail ready patch from a diff.
     ///
     /// Matches the format created by `git format-patch`
+    #[doc(hidden)]
+    #[deprecated(note = "refactored to `email_from_diff` to match upstream")]
```
alexcrichton, line 258: I think this naming should be `Email::from_diff` perhaps?

libgit2-sys/lib.rs
```diff
@@ -3559,6 +3581,7 @@ extern "C" {
         ancestor: *const git_oid,
     ) -> c_int;
 
+    #[deprecated(note = "refactored to `email_from_diff` to match upstream")]
```
alexcrichton, line 3584: I think it's ok to skip the deprecation here since that's more for the API layer rather than the sys layer

src/email.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn default() -> Self {
+        // Defaults options created in corresponding to `GIT_EMAIL_CREATE_OPTIONS_INIT`
+        let mut diff_opts = DiffOptions::new();
+        diff_opts.show_binary(true).context_lines(3);
+        unsafe {
+            let default_options = raw::git_email_create_options {
+                version: raw::GIT_EMAIL_CREATE_OPTIONS_VERSION,
+                flags: raw::GIT_EMAIL_CREATE_DEFAULT as u32,
+                diff_opts: ptr::read(diff_opts.raw()),
+                diff_find_opts: ptr::read(DiffFindOptions::new().raw()),
```
alexcrichton, line 30: Instead of using unsafe code like `ptr::read` here I think this builder should probably embedd the builders for the other options? That way all the options can continue to be configurable

src/email.rs
```diff
@@ -0,0 +1,173 @@
...
+        let buf = Buf::new();
+        unsafe {
+            try_call!(raw::git_email_create_from_diff(
+                buf.raw(),
+                Binding::raw(diff),
+                patch_idx,
+                patch_count,
+                Binding::raw(commit_id),
+                summary.into_c_string()?.as_ptr(),
+                body.into_c_string()?.as_ptr(),
```
alexcrichton, line 152: While I think this may technically work I think it would be better to extract this to a local variable above to explicitly prevent the owned `CString` from being dropped too soon.

f0rget-the-sad: Done.

src/email.rs
f0rget-the-sad: makes sense, changed.

src/email.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::{mem, ptr};
+
+use crate::util::Binding;
+use crate::{raw, Buf, Commit, DiffFindOptions, DiffOptions, Error, IntoCString};
+use crate::{Diff, Oid, Signature};
+
+/// A structure to represent patch in mbox format for sending via email
+pub struct Email {
+    /// Buffer to store the e-mail patch in
+    pub buf: Buf,
```
alexcrichton, line 11: Instead of making this a `pub` field could public accessors be provided for this? (sorry missed this earlier)

Ideally byte slices would be returned as well instead of `&Buf` too

src/email.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Options for controlling the formatting of the generated e-mail.
+pub struct EmailCreateOptions {
+    /// Options to use when creating diffs
+    pub diff_options: DiffOptions,
+    /// Options for finding similarities within diffs
+    pub diff_find_options: DiffFindOptions,
```
alexcrichton, line 19: Instead of making these `pub` could method accessors be provided?

alexcrichton (approved):
Thanks! Two more small things but otherwise seems good to go to me.

src/email.rs
f0rget-the-sad: Done.

src/email.rs
f0rget-the-sad: Done.

PR #858, opened 2022-07-14
bump libgit2 to 1.5.0

davidkna:
Fixup for #839
libgit2 1.5.0 was released shortly after it was merged, so switch from `main` to the stable `v1.5.0` tag and amend `cfg.range_version` to include v1.5.0.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -14,7 +14,7 @@ fn main() {
     let try_to_use_system_libgit2 = !vendored && !zlib_ng_compat;
     if try_to_use_system_libgit2 {
         let mut cfg = pkg_config::Config::new();
-        if let Ok(lib) = cfg.range_version("1.4.4".."1.5.0").probe("libgit2") {
+        if let Ok(lib) = cfg.range_version("1.4.4".."1.6.0").probe("libgit2") {
             for include in &lib.include_paths {
                 println!("cargo:root={}", include.display());
             }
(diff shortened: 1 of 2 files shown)


libgit2-sys/build.rs
```diff
@@ -14,7 +14,7 @@ fn main() {
     let try_to_use_system_libgit2 = !vendored && !zlib_ng_compat;
     if try_to_use_system_libgit2 {
         let mut cfg = pkg_config::Config::new();
-        if let Ok(lib) = cfg.range_version("1.4.4".."1.5.0").probe("libgit2") {
```
extrawurst, line 17: why do we need to support `1.6.0` here already?

extrawurst: good point. idk

weihanglo: Got to be careful when upgrading libgit2. It doesn't guarantee stable ABI between minor versions.

- https://github.com/rust-lang/git2-rs/pull/817
- https://github.com/libgit2/libgit2/blob/fbea439d4b6fc91c6b619d01b85ab3b7746e4c19/docs/api-stability.md?plain=1#L44-L57

davidkna: @weihanglo The in-repo libgit2 submodule uses a revision from the libgit2 main branch that almost matches libgit2 v1.5.0.

ehuss (approved):
Thanks, looks good to me. AFAICT, compatibility with 1.4.4 should be fine, I don't see any ABI changes.

PR #862, opened 2022-07-21
On GLIBC, compile libgit2 with `-D_FILE_OFFSET_BITS=64`.

sunfishcode:
On GLIBC, compile libgit2 with `-D_FILE_OFFSET_BITS=64`. On 32-bit
architectures, this is needed for libgit2 to access files larger than 2 GiB.

While here, also use `-D_TIME_BITS=64`, which is needed on GLIBC on 32-bit
architectures for libgit2 to work correctly in in the year 2038 and beyond.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -112,6 +112,12 @@ fn main() {
         cfg.define("__EXTENSIONS__", None);
     }
 
+    // Support large files and dates after the year 2038 on 32-bit GLIBC.
+    if target.contains("gnu") {
+        cfg.define("_FILE_OFFSET_BITS", Some("64"));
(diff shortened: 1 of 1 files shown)


joshtriplett:
@sunfishcode To what extent does this change the public ABI of the libgit2 C library? If people are calling functions in libgit2-sys directly, will any of that code break?

sunfishcode:
libgit2 uses [its own typedefs](https://github.com/libgit2/libgit2/blob/main/include/git2/types.h#L56) for file offsets and times in its public API, specifically to avoid being affected by this issue.

sunfishcode:
Ah, looking back on this, I see that it's not just about the public API. In that case, I don't know how to do this transition in a reasonable amount of effort.

PR #868, opened 2022-08-12
Add reminder to update README version number

Piturnah:
In [v0.15](https://github.com/rust-lang/git2-rs/tree/0.15.0), the project readme says

```toml
[dependencies]
git2 = "0.14"
```

Until it was later corrected in 0b340ffcdb21a7959f81c8ce77015723f5fc6037. 

This PR proposes adding a reminder to the Cargo.toml. I understand, however, if you consider this change to be clutter! Just a suggestion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git2"
-version = "0.15.0"
+version = "0.15.0" # update version in README.md!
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks for the PR! There is a checklist in https://github.com/rust-lang/git2-rs/blob/master/CONTRIBUTING.md for the release process, and I think I'd prefer to just keep this information there. There are quite a few other steps that need to be taken, and they all need to be followed. Hopefully will try to keep it managed better in the future, and maybe some day automated.

PR #869, opened 2022-08-13
Add support for SSH interactive authentication

exactly-one-kas:
For lifetime reasons, `Cred` now doesn't always carry a `git_cred` and therefore `Cred::unwrap()` may now panic if called on a `Cred` created using `Cred::ssh_interactive()`; this is the only backwards-incompatible change

Without leaking memory or memorizing created handlers in `RemoteCallbacks` it's not possible to have individual handler for each SSH interactive `Cred`, so I compromized and put the handler directly into `RemoteCallbacks`

Also, since `libgit2` only ships with dummy `typedef`s for `LIBSSH2_USERAUTH_KBDINT_PROMPT` and `LIBSSH2_USERAUTH_KBDINT_PROMPT`, `libssh2-sys` is now a direct dependency for the `ssh` feature

### DIFF
--- a/src/remote_callbacks.rs
+++ b/src/remote_callbacks.rs
@@ -1,4 +1,4 @@
-use libc::{c_char, c_int, c_uint, c_void, size_t};
+use libc::{c_char, c_int, c_uint, c_void, malloc, size_t};
 use std::ffi::{CStr, CString};
 use std::mem;
 use std::ptr;
@@ -6,6 +6,7 @@ use std::slice;
 use std::str;
 
 use crate::cert::Cert;
+use crate::cred::{CredInner, SshInteractivePrompt};
 use crate::util::Binding;
 use crate::{
     panic, raw, Cred, CredentialType, Error, IndexerProgress, Oid, PackBuilderStage, Progress,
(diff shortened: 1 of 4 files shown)


src/remote_callbacks.rs
```diff
@@ -450,3 +503,63 @@ extern "C" fn pack_progress_cb(
...
+        }
+
+        for i in 0..num_prompts {
+            let response = &mut *responses.offset(i as isize);
+            let response_bytes = wrapped_responses[i as usize].as_bytes();
+
+            // libgit2 frees returned strings
+            let text = malloc(response_bytes.len());
+            response.text = text as *mut c_char;
+            response.length = response_bytes.len() as u32;
```
ehuss, line 562: There seems to be an issue here. The response is never copied into the buffer created via `malloc`. 

I think this also needs a check for a null return from `malloc`.

I'm only partially confident this is the correct way to handle this. From what I can tell, ssh2 has its own allocation routines that are part of the session, but I don't think the rust side exposes that, and I think it always defaults to libc malloc/free, so I don't think it is a problem. However, I'm a bit nervous about this manual memory handling with APIs that don't really document the behavior.

src/remote_callbacks.rs
```diff
@@ -450,3 +503,63 @@ extern "C" fn pack_progress_cb(
...
+                &wrapped_prompts[..],
+                &mut wrapped_responses[..],
+            );
+        }
+
+        for i in 0..num_prompts {
+            let response = &mut *responses.offset(i as isize);
+            let response_bytes = wrapped_responses[i as usize].as_bytes();
+
+            // libgit2 frees returned strings
```
ehuss, line 559: This is the ssh2 library that is freeing the memory, not git2 right?

ehuss (commented):
Sorry for the long delay on the review.

I just wanted to confirm, the ssh2 library does not provide any way to indicate an error in the callback?  Is the intended way to handle an error is to print a message and just not fill in the responses?

Issue #872, opened 2022-08-30
Update libssh


darakeon:
Recently I opened an issue at libgit2 asking to upgrade libssh version, because after github updated their polices I could not fetch my repo anymore. But they pointed the the libraries that use libssh bring their own libssh.

I thought about opening a pull request, but got lost trying to upgrade it. Could not find a way to do this.

Original issue in libgit2: https://github.com/libgit2/libgit2/issues/6381

## Reproduction steps

Create a SSH key to communicate with github.

```bash
ssh-keygen -t ecdsa -C "your-email"
```

Create a rust program with cargo that tries to pull a repository using fetch. Use the callback as:

```rust
fn git_credentials_callback(
	_url: &str,
	username: Option<&str>,
	_cred_type: CredentialType,
) -> Result<Cred, Error> {
	let public = get_key_path(Some("pub"));
	let private = get_key_path(None);

	let password = env::var("GIT_PASSWORD")
		.expect("No GIT_PASSWORD environment variable found"); // or some hardcoded password

	println!("{:?}", _cred_type);

	let cred = Cred::ssh_key(
		username.unwrap(),
		Some(public.as_path()),
		private.as_path(),
		Some(&password),
	);

	let result = cred.unwrap();
	println!("{:?}", result.credtype());

	return Ok(result);
}
```

The fetch:

```rust
pub fn update_local() {
	let repo = Repository::open("..").unwrap();
	let mut remote = repo.find_remote("origin").unwrap();

	let mut callbacks = RemoteCallbacks::new();
	callbacks.credentials(&git_credentials_callback);

	let mut options = FetchOptions::new();
	options.remote_callbacks(callbacks);
	options.prune(FetchPrune::On);

	remote.fetch(
		&[] as &[&str],
		Some(&mut options),
		Some("updating local")
	).unwrap();
}
```

The get_key_path of mine:

```rust
fn get_key_path(extension: Option<&str>) -> PathBuf {
	let mut path = home_dir().unwrap();
	path.push(".ssh");
	path.push("cargo");

	if let Some(extension_value) = extension {
		path.set_extension(extension_value);
	}

	return path;
}
```

The entire code: https://github.com/darakeon/dfm/blob/main/version/src/git.rs

## Expected behavior

Pull succeeds.

## Actual behavior

Call the credentials function a bunch of times and gives the error:
```
 { code: -16, klass: 23, message: "Failed to retrieve list of SSH authentication methods: Failed getting response" }
```

## Version of git2

[...]

PR #876, opened 2022-09-01
Fix Windows SSH authentication issues

WSSDude:
As was commented in issue #872 in [comment](https://github.com/rust-lang/git2-rs/issues/872#issuecomment-1234024308), libssh2 on Windows uses WinCNG backend by default.

Its implementation is old and does not support modern key formats, making libgit2 (and by extension git2-rs bindings) unusable on Windows when one wishes to use SSH authentication, as only formats that won't cause infinite loop when trying to authenticate or fail are the ones that GitHub and others don't support anymore.

Utilizing openssl-on-win32 ssh2-rs feature allows for proper use of SSH authentication method on Windows builds without infinite loops and fails inside git2-rs.
This PR does just that - enables this feature.

I don't know if this is the best approach for this (I'm still a bit new to Rust), but doing this solved all SSH auth issues I had (and which probably cannot be solved otherwise on Windows OS without libssh2 updating WinCNG backend...)

### DIFF
--- a/libgit2-sys/Cargo.toml
+++ b/libgit2-sys/Cargo.toml
@@ -22,7 +22,7 @@ path = "lib.rs"
 
 [dependencies]
 libc = "0.2"
-libssh2-sys = { version = "0.2.19", optional = true }
+libssh2-sys = { version = "0.2.19", optional = true, features = ["openssl-on-win32"] }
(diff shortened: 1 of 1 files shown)


WSSDude420:
Think CI may also need an update, looking at the output... Dont know what exactly went wrong, OpenSSL for the test seems misconfigured having quick glance at the output.

PR #882, opened 2022-09-13
Add bindings for git_reference_name_is_valid, git_remote_name_is_vali…

sanjayts:
This fixes https://github.com/rust-lang/git2-rs/issues/714

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2247,6 +2247,7 @@ extern "C" {
     ) -> c_int;
     pub fn git_remote_get_refspec(remote: *const git_remote, n: size_t) -> *const git_refspec;
     pub fn git_remote_is_valid_name(remote_name: *const c_char) -> c_int;
+    pub fn git_remote_name_is_valid(valid: *mut c_int, remote_name: *const c_char) -> c_int;
     pub fn git_remote_list(out: *mut git_strarray, repo: *mut git_repository) -> c_int;
     pub fn git_remote_rename(
         problems: *mut git_strarray,
@@ -2398,6 +2399,7 @@ extern "C" {
     pub fn git_reference_is_remote(r: *const git_reference) -> c_int;
     pub fn git_reference_is_tag(r: *const git_reference) -> c_int;
     pub fn git_reference_is_valid_name(name: *const c_char) -> c_int;
+    pub fn git_reference_name_is_valid(valid: *mut c_int, refname: *const c_char) -> c_int;
     pub fn git_reference_lookup(
         out: *mut *mut git_reference,
         repo: *mut git_repository,
@@ -3209,6 +3211,7 @@ extern "C" {
     pub fn git_tag_target(target_out: *mut *mut git_object, tag: *const git_tag) -> c_int;
     pub fn git_tag_target_id(tag: *const git_tag) -> *const git_oid;
     pub fn git_tag_target_type(tag: *const git_tag) -> git_object_t;
+    pub fn git_tag_name_is_valid(valid: *mut c_int, tag_name: *const c_char) -> c_int;
 
     // checkout
     pub fn git_checkout_head(repo: *mut git_repository, opts: *const git_checkout_options)
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -8,8 +8,8 @@ use std::str;
 use crate::object::CastOrPanic;
 use crate::util::{c_cmp_to_ordering, Binding};
 use crate::{
-    raw, Blob, Commit, Error, Object, ObjectType, Oid, ReferenceFormat, ReferenceType, Repository,
-    Tag, Tree,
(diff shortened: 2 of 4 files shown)


extrawurst:
lgtm

sanjayts:
@extrawurst Great, can this please be merged?

src/reference.rs
```diff
@@ -65,6 +65,20 @@ impl<'repo> Reference<'repo> {
...
+        crate::init();
+        let refname = CString::new(refname)?;
+        let mut valid: libc::c_int = 0;
+        unsafe {
+            try_call!(raw::git_reference_name_is_valid(
+                &mut valid,
+                refname.as_ptr()
+            ));
+        }
+        Ok(valid == 1)
```
ehuss, line 79: Since this is identical to `git_reference_is_valid_name`, instead of adding a new method, can you just update the `is_valid_name` to call the new `git_reference_name_is_valid` function?

Otherwise, it seems quite confusing to have two nearly identically named functions which do the exact same thing.

.gitignore
```diff
@@ -1,3 +1,4 @@
 target
 Cargo.lock
 src/main.rs
+.idea
```
ehuss, line 4: Can this change be removed? Tracking the ignore settings for various tools would be an endless collection that I don't think each project should be tracking. If possible, I recommend putting this in a global ignore, or use local settings.

src/remote.rs
```diff
@@ -95,6 +95,20 @@ impl<'repo> Remote<'repo> {
         unsafe { raw::git_remote_is_valid_name(remote_name.as_ptr()) == 1 }
     }
 
+    /// Ensure the remote name is well-formed.
+    pub fn name_is_valid(remote_name: &str) -> Result<bool, Error> {
```
ehuss, line 99: Same comment about reusing the existing method to call the new API.

.gitignore
sanjayts: Fair enough, I'll revert this change.

src/reference.rs
sanjayts: @ehuss You bring up a good point, though one concern I have here is deciding between:
* Silently gobbling up error and panicking as done in the existing implementing by calling unwrap (`let refname = CString::new(refname).unwrap();`)
* Updating the existing contract (and make a breaking change) by returning a Result as opposed to a scalar bool

WDYT?

ehuss: Hm. I believe the only way `git_reference_name_is_valid` can return an error is on OOM, right? I would be inclined to just panic if `git_reference_name_is_valid` fails. I probably wouldn't bother with changing the existing behavior with `CString::new`, but if I were starting from scratch, I would also change that to return `false` since internal nuls aren't really valid.

I think perhaps in a future semver breaking release, the method could be changed to return a `Result`, but I probably wouldn't bother.

sanjayts: @ehuss I've made the changes -- I've also added comments to the lib.rs file to ensure there is no confusion around why we don't use a given binding.

sanjayts: @ehuss Thoughts?

sanjayts:
Just wanted to check in to see if this can be merged or should I close this PR?

.gitignore
```diff
@@ -1,3 +1,3 @@
 target
 Cargo.lock
-src/main.rs
+src/main.rs
```
ehuss, line 3: This looks like a stray edit?

libgit2-sys/lib.rs
```diff
@@ -2246,7 +2246,14 @@ extern "C" {
         remote: *const git_remote,
     ) -> c_int;
     pub fn git_remote_get_refspec(remote: *const git_remote, n: size_t) -> *const git_refspec;
+
+    // This function is no longer used by kept here for the sake of completeness and to ensure
```
ehuss, line 2250: Not sure about the wording here?

```suggestion
    // This function is no longer used but is kept here for the sake of completeness and to ensure
```

Also, I'm not really sure this comment is necessary. There are quite a few other functions that I believe are no longer used. We don't do any sort of auditing of what is missing or not.

src/reference.rs
ehuss: I'm a bit confused by the current PR since it seems to still have two methods doing the same thing.  Would it be possible to keep just `is_valid_name` with its signature and panic if `git_reference_name_is_valid` returns an error? I'm not sure the error return is particularly useful here, and I'd like to avoid complicating things.

libgit2-sys/lib.rs
sanjayts: @ehuss I have removed the comments and actioned the changes requested below. Please take a look.

I still think there is scope for improvement of the API by propagating errors but I'm assuming that can only be done with a major release. LMK if you want me to create a ticket for this?

ehuss (approved):
Thanks!

PR #883, opened 2022-09-21
Add implementation and test for discover_path

mikayla-maki:
Hi y'all,

We're looking to use this library to build the git integration in our code editor (zed.dev). In order to do so, we wanted to cache the Repository::open_ext() call based on the repository path. But git2-rs's implementation of `Repository::discover()` automatically opens the repository for us, making it impossible to cache :(

I've added a `Repository::discover_path()` function that more closely matches the libgit2 `git_repository_discover()` call. I had to add an extra allocation to convert the git buffer into a `PathBuf` and would appreciate feedback on if there's a way to avoid the allocation.

I also added 2 tests, one for a successful discovery and one for an unsuccessful one that uses `ceiling_dirs`.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -3,7 +3,7 @@ use std::env;
 use std::ffi::{CStr, CString, OsStr};
 use std::iter::IntoIterator;
 use std::mem;
-use std::path::Path;
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #886, opened 2022-10-01
note that a commit is not a descendant of itself

RalfJung:
See upstream change https://github.com/libgit2/libgit2/pull/4362; this seems just as relevant to readers of the Rust documentation.


ehuss (approved):
Thanks!

PR #887, opened 2022-10-02
Add support for GIT_FILEMODE_BLOB_GROUP_WRITABLE

malor:
This type no longer exists in modern git, but there are existing git repositories out there with tree entries of this type (e.g. Linux kernel). Currently, traversing commits of a repository like that leads to a panic in DiffFile::mode().


ehuss (approved):
Thanks!

This is a API-breaking change since it adds to an enum without non_exhaustive. However, since we are bumping the version it should be ok to land now.

PR #890, opened 2022-10-28
PartialEq, Eq for Signature

kim:
Necessarily hand-derived structural equality, as the raw binding stores
pointers for the string fields.


ehuss (approved):
Thanks!

PR #893, opened 2022-11-12
Add `Reference::symbolic_set_target`

### LINKED ISSUE #852: Missing bindings for `git_reference_symbolic_set_target`
I couldn't find bindings to the `git_reference_symbolic_set_target` function.

I am expecting a method like `git2::Reference::set_symbolic_target` that does the same as `git2::Reference::set_target` but takes a `&str` corresponding to the new target reference, just like `git_reference_symbolic_set_target` in `libgit2`.

nasso:
Fixes #852 

I've also modified the documentation of `Repository::reference_symbolic` because the original docs for `git_reference_symbolic_set_target` made a reference to `git_reference_symbolic_create`.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -2437,6 +2437,12 @@ extern "C" {
         id: *const git_oid,
         log_message: *const c_char,
     ) -> c_int;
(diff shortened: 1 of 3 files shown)


ehuss (approved):
Thanks!

PR #896, opened 2022-11-12
Correct/relax lifetime requirements on getter methods of Commit and Signature

### LINKED ISSUE #895: Wrong elided lifetimes for Commit and Signature
This issue is very similar to [#299](https://github.com/rust-lang/git2-rs/issues/299), which was resolved in this PR: https://github.com/rust-lang/git2-rs/pull/302

In addition to filing this issue, I'm also submitting a new PR to address it.

# Motivating example

This code should compile

```
struct Wrapper<'repo> {
    inner: git2::Commit<'repo>,
}

impl<'repo> Wrapper<'repo> {
    fn email(&self) -> Option<&'repo str> {
        let signature = self.inner.author();
        let email = signature.email();
        email
    }
}
```

but fails with

```
error: lifetime may not live long enough
 --> src/lib.rs:9:9
  |
5 | impl<'repo> Wrapper<'repo> {
  |      ----- lifetime `'repo` defined here
6 |     fn email(&self) -> Option<&'repo str> {
  |              - let's call the lifetime of this reference `'1`
...
9 |         email
  |         ^^^^^ associated function was supposed to return data with lifetime `'repo` but it is returning data with lifetime `'1`

error[E0515]: cannot return value referencing local variable `signature`
 --> src/lib.rs:9:9
  |
8 |         let email = signature.email();
  |                     ----------------- `signature` is borrowed here
9 |         email
  |         ^^^^^ returns a value referencing data owned by the current function

For more information about this error, try `rustc --explain E0515`.
```

# Underlying issue

On the `Commit` struct, there are several "getter" methods where the lifetime of the return object is tied to a borrow of `Self`:
* `pub fn author(&self) -> Signature<'_> {`
* `pub fn body(&self) -> Option<&str> {`
* Etc.

In other words, written without elision, these signatures all look like
* `pub fn author<'a>(&'a self) -> Signature<'a> {` 
* `pub fn body<'a>(&'a self) -> Option<&'a str> {`
* Etc.

But the pointers that are returned by these methods all live as long as the `'repo` lifetime of the `Commit` struct itself if you look at the C code. In other words, they should all have these (more flexible) signatures:

```rust
struct Commit<'repo> { ... }

impl<'repo> Commit<'repo> {
    pub fn author(&self) -> Signature<'repo> { .. } 
    pub fn body(&self) -> Option<&'repo str> { .. }
    // Etc.
}
```

There is a very similar problem happening for the "getter" methods of the `Signature` struct.

# The C code

The equivalent "getter" methods are defined in the underlying C code is here: 
* https://github.com/libgit2/libgit2/blob/a3841af5eecc6301e87f8302c7fdce6555e39247/src/libgit2/commit.c#L521

The pointers they return live here:
* https://github.com/libgit2/libgit2/blob/main/src/libgit2/commit.h#L19

I believe all these pointers live as long as this raw `*mut raw::git_commit` pointer here:
* https://github.com/rust-lang/git2-rs/blob/master/src/commit.rs#L15

which has the `'repo` lifetime.

bcmyers:
This PR fixes #895

It is very similar to https://github.com/rust-lang/git2-rs/pull/302, which addressed a similar issue: https://github.com/rust-lang/git2-rs/issues/299

# Overview

On the `Commit` struct, there are several "getter" methods where the lifetime of the return object is tied to a borrow of `Self`:
* `pub fn author(&self) -> Signature<'_> {`
* `pub fn body(&self) -> Option<&str> {`
* Etc.

In other words, written without elision, these signatures all look like
* `pub fn author<'a>(&'a self) -> Signature<'a> {` 
* `pub fn body<'a>(&'a self) -> Option<&'a str> {`
* Etc.

But the pointers that are returned by these methods all live as long as the `'repo` lifetime if you look at the C code. In other words, they should all have these (more flexible) signatures:

```rust
struct Commit<'repo> { ... }

impl<'repo> Commit<'repo> {
    pub fn author(&self) -> Signature<'repo> { .. } 
    pub fn body(&self) -> Option<&'repo str> { .. }
    // Etc.
}
```

There is a very similar problem happening for the "getter" methods of the `Signature` struct.

This PR fixes these methods so that they return references with the correct (more flexible) lifetimes.

# Motivating example

This code should compile

```
struct Wrapper<'repo> {
    inner: git2::Commit<'repo>,
}

impl<'repo> Wrapper<'repo> {
    fn email(&self) -> Option<&'repo str> {
        let signature = self.inner.author();
        let email = signature.email();
        email
    }
}
```

but fails with

```
error: lifetime may not live long enough
 --> src/lib.rs:9:9
  |
5 | impl<'repo> Wrapper<'repo> {
  |      ----- lifetime `'repo` defined here
6 |     fn email(&self) -> Option<&'repo str> {
  |              - let's call the lifetime of this reference `'1`
...
9 |         email
  |         ^^^^^ associated function was supposed to return data with lifetime `'repo` but it is returning data with lifetime `'1`

error[E0515]: cannot return value referencing local variable `signature`
 --> src/lib.rs:9:9
  |
8 |         let email = signature.email();
  |                     ----------------- `signature` is borrowed here
9 |         email
  |         ^^^^^ returns a value referencing data owned by the current function

For more information about this error, try `rustc --explain E0515`.
```

# The C code

The equivalent "getter" methods are defined in the underlying C code is here: 
* https://github.com/libgit2/libgit2/blob/a3841af5eecc6301e87f8302c7fdce6555e39247/src/libgit2/commit.c#L521

The pointers they return live here:
* https://github.com/libgit2/libgit2/blob/main/src/libgit2/commit.h#L19

I believe all these pointers live as long as this raw `*mut raw::git_commit` pointer here:
* https://github.com/rust-lang/git2-rs/blob/master/src/commit.rs#L15

which has the `'repo` lifetime.

### DIFF
--- a/src/commit.rs
+++ b/src/commit.rs
@@ -65,7 +65,7 @@ impl<'repo> Commit<'repo> {
     /// potential leading newlines.
     ///
     /// `None` will be returned if the message is not valid utf-8
-    pub fn message(&self) -> Option<&str> {
+    pub fn message(&self) -> Option<&'repo str> {
         str::from_utf8(self.message_bytes()).ok()
     }
 
@@ -73,35 +73,36 @@ impl<'repo> Commit<'repo> {
     ///
     /// The returned message will be slightly prettified by removing any
     /// potential leading newlines.
-    pub fn message_bytes(&self) -> &[u8] {
-        unsafe { crate::opt_bytes(self, raw::git_commit_message(&*self.raw)).unwrap() }
+    pub fn message_bytes(&self) -> &'repo [u8] {
+        unsafe { crate::opt_bytes_very_unsafe(raw::git_commit_message(&*self.raw)).unwrap() }
     }
 
     /// Get the encoding for the message of a commit, as a string representing a
     /// standard encoding name.
     ///
     /// `None` will be returned if the encoding is not known
-    pub fn message_encoding(&self) -> Option<&str> {
-        let bytes = unsafe { crate::opt_bytes(self, raw::git_commit_message_encoding(&*self.raw)) };
+    pub fn message_encoding(&self) -> Option<&'repo str> {
+        let bytes =
+            unsafe { crate::opt_bytes_very_unsafe(raw::git_commit_message_encoding(&*self.raw)) };
         bytes.and_then(|b| str::from_utf8(b).ok())
     }
 
     /// Get the full raw message of a commit.
     ///
     /// `None` will be returned if the message is not valid utf-8
-    pub fn message_raw(&self) -> Option<&str> {
+    pub fn message_raw(&self) -> Option<&'repo str> {
         str::from_utf8(self.message_raw_bytes()).ok()
     }
 
     /// Get the full raw message of a commit.
-    pub fn message_raw_bytes(&self) -> &[u8] {
-        unsafe { crate::opt_bytes(self, raw::git_commit_message_raw(&*self.raw)).unwrap() }
+    pub fn message_raw_bytes(&self) -> &'repo [u8] {
+        unsafe { crate::opt_bytes_very_unsafe(raw::git_commit_message_raw(&*self.raw)).unwrap() }
     }
 
     /// Get the full raw text of the commit header.
     ///
     /// `None` will be returned if the message is not valid utf-8
-    pub fn raw_header(&self) -> Option<&str> {
+    pub fn raw_header(&self) -> Option<&'repo str> {
         str::from_utf8(self.raw_header_bytes()).ok()
     }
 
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks for the PR!  However, I feel like this isn't correct.

Looking at `author()` specifically, the `git_signature` is owned by the `git_commit` object. The function [`git_commit__free`](https://github.com/libgit2/libgit2/blob/12832bab7363e92bfa45d1ef3c3bc10495d71abd/src/libgit2/commit.c#L26-L42) takes care of freeing the signature. Although git objects are reference counted, I believe there are scenarios where those internal caches can get cleared which can evict those objects. Once the object is freed, it will free the signature, leaving the Rust `Signature` object pointing at already freed memory.

bcmyers:
Yep. I think you're right. Thanks for looking into this. I'll close this out.

PR #903, opened 2022-12-12
Add bindings for git_index_find_prefix

Zeerooth:
Plus a test that checks `find_prefix` and `get_path`

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -600,10 +600,10 @@ impl Index {
     /// Find the first position of any entries matching a prefix.
     ///
     /// To find the first position of a path inside a given folder, suffix the prefix with a '/'.
-    pub fn find_prefix(&self, path: &Path) -> Result<Option<usize>, Error> {
+    pub fn find_prefix<T: IntoCString>(&self, prefix: T) -> Result<Option<usize>, Error> {
         unsafe {
             let mut at_pos: size_t = 0;
-            let entry_path = path.into_c_string()?;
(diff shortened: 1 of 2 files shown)


src/index.rs
```diff
@@ -596,6 +596,26 @@ impl Index {
...
+            let result = call!(raw::git_index_find_prefix(
+                &mut at_pos,
+                self.raw,
+                entry_path
+            ));
+            if result == 0 {
+                Ok(Some(at_pos))
+            } else {
+                Ok(None)
+            }
```
ehuss, line 616: This doesn't ever seem to return an error. Should this perhaps be:

```suggestion
            if result == 0 {
                Ok(Some(at_pos))
            } else if result == raw::GIT_ENOTFOUND {
                Ok(None)
            } else {
                Err(Error::last_error(result).unwrap())
            }
```

I understand `git_index_find_prefix` currently never returns any other error, but I'm not very comfortable assuming that will be the case forever.

Also, I'm a little uncomfortable doing the ENOTFOUND check. I think it makes for a more natural Rust interface, but I don't see any other methods in the library doing that (they all just return NotFound). What were your thoughts on this?

Zeerooth: Indeed, I imagined that returning None in situations when the prefix hasn't been found would be more rusty, but after reviewing some other similar functions, it seems like NotFound error is never handled this way, and as there's no point in going against the design of the rest of the library I just decided to leave it inside the Error, as you suggested.

I hope it's okay now

PR #911, opened 2023-01-12
Add bindings to git_indexer

kim:
The indexer API is a lower-level interface for storing and indexing pack
files, which, unlike `git_odb_write_pack`, allows the ouput to be
written to an arbitrary directory. This can be useful when working with
unusual validation requirements or non-standard repository layouts.

### DIFF
--- a/src/indexer.rs
+++ b/src/indexer.rs
@@ -1,7 +1,13 @@
-use std::marker;
+use std::ffi::CStr;
+use std::mem::MaybeUninit;
+use std::path::Path;
+use std::{io, marker, mem, ptr};
 
-use crate::raw;
+use libc::c_void;
+
+use crate::odb::{write_pack_progress_cb, OdbPackwriterCb};
 use crate::util::Binding;
+use crate::{raw, Error, IntoCString, Odb};
 
 /// Struct representing the progress by an in-flight transfer.
 pub struct Progress<'a> {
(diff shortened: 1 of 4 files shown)


kim:
Sorry the test does not actually make sense, will amend

kim:
Amended to make odb optional

kim:
@ehuss any feedback on this?

src/indexer.rs
```diff
@@ -94,3 +100,159 @@ impl<'a> Binding for Progress<'a> {
...
+    /// `mode` is the permissions to use for the output files, use `0` for defaults.
+    ///
+    /// If `verify` is `false`, the indexer will bypass object connectivity checks.
+    pub fn new(odb: Option<&Odb<'a>>, path: &Path, mode: u32, verify: bool) -> Result<Self, Error> {
+        let path = path.into_c_string()?;
+
+        let odb = odb.map(Binding::raw).unwrap_or_else(ptr::null_mut);
+
+        let mut out = ptr::null_mut();
+        let progress = MaybeUninit::uninit();
```
ehuss, line 131: I'm having trouble seeing where `progress` is ever initialized. I see it created here, and then a few calls to `as_mut_ptr`, but I don't see where it gets initialized before `as_mut_ptr` is called. Can you help me understand how this works?

src/indexer.rs
```diff
@@ -94,3 +100,159 @@ impl<'a> Binding for Progress<'a> {
...
+impl io::Write for Indexer<'_> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        unsafe {
+            let ptr = buf.as_ptr() as *mut c_void;
+            let len = buf.len();
+
+            let res = raw::git_indexer_append(self.raw, ptr, len, self.progress.as_mut_ptr());
+
+            if res < 0 {
+                Err(io::Error::new(io::ErrorKind::Other, "Write error"))
```
ehuss, line 197: This is unfortunate that it erases the libgit2 error. I assume you are just copying this from `OdbWriter`. I'm curious why this was implemented as `io::Write` instead of an inherent `write` method that would expose a native error message.

src/odb.rs
```diff
@@ -1,6 +1,7 @@
 use std::io;
 use std::marker;
 use std::mem::MaybeUninit;
+
```
ehuss, line 4: Looks like some stray edits, please try to avoid these.

src/odb.rs
```diff
@@ -10,6 +11,7 @@ use libc::{c_char, c_int, c_uint, c_void, size_t};
 
 use crate::panic;
 use crate::util::Binding;
+
```
ehuss, line 14: Similar stray newline.

src/indexer.rs
kim: I'm not sure, actually, `OdbPackwriter` does the same. I _assumed_ it works because `MaybeUninit<T>` has the same size and alignment as `T`, so the C side can just write to it (i.e. it is initialized after the first time we pass `as_mut_ptr` over to `libgit2`).

I don't know why this would be preferable over deriving `Default` for `git_indexer_progress`, and casting to a raw pointer where needed. If I made this change, would you want me to do it for `OdbPackwriter` as well?

src/indexer.rs
kim: I think `io::Write` makes sense because we can just `io::copy` the pack bytes from e.g. a network socket.

That being said, `Error::last_error` would indeed give the native error here (while it wouldn't for `OdbPackwriter`), so I'll change that.

src/indexer.rs
ehuss: Yea, it looks like OdbPackwriter should also be updated. From what I can tell, the C side is writing to uninitialized memory which is UB to my understanding.

ehuss (approved):
Thanks!

PR #913, opened 2023-01-14
Added some more documentation to the certificate_check function

alexanderkjall:
In order to underline how important it is to handle.

I missed the implications of implementing this callback in my project, and I fear that many others have done the same. Improving the documentation might help to reduce the risk that people in the future also misses this.


ehuss:
Thanks for the PR! I'm going to close as the security issue has now been resolved via the release of libgit2-sys 0.14.2 in conjunction with [GHSA-8643-3wh5-rmjq](https://github.com/libgit2/libgit2/security/advisories/GHSA-8643-3wh5-rmjq). Not implementing the callback shouldn't have the same implications as before.

PR #914, opened 2023-01-18
Bind git_submodule_repo_init

Hyask:
This allows a real initialization of a submodule's repository, where `git_submodule_init` only leaves you with nothing truly usable.


Hyask:
> Thanks, can you add a test for it?

I thought about it, but after a quick look at the end of `submodule.rs`, I wasn't really sure where to start. Are there the only submodule tests? I'll try to add something based on existing `clone_submodule()` test.

ehuss:
Yea, I think it would go at the bottom of `submodule.rs`. And yea, I would just start be copying one of the other tests and then adjusting it to utilize the new method.

Hyask:
I've added a test, do you think it's enough, or is more needed?

ehuss (approved):
Looks good, thanks!

Hyask:
@ehuss thanks for the quick review and the merge. Do you have any idea when this will be released in a tag?

PR #924, opened 2023-01-28
Implement Send for all simple git types

orf:
This implements Send for all simple libgit2-sys types. According to https://github.com/libgit2/libgit2/blob/main/docs/threading.md this is fine - objects can be sent between threads.

Refs https://github.com/rust-lang/git2-rs/issues/194

### DIFF
--- a/src/blame.rs
+++ b/src/blame.rs
@@ -11,6 +11,8 @@ pub struct Blame<'repo> {
     _marker: marker::PhantomData<&'repo Repository>,
 }
 
+unsafe impl<'repo> Send for Blame<'repo> {}
+
 /// Structure that represents a blame hunk.
 pub struct BlameHunk<'blame> {
     raw: *mut raw::git_blame_hunk,
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -14,6 +14,8 @@ pub struct Blob<'repo> {
     _marker: marker::PhantomData<Object<'repo>>,
 }
 
+unsafe impl<'repo> Send for Blob<'repo> {}
+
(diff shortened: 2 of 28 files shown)


mitermayer:
Are there any plans on merging this any time soon ?

orf:
I’m not sure the best way to test this, but I’ve used this branch on an extensively multi-threaded application that opens, utilises and exchanges hundreds of repositories between threads without issue.

That’s definitely a bit “source: trust me bro” but it does seem perfectly fine.

mitermayer:
I really wish this PR could be merged, or if in the docs was some clear documentation with workarounds

ngalaiko:
@orf did you find some issues with that, or just closed due to testing issues?

orf:
Oh, no, I just deleted a bunch of my forks. It works fine

Piturnah:
Any chance of this being reopened?

clouds56:
> I did actually encounter one issue - multiple threads writing to a single odb instance in parallel using the odb writer interface seems to encounter race conditions and suffers from incomplete objects being written.
> 
> That’s one of the few objects explicitly and already marked as “safe to use from multiple threads”.

According to this PR https://github.com/libgit2/libgit2/pull/5595, I think odb of libgit2 should be "thread-safe".
If it doesn't, I think it's a bug in libgit2, not ours.

BTW, although we could use odb in multithread, it could still failed to call due to "locked" by another thread, have you check the returning error code?

I think we should mark them "Send" now and when we encounter some bug in the future, we could help open an issue in libgit2.

joshbenz:
+1

PR #926, opened 2023-02-01
Add bindings for push_negotiation callback

### LINKED ISSUE #733: Add support for push negotiation callback
I'd like to implement something like `git push --force-with-lease`. libgit2/libgit2#4286 says that one can probably do that using the "push negotiation" callback, but it seems that git2-rs doesn't expose that yet.

karin0:
This fixes #733 .

This PR adds `RemoteCallbacks::push_negotiation`, `PushNegotiation` and `PushUpdate` to expose the `push_negotiation` callback. A test is added at the end of `src/remote.rs`.


src/push_update.rs
```diff
@@ -0,0 +1,55 @@
+use crate::util::Binding;
+use crate::{raw, Oid};
+use std::marker;
+use std::str;
+
+/// Represents an update which will be performed on the remote during push.
+pub struct PushUpdate<'a> {
+    raw: *const raw::git_push_update,
+    _marker: marker::PhantomData<&'a raw::git_push_update>,
```
samueltardieu, line 9: I don't think you need the `PhantomData` here: it signals that dropping a `PushUpdate` may drop a `raw::git_push_update`, but:

- you didn't implement `Drop` for `PushUpdate` (and you don't need to)
- libgit2 will take care of freeing the `raw::git_push_update` structure

Unless there is another reason you need the `PhantomData` here, you might want to remove this field.

samueltardieu (commented):
I think this is a great addition to git2-rs, I hope it can be merged soon.

src/remote_callbacks.rs
```diff
@@ -87,6 +89,14 @@ pub type PushTransferProgress<'a> = dyn FnMut(usize, usize, usize) + 'a;
 ///     * total
 pub type PackProgress<'a> = dyn FnMut(PackBuilderStage, usize, usize) + 'a;
 
+/// Callback used to inform of upcoming updates.
+///
+/// The argument is a slice containing the updates which will be sent as
+/// commands to the destination.
+///
+/// The return value indicates whether the push should continue.
+pub type PushNegotiation<'a> = dyn FnMut(&[PushUpdate<'_>]) -> bool + 'a;
```
ehuss, line 98: Is there any particular reason this returns a bool? I was thinking it could return a `Result<(), Error>`, similar to other callbacks so that it is clear that it is triggering an error. A bool seems more innocuous, and not a typical way to do error handling in Rust.

ehuss (commented):
Thanks, sorry for the long delay.

src/remote_callbacks.rs
karin0: Thanks for the review! I've made the changes. I might have been emulating `sideband_progress`, where the callback also returns a `bool` to decide whether to cancel the operation, but using a `Result` is indeed more idiomatic here.

ehuss (approved):
Thanks!

Yea, after looking more closely I realized that the callbacks aren't very consistent if they use bools or Results.

Also, thanks for adding the detailed test!

PR #930, opened 2023-02-17
Support partial stashing

extrawurst:
Work in Progress. 
Need some help here: for some reason `cargo test` fails for me locally. hoping its about my local setup and works in the CI


extrawurst:
so the API is done, but I am a little lost as to why the test does not work, even though I specify a pathspec, both file changes are gone after the stash 🤯

extrawurst:
this PR is ready for merge @ehuss

libgit2-sys/lib.rs
```diff
@@ -1754,6 +1755,17 @@ git_enum! {
     }
 }
 
+#[repr(C)]
+pub struct git_stash_save_options {
+    pub version: c_uint,
+    pub flags: c_uint,
```
ehuss, line 1761: It looks like this is `uint32_t`

```suggestion
    pub flags: u32,
```

src/repo.rs
```diff
@@ -2844,6 +2844,25 @@ impl Repository {
         }
     }
 
+    ///
```
ehuss, line 2847: Looks like there is some missing documentation here.

src/stash.rs
```diff
@@ -1,10 +1,73 @@
...
+                    &mut opts.raw_opts,
+                    raw::GIT_STASH_SAVE_OPTIONS_VERSION,
+                )
+            },
+            0
+        );
+        opts
+    }
+
+    ///
```
ehuss, line 42: Looks like some missing documentation here.

src/stash.rs
```diff
@@ -258,4 +321,40 @@ mod tests {
 
         assert!(stash_name.starts_with("WIP on main:"));
     }
+
+    fn create_file(r: &Repository, name: &str, data: &str) -> PathBuf {
+        let p = Path::new(r.workdir().unwrap()).join(name);
+        fs::File::create(&p)
+            .unwrap()
+            .write(data.as_bytes())
+            .unwrap();
```
ehuss, line 330: This can be written a little more briefly as:

```suggestion
        fs::write(&p, data).unwrap();
```

I realize you're probably just copying what is done above. Feel free to update the tests above to use this function, too.

extrawurst:
@ehuss I addressed your feedback

ehuss (approved):
Thanks!

extrawurst:
@ehuss thanks for merging can you cut a release? :)

ehuss:
Sure! It should be published now.

PR #939, opened 2023-03-22
Add a LIBGIT2_NO_VENDOR environment variable to build.rs

kpcyrd:
This does the following things:

- To test for cargo features you'd usually use `if cfg!(feature = "zlib-ng-compat") { ... }` instead of accessing the environment variable, but since this environment variable could also be set from outside of cargo I'm generating `cargo:rerun-if-env-changed=` output to let cargo know it needs to rebuild if they're changed
- I'm adding a `LIBGIT2_NO_VENDOR` environment variable, this variable is supposed to deny vendoring even if the cargo feature `libgit2-sys/vendored` is enabled by any crate. If the system libgit2 is not present or not compatible it errors the build to let the maintainer know something needs to be fixed. This option would be especially useful for Linux package maintainers who always want to link the system library to avoid "shadow copies" of libgit2 embedded in binaries that are not known to the security team.
- Fix one clippy warning

The `LIBGIT2_NO_VENDOR` variable is analogous to `OPENSSL_NO_VENDOR` in openssl-sys: https://github.com/sfackler/rust-openssl/pull/1253

There's currently no `Error` struct in the build.rs so I'm using a `Result<(), ()>`, if you prefer anything else just let me know.

### DIFF
--- a/libgit2-sys/build.rs
+++ b/libgit2-sys/build.rs
@@ -1,32 +1,59 @@
 use std::env;
+use std::ffi::OsString;
 use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+fn env_var(name: &str) -> Option<OsString> {
+    let var = env::var_os(name);
+    println!("cargo:rerun-if-env-changed={}", name);
+    var
+}
+
+fn use_system_libgit2() -> Result<(), ()> {
+    let mut cfg = pkg_config::Config::new();
+    if let Ok(lib) = cfg.range_version("1.6.4".."1.7.0").probe("libgit2") {
+        for include in &lib.include_paths {
+            println!("cargo:root={}", include.display());
+        }
+        Ok(())
+    } else {
+        Err(())
+    }
+}
+
 fn main() {
-    let https = env::var("CARGO_FEATURE_HTTPS").is_ok();
-    let ssh = env::var("CARGO_FEATURE_SSH").is_ok();
-    let vendored = env::var("CARGO_FEATURE_VENDORED").is_ok();
-    let zlib_ng_compat = env::var("CARGO_FEATURE_ZLIB_NG_COMPAT").is_ok();
+    let https = env_var("CARGO_FEATURE_HTTPS").is_some();
+    let ssh = env_var("CARGO_FEATURE_SSH").is_some();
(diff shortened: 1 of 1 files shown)


classabbyamp:
This would be very useful to void linux, as we've had issues with things not linking to our system libgit2, +1

lovesegfault:
This needs a rebase

kpcyrd:
I've rebased the patch

libgit2-sys/build.rs
```diff
@@ -1,32 +1,59 @@
...
 fn main() {
-    let https = env::var("CARGO_FEATURE_HTTPS").is_ok();
-    let ssh = env::var("CARGO_FEATURE_SSH").is_ok();
-    let vendored = env::var("CARGO_FEATURE_VENDORED").is_ok();
-    let zlib_ng_compat = env::var("CARGO_FEATURE_ZLIB_NG_COMPAT").is_ok();
+    let https = env_var("CARGO_FEATURE_HTTPS").is_some();
+    let ssh = env_var("CARGO_FEATURE_SSH").is_some();
+    let vendored = env_var("CARGO_FEATURE_VENDORED").is_some();
+    let zlib_ng_compat = env_var("CARGO_FEATURE_ZLIB_NG_COMPAT").is_some();
+    let no_vendor = env_var("LIBGIT2_NO_VENDOR").map_or(false, |s| s != "0");
```
weihanglo, line 31: We should doc this new environment variable about why and when people should use it. Maybe in `README.md` or elsewhere.

weihanglo (commented):
> To test for cargo features you'd usually use `if cfg!(feature = "zlib-ng-compat") { ... }` instead of accessing the environment variable, but since this environment variable could also be set from outside of cargo I'm generating `cargo:rerun-if-env-changed=` output to let cargo know it needs to rebuild if they're changed

I am not quite understand the change here. Could we explain a bit more what “you” means here? git2 maintainers of git2 users?`CARGO_FEATURE_*` is something set by cargo for build script to inform external build tools. Cargo features behavior AFAIK is not affected by someone setting those env vars.

> Fix one clippy warning

Should we do one thing per commit? (Or per pull request if you'd like to)

ehuss:
Closing as completed from #966.

PR #941, opened 2023-03-23
Add an empty-iterator example argument

LemmingAvalanche:


LemmingAvalanche:
Drive-by commit since I was wondering about how to pass in an “empty argument” until I looked at Triplett’s test.

Maybe I’m the only one.

ehuss (approved):
Thanks!

PR #948, opened 2023-04-13
Bump libgit2 to 1.6.4

edrevo:
1.6.4 contains just a minor bugfix.


ehuss (approved):
Thanks!

I made a small change to the changelog, since it seemed to be a copy from the previous one (mentioning 1.6.3).

PR #955, opened 2023-05-27
Implemented FusedIterator for various iterators

vallentin:
Implemented `FusedIterator` for various iterators. All the iterators have one thing in common, they wrap a `Range`, which implements `FusedIterator`, thus guaranteeing the outer iterator is also fused.


edrevo (approved):

ehuss (approved):
Thanks!

PR #968, opened 2023-07-20
bump libgit2 to v1.7.0

weihanglo:
[Changelog](https://github.com/libgit2/libgit2/releases/tag/v1.7.0)

See each commit for details.

I'll hold off shallow clone binding support in `git2` crate. We could add that in a follow-up.

### DIFF
--- a/git2-curl/Cargo.toml
+++ b/git2-curl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git2-curl"
-version = "0.18.0"
+version = "0.19.0"
 authors = ["Josh Triplett <josh@joshtriplett.org>", "Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/rust-lang/git2-rs"
@@ -16,7 +16,7 @@ edition = "2018"
 curl = "0.4.33"
 url = "2.0"
 log = "0.4"
-git2 = { path = "..", version = "0.17", default-features = false }
+git2 = { path = "..", version = "0.18", default-features = false }
 
 [dev-dependencies]
 civet = "0.11"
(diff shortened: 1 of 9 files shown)


weihanglo:
@ehuss please have a look. No hurry!

This is my very first time bumping libgit2 version, so might have done something wrong.

git2-curl/Cargo.toml
```diff
@@ -16,7 +16,7 @@ edition = "2018"
 curl = "0.4.33"
 url = "2.0"
 log = "0.4"
-git2 = { path = "..", version = "0.17", default-features = false }
+git2 = { path = "..", version = "0.18", default-features = false }
```
ehuss, line 19: I think the version of git2-curl itself also needs to be updated.

src/remote.rs
```diff
@@ -588,6 +588,7 @@ impl<'cb> Binding for FetchOptions<'cb> {
             prune: crate::call::convert(&self.prune),
             update_fetchhead: crate::call::convert(&self.update_fetchhead),
             download_tags: crate::call::convert(&self.download_tags),
+            depth: 1, // See `GIT_FETCH_OPTIONS_INIT`.
```
ehuss, line 591: I think this needs to be 0. `GIT_FETCH_OPTIONS_INIT` defines the defaults for all the fields up to `proxy_opts`. All fields after that default to 0/null (depth, follow_redirects, and custom_headers). A depth of 1 means to do a shallow clone/fetch.

```suggestion
            depth: 0, // GIT_FETCH_DEPTH_FULL
```

ehuss (commented):
Thank you! This is very much appreciated.

src/remote.rs
weihanglo: I forgot what I was thinking… Anyway, thanks!

git2-curl/Cargo.toml
weihanglo: Updated!

ehuss: Hm, perhaps I'm not seeing it (or wasn't clear)? The version of `git2-curl` should probably be bumped from 0.18.0 to 0.19.0.

ehuss (approved):
Thanks! 😄

PR #969, opened 2023-07-21
chore: switch from ctest to ctest2

weihanglo:
ctest is nearly unmaintained and emits lots of warnings. Switch to ctest2 to fix those. (rust-lang/libz-sys did this a couple years ago)

Also, with `ctest2` we won't be blocked on https://github.com/gnzlbg/ctest/issues/90, so enable systest on beta and nightly channels.


ehuss (approved):
Thanks!

PR #970, opened 2023-07-24
`FnMut` type should be used with a mutable reference

EFanZh:


ehuss:
Thanks, can you provide some more context on this change? What is the motivation for it?

EFanZh:
Since `C` is of type `FnMut(Oid) -> bool`, calling it would require a mutable reference. Currently, `with_hide_callback` accepts a shared reference (`&C`), intuitively, it should not able to call the function.

src/revwalk.rs
```diff
@@ -170,7 +170,7 @@ impl<'repo> Revwalk<'repo> {
             raw::git_revwalk_add_hide_cb(
                 r.revwalk.raw(),
```
ynjqqqnt, line 171: 111

src/revwalk.rs
```diff
@@ -170,7 +170,7 @@ impl<'repo> Revwalk<'repo> {
...
                 Some(revwalk_hide_cb::<C>),
```
ynjqqqnt, line 172: 1
```suggestion
                Some(revwalk_hide_cb::<C>),
```

ehuss (approved):
OK, I think I understand now. The previous API would allow creating multiple immutable references when it shouldn't. I'm not entirely clear how that can manifest as an issue (since there are several lifetime restrictions in place). It would be interesting to see some example that would be a problem.

I'm also curious if there is some more ergonomic API that could handle this situation. Requiring `&mut` for closures that aren't FnMut seems a bit unfortunate. I tried a few ideas but couldn't come up with anything that would work. If anyone has ideas, it would be appreciated.

I'm going to go ahead and merge since this seems correct to me, and I don't have any ideas on how to make it better.

PR #979, opened 2023-08-08
Add shallow cloning capability

blyxyas:
Add shallow cloning (as well as customizable depth), introduced in `libgit2 v1.7.0`

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -41,6 +41,7 @@ pub struct RemoteHead<'remote> {
 /// Options which can be specified to various fetch operations.
 pub struct FetchOptions<'cb> {
     callbacks: Option<RemoteCallbacks<'cb>>,
+    depth: i32,
     proxy: Option<ProxyOptions<'cb>>,
     prune: FetchPrune,
     update_fetchhead: bool,
@@ -509,6 +510,7 @@ impl<'cb> FetchOptions<'cb> {
             follow_redirects: RemoteRedirect::Initial,
             custom_headers: Vec::new(),
             custom_headers_ptrs: Vec::new(),
+            depth: 0, // Not limited depth
         }
     }
 
@@ -535,6 +537,17 @@ impl<'cb> FetchOptions<'cb> {
     /// Defaults to `true`.
     pub fn update_fetchhead(&mut self, update: bool) -> &mut Self {
         self.update_fetchhead = update;
+        self
+    }
+
+    /// Set fetch depth, a value less or equal to 0 is interpreted as pull
+    /// everything (effectively the same as not declaring a limit depth).
+
+    // FIXME(blyxyas): We currently don't have a test for shallow functions
+    // because libgit2 doesn't support local shallow clones.
+    // https://github.com/rust-lang/git2-rs/pull/979#issuecomment-1716299900
+    pub fn depth(&mut self, depth: i32) -> &mut Self {
+        self.depth = depth.max(0);
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/build.rs
```diff
@@ -171,6 +171,28 @@ impl<'cb> RepoBuilder<'cb> {
         self
     }
 
+    /// Specify fetch depth, a value less than zero is interpreted as pull everything (effectively the same as no declaring a limit depth)
+    pub fn depth(&mut self, depth: i32) -> &mut RepoBuilder<'cb> {
```
ehuss, line 175: Instead of adding a new method here, would it be possible to leave this as a setting that required calling `.fetch_options(...)` instead? I am reluctant to try to repeat low-level option settings, since that would require deciding which settings are or are not replicated, and possibly causing confusion when something could be specified in multiple places.

src/build.rs
```diff
@@ -171,6 +171,28 @@ impl<'cb> RepoBuilder<'cb> {
...
+
+        self
+        // if let Some(ref fetch_opts) = &self.fetch_opts {
+        //     fetch_opts.depth(depth);
+        // } else {
+        //     let mut fo = FetchOptions::new();
+        //     fo.depth(depth);
+        //     self.fetch_options(fo);
+        // }
+        // self
```
ehuss, line 193: Looks to be some vestigial work?

src/repo.rs
```diff
@@ -341,6 +341,13 @@ impl Repository {
         Ok(repo)
     }
 
+    /// Clone a remote repository with a depth of one (shallow cloning)
+    ///
+    /// Similar to `git clone --depth 1`
+    pub fn shallow_clone<P: AsRef<Path>>(url: &str, into: P) -> Result<Repository, Error> {
```
ehuss, line 347: Similar to the comment on `RepoBuilder`, I think I would like to avoid having too many convenience functions, since there is a somewhat combinatorial explosion of all the different options someone might want. Using a builder pattern can avoid adding all sorts of variants of different methods.

ehuss (commented):
Thanks for adding this!

Can you also include a test for it?

blyxyas:
I'm not sure how to add a test for this, it seems like there aren't any tests for cloning features.

ehuss:
> I'm not sure how to add a test for this, it seems like there aren't any tests for cloning features.

At the bottom of `remote.rs` there are various tests which set `RemoteCallbacks` options. I would set up a test that creates a repository with 2 commits. Then clone it with depth=1, and verify there is only one commit in the clone. The `test.rs` module contains some testing utilities for creating commits and such.

src/build.rs
blyxyas: So this should be deleted? I find the `depth` parameter to be kinda (very) important imo, but it can be removed without any problems.

blyxyas:
This new commit addresses the comments and contains a first attempt to test it. It should work but it doesn't, I need some guidance here.

The test creates a repo and commits two times to that new repo. Then creates a `TempDir` to clone the original repo using fetch options with `depth` set to 1. But when checking for the first commit that shouldn't exist because `depth` is 1, it's still there.

rawhuul:
Any work on this?

ehuss:
Unfortunately it looks like libgit2 does not do shallow clones on local repositories. I think it requires the smart protocol and relies on the server to figure out what to fetch. I don't think we have any testing infrastructure for the smart protocol, so I think we'll just need to not include a test for this for now.

I have filed https://github.com/libgit2/libgit2/issues/6634 on that issue.

blyxyas:
So for now, it can go without a test? (I'll add a fixme)

ehuss:
Yep.

rawhuul:
Why it has not been merged yet?

ehuss (approved):
Thanks!

PR #981, opened 2023-08-16
Add binding for `git_blame_buffer`

paulvandermeijs:
This pull request adds a binding for `git_blame_buffer`.

While testing I did notice some issues with `git_blame_buffer` but these seem to originate from libgit2 and should be fixed once this pull request is merged: https://github.com/libgit2/libgit2/pull/6572


ehuss (approved):
Thanks!

Issue #995, opened 2023-10-24
Allow setting GIT_OPT_SET_SSL_CERT_LOCATIONS option in libgit2


aharbis:
The [`git2::opts`](https://docs.rs/git2/latest/git2/opts/index.html) module exposes some options, but not all. I'm interested in being able to set `GIT_OPT_SET_SSL_CERT_LOCATIONS` to control where `libgit2` looks for SSL certificates.

If there is some workaround (maybe a build-time configuration?) that would also be helpful.

ehuss:
A PR to add a binding for that option would be welcome. I think it would just roughly follow the form of the other options.

aharbis:
PR raised: https://github.com/rust-lang/git2-rs/pull/997

DanielEScherzer:
It looks like this was done, can someone with the permissions to do so close the task?

weihanglo:
Thanks for the issue queue cleanup!

PR #997, opened 2023-10-31
Add binding for GIT_OPT_SET_SSL_CERT_LOCATIONS

aharbis:
Adds a binding for `git_libgit2_opts` key `GIT_OPT_SET_SSL_CERT_LOCATIONS` in `opts` module.

https://github.com/rust-lang/git2-rs/issues/995

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -195,22 +195,38 @@ pub unsafe fn set_verify_owner_validation(enabled: bool) -> Result<(), Error> {
     Ok(())
 }
 
-/// Set the SSL certificate-authority locations. `file` is the location of a file containing
-/// several certificates concatenated together. `path` is the location of a directory holding
-/// several certificates, one per file. Either parameter may be `None`, but not both.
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks!  I was wondering, instead of surfacing the kinda weird low-level API (which looks to be inherited from OpenSSL's `SSL_CTX_load_verify_locations`) if it could have something that is a little more explicit that doesn't require the use of `Option`. Perhaps something like separate functions:

```rust
pub unsafe fn set_ssl_cert_file<P: IntoCString>(file: P) -> Result<(), Error>
pub unsafe fn set_ssl_cert_dir<P: IntoCString>(path: P) -> Result<(), Error>
```

That way it is a little clearer and more deliberate as to what the caller is doing. WDYT?

aharbis:
@ehuss Good idea, I've updated the PR with the proposed change.

ehuss (approved):
Thanks!

PR #1000, opened 2023-11-05
Example for commit

c-git:
Add comment indicating that there is more content than init, specifically commit. 

I did this because I spent quite a bit of time trying to understand how to commit. I read the docs but that alone wasn't quite enough to get me on my way as I'm very new to this library. I previously looked at this example but only needed how to `init` a repo and missed the example of how to commit that it included. If I understood it well enough I'd create an new file in examples for `commit` how there is a file for the other functions to improve discoverability or add one to the documentation for commit. Maybe one day if I understand better I'll come back and do that but or now adding this to the description of this example to help the next person.
Alternatively maybe we could point to this example from the docs, if that is easy.

### DIFF
--- a/examples/init.rs
+++ b/examples/init.rs
@@ -1,5 +1,5 @@
 /*
- * libgit2 "init" example - shows how to initialize a new repo
+ * libgit2 "init" example - shows how to initialize a new repo (also includes how to do an initial commit)
  *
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1002, opened 2023-12-10
Implement minimal support for chained credential store helpers

mschlumpp:
Tools such as [git-credential-oauth](https://github.com/hickford/git-credential-oauth) rely on a chained helper configuration to store the acquired credentials.

Unimplemented features in this PR: 
- This does not pass on properties other than `username` and `password` (Notable other properties would be `password_expiry_utc`, `oauth_refresh_token`).
- It does not check the `password_expiry_utc` property, and instead relies on the `timeout` config of the cache helper.

Note: `git-credential-oauth` does not seem to make use of these features at the moment.


mschlumpp:
I don't need this anymore (I switched back to SSH keys). But if somebody wants to continue with this, feel free to reuse any code from this PR.

PR #1004, opened 2023-12-31
Tree iteration: Make skipping ahead more efficient

joshtriplett:
Implement `TreeIter::nth` to make jumping ahead more efficient, and add tests for it.

Might be easiest to read the three commits separately.


ehuss (approved):
Thanks!

PR #1014, opened 2024-02-02
Allow find object by prefix

Brooooooklyn:


ehuss (commented):
Thanks, can you add a test to the bottom of the file?

ehuss (approved):
Thanks!

PR #1030, opened 2024-02-29
feat: custom message field

yangnianbing:
add  `message`  func， can set message when call `stash_save_ext`


ehuss:
Thanks! Can you add a test for this at the bottom of the file?

Issue #904, opened 2022-12-28
git2 holds open deleted packfiles, effectively leaking disk space


roy-work:
We have a long-running daemon that does some stuff with a repository.

We noticed that the disk backing the repository was "full": 

```
thing@thing-0:/var/cache/thing$ du -chs .
7.3G	.
7.3G	total
thing@thing-0:/var/cache/thing$ df -h .
Filesystem      Size  Used Avail Use% Mounted on
/dev/sdb         16G   16G     0 100% /var/cache/thing
```
That might see illogical, but that's a hit that deleted files are being held open. If we `ls` our daemon, that confirms it:

```
0 -> /dev/null
1 -> 'pipe:[16731674]'
10 -> 'socket:[16737720]'
11 -> 'anon_inode:[eventpoll]'
12 -> 'anon_inode:[eventfd]'
13 -> 'anon_inode:[eventpoll]'
14 -> 'anon_inode:[eventfd]'
15 -> 'anon_inode:[eventpoll]'
16 -> 'socket:[16729776]'
17 -> 'anon_inode:[eventpoll]'
18 -> 'anon_inode:[eventfd]'
19 -> 'anon_inode:[eventpoll]'
2 -> 'pipe:[16731675]'
20 -> 'socket:[16729776]'
21 -> '/var/cache/thing/monorepo/.git/objects/pack/pack-a02eda73e77509f55e734b4f80f0e49c60093184.pack (deleted)'
22 -> '/var/cache/thing/environments/.git/objects/pack/pack-be204a28036cbc4cd01c2248a6ee2168a1989a96.pack (deleted)'
23 -> '/var/cache/thing/environments/.git/objects/pack/pack-6e09f3d300e72f084b05ba7bc1ccce40c10f582c.pack (deleted)'
24 -> '/var/cache/thing/monorepo/.git/objects/pack/pack-daec94cd70f17f2b1d3a638ce87cb4c086ce4629.pack (deleted)'
25 -> '/var/cache/thing/environments/.git/objects/pack/pack-2ad6f4562cd68fbf4ecc4d5af17579aa590e7263.pack (deleted)'
26 -> '/var/cache/thing/environments/.git/objects/pack/pack-aed3c72e24e8ec9a5db6b11bc2a81f071b5596f6.pack (deleted)'
27 -> '/var/cache/thing/environments/.git/objects/pack/pack-d4e7ed286a0e2e79cd36688ac33c5f98948e6b6a.pack (deleted)'
[...]

Issue #626, opened 2020-10-07
Too many open files issue from libgit2


WaffleLapkin:
My program, that uses `git2` crushes ~every week because of leaking descriptors. This is probably caused by https://github.com/libgit2/libgit2/issues/2758, which was fixed a while ago.

Is it possible to fix the rust binding too? (If I understand right, this requires an update of the `libgit2` dependency)

alexcrichton:
Updates of the libgit2 module should be fine! Feel free to send a PR which updates it to the latest master branch of libgit2.

ehuss:
@WaffleLapkin libgit2-sys-0.12.13+1.0.1 already included https://github.com/libgit2/libgit2/pull/5396.  Which version are you using?  Also, libgit2 was just updated in #627 (released as 0.12.14+1.1.0), you can maybe try that to see if it helps.

WaffleLapkin:
@ehuss thanks for the callout!

I was using `git2 v0.13.6` with `libgit2-sys v0.12.7+1.0.0`. I'll try to update it and see if it will help.

WaffleLapkin:
Hm, it seems like even with `git2 v0.13.12`/`libgit2-sys v0.12.14+1.1.0` descriptors are still leaking (most of the descriptors are `.git/objects/pack/pack-<...>.{idx,pack}`). I've updated everything yesterday and today there are already ~`3028` open file descriptors (output of `lsof -p <PID> | wc -l`)

lhchavez:
@WaffleLapkin are these mostly `.idx` files? if so, that's expected, since we `mmap(2)` those files and keep them around: https://github.com/libgit2/libgit2/pull/5396#issuecomment-585409066

also, can you try setting the [`GIT_OPT_SET_MWINDOW_FILE_LIMIT`](https://github.com/libgit2/libgit2/blob/27e34f9b9843f7bcc33a4ccfe3e395fe303cba63/include/git2/common.h#L240-L243) configuration option to see if that addresses the issue? The default is unlimited.

initcrash:
@WaffleLapkin Are you keeping an instance of `git2::Repository` around for a long time?
I used to do that and also observed file handles never getting closed but during the lifetime of `git2::Repository` this
is ok I think. If you are implementing a long process it is probably a better idea to `git2::Repository::open(..)` whenever there is any work to do and then discard it when done.
If you are already doing it this way and there are still leaking descriptors that would be a bigger problem.

[...]

PR #1035, opened 2024-03-11
Add bindings for mwindow opts

rpelliard:
This adds bindings for the mwindow options (https://libgit2.org/libgit2/#HEAD/group/libgit2/git_libgit2_opts)

This will avoid calling libgit2-sys directly to set these options, see related issues where this would be useful:
- https://github.com/rust-lang/git2-rs/issues/626#issuecomment-765654455
- https://github.com/rust-lang/git2-rs/issues/904#issuecomment-1382129456

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -234,94 +234,106 @@ where
 }
 
 /// Get the maximum mmap window size
-pub fn get_mwindow_size() -> Result<libc::size_t, Error> {
+///
+/// # Safety
+/// This function is reading a C global without synchronization, so it is not
+/// thread safe, and should only be called before any thread is spawned.
+pub unsafe fn get_mwindow_size() -> Result<libc::size_t, Error> {
     crate::init();
 
     let mut size = 0;
 
-    unsafe {
-        try_call!(raw::git_libgit2_opts(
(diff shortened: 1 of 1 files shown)


rpelliard:
@ehuss Happy to do that change ! Just to make sure I understand correctly, do you suggest casting the `usize` to `libc::size_t` (or `core::ffi::size_t`?) before passing it to libgit2, or completely changing the param/return types of the public interface to stop using `usize` completely ?

ehuss:
Good question. I would probably just change the types of the function signatures to be `libc::size_t`. Since it is a type alias, it probably won't matter much, and I believe will allow the caller to use usize values without any issues.

It is a bit unfortunate that the usize↔size_t relationship isn't clearer. Being a type alias (and not a newtype wrapper) makes it even more ambiguous.

rpelliard:
Makes sense, I pushed an update to use `libc::size_t` as parameter and return type

ehuss:
Thanks! One more thing, I'm pretty sure these need to be unsafe. There is no synchronization in modifying these values, so they should only be done on the main thread before any threads are spawned. (At least, that is my understanding. Interaction with C globals seems like a dark art.)

rpelliard:
Good point, done !

ehuss (approved):
Thanks!

PR #1007, opened 2024-01-07
examples: replace structopt with clap

tamird:
See https://github.com/TeXitoi/structopt/issues/525.

Currently git2 is the most downloaded dependent of structopt:
https://crates.io/crates/structopt/reverse_dependencies.

PR #1036, opened 2024-03-18
Release 0.18.3

rpelliard:
Release containing #1007 and #1035. 

All other changes seem to be unnecessary imports removal & FUNDING.json, not sure if worth mentioning in the Changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+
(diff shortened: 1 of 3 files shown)


ehuss:
Thanks! Unless there is some specific reason to make this a semver-breaking change, I think it would be fine to make it 0.18.3.

ehuss (approved):
Thanks!

PR #1044, opened 2024-04-14
Improve docs for PushUpdate, push_negotiation, plus a bit more

ilyagr:
For PushUpdate, I took some of the phrasing from
https://libgit2.org/libgit2/#HEAD/type/git_push_update

Before looking at the code, I didn't realize that the functions returning `None` means the reference name is not UTF-8, as opposed to the reference not existing.

(I wish the return value for `..._refname` functions would just be `Result<&str, UTFDecodeError>`, but that would be a breaking change.)

For PushNegotiation and push_negotiation, only the latter has docs accessible via docs.rs, but its docstring was less helpful. See also the commit description.

### DIFF
--- a/src/push_update.rs
+++ b/src/push_update.rs
@@ -28,17 +28,17 @@ impl PushUpdate<'_> {
         unsafe { crate::opt_bytes(self, (*self.raw).src_refname).unwrap() }
     }
 
-    /// Returns the source name of the reference.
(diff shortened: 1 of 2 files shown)


ehuss (approved):
Thanks!

I pushed a small change to fix the indentation in the documentation.

PR #1047, opened 2024-04-28
WIP: ( feat: create function to allow setting commit_create_cb while rebasing )

### LINKED ISSUE #850: Support git_commit_create_cb in RebaseOptions
Currently, the `RebaseOptions` structure is missing support for `commit_create_cb` from the `git_rebase_options` C structure, which is used to facilitate things like commit signing in rebase operations. It would be useful to have safe access to this field to allow for custom commit signing abilities for rebase operations. It looks like the `git_rebase_options` structure has a `payload` field at the end which can be used for maintaining state across the FFI boundary, so implementation is hopefully not too complicated.

Followup discussion: `signing_cb` is also unimplemented, but marked as deprecated in the libgit2 C api. As such, it does not seem necessary to me to implement support for it. `commit_create_cb` seems to supersede its functionality.

Pranav2612000:
Fixes #850

### DIFF
--- a/src/rebase.rs
+++ b/src/rebase.rs
@@ -1,10 +1,14 @@
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::{marker, mem, ptr, str};
+
+use libc::{c_char, c_int, c_void};
+use raw::{git_commit, git_oid, git_signature, git_tree, GIT_PASSTHROUGH};
 
 use crate::build::CheckoutBuilder;
 use crate::util::Binding;
-use crate::{raw, Error, Index, MergeOptions, Oid, Signature};
+use crate::{panic, raw, signature, Commit, Error, Index, MergeOptions, Oid, Signature, Tree};
 
+pub type CommitCreate<'a> = dyn FnMut(&Oid, &Signature<'_>, &Signature<'_>, Option<&str>, Option<&str>, &Tree<'_>, usize, &Commit<'_>) -> i32 + 'a;
 /// Rebase options
 ///
 /// Use to tell the rebase machinery how to operate.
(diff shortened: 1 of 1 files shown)


Pranav2612000:
I'll make the pipelines pass but I just wanted to get an early review on the approach.
Do you think this could work? The `commit_create_cb_op` expects a callback fn which has arguments slightly different from the one expected by `raw.commit_create_cb` ( e.g Oid instead of *mut git_oid, Signature instead of *const git_signature ) but the rust typechecker did not throw an error.

CouleeApps:
Just throwing my opinion on here since I was tagged on the issue: I don't think that transmute is safe (the rust parameter types probably don't map cleanly to c ffi). I would use something like this: (example from my job) https://github.com/Vector35/binaryninja-api/blob/dev/rust/src/debuginfo.rs#L181-L246 where Rust puts all of its information into a Box type provided to the userdata (void*) parameter of the callback, and a static callback function which takes the boxed data and interprets it properly.

ehuss:
`remote_callbacks.rs` has some examples for handling callbacks (using an `extern "C"` callback with a `void *` payload for the function pointer). That might provide some inspiration.

Pranav2612000:
Thank you for the help @CouleeApps @ehuss 
I'll take a look at the code you shared and take another stab at it soon

Pranav2612000:
@ehuss I've followed an approach similar to the one used in `remote_callback.rs`. Can you take a look again?

Pranav2612000:
Also, I'm testing the changes through an application I have and passing `GITPASSTHROUGH` works as expected, but passing other values, which should result in the stopping and returning a failure crashes the app. The error is inside `rebase.commit` func. Here's a sample code I'm using

```
                if let Ok(commit_id) = rebase.commit(None, &committer.clone().into(), None) {
                    last_rebase_head = commit_id.into();
                } else {
                    rebase_success = false;
                    break;
                }
```
I've added print statements and ensured that it doesn't reach the `else` block before crashing.

Do you know what may be happening?

Pranav2612000:
Bumping this up

PR #1061, opened 2024-06-22
 odb: fix OdbReader read return value

gensmusic:
git_odb_stream_read return the number of bytes read.

This should be a typo, and will cuase [Read.read_to_end](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end) to read for ever since it will continuously call  `read` to append more data to buf until `read` returns either `Ok(0)`
 or an error of non-ErrorKind::Interrupted kind.

### DIFF
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -364,7 +364,7 @@ impl<'repo> io::Read for OdbReader<'repo> {
             if res < 0 {
                 Err(io::Error::new(io::ErrorKind::Other, "Read error"))
             } else {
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! I believe this is correct, but the libgit2 docs say it returns 0. I believe the docs are incorrect, so I opened https://github.com/libgit2/libgit2/pull/6837 just to verify.

ehuss (approved):
Looks like upstream got merged. Thanks!

PR #1066, opened 2024-07-10
Update `log` crate and create shim method(s) between `git2`'s tracing and `log`

vcfxb:
This change came out of some work I'm doing on https://github.com/mitre/hipcheck, and it seemed like it would be useful upstream as well.

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -11,22 +11,22 @@ use crate::{panic, raw, util::Binding};
 pub enum TraceLevel {
     /// No tracing will be performed.
     None,
-    
+
     /// Severe errors that may impact the program's execution
     Fatal,
-    
+
     /// Errors that do not impact the program's execution
     Error,
-    
+
     /// Warnings that suggest abnormal data
     Warn,
-    
+
     /// Informational messages about program execution
     Info,
-    
+
     /// Detailed data that allows for debugging
     Debug,
-    
+
     /// Exceptionally detailed debugging data
     Trace,
 }
@@ -65,7 +65,7 @@ impl TraceLevel {
     /// are trivial.
     pub const fn as_log_level_filter(self) -> log::LevelFilter {
         use log::LevelFilter;
-        
+
         match self {
             Self::None => LevelFilter::Off,
             Self::Fatal | Self::Error => LevelFilter::Error,
@@ -75,7 +75,7 @@ impl TraceLevel {
             Self::Trace => LevelFilter::Trace,
         }
     }
-    
+
     /// Attempt to convert this [TraceLevel] to a [log::LevelFilter].
     ///
     /// This is done trivially with two exceptions:
@@ -83,7 +83,7 @@ impl TraceLevel {
     /// - [TraceLevel::Fatal] goes to [log::Level::Error].
     pub const fn as_log_level(self) -> Option<log::Level> {
         use log::Level;
-        
+
         match self {
             Self::None => None,
             Self::Fatal | Self::Error => Some(Level::Error),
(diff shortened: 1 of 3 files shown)


src/tracing.rs
```diff
@@ -57,14 +58,51 @@ impl Binding for TraceLevel {
     }
 }
 
+impl TraceLevel {
+    /// Convert the [TraceLevel] to a [log::LevelFilter].
+    ///
+    /// [TraceLevel::Fatal] becomes [log::LevelFilter::Error] but otherwise all the conversions
+    /// are trivial.
+    pub const fn as_log_level_filter(self) -> log::LevelFilter {
```
ehuss, line 66: What is the use case for this function?

ehuss (commented):
I'm a little uncertain about adding this, because it makes `log` become a public dependency of git2. Updating public dependencies can cause problems, and I generally would like to avoid them if at all possible. It's also not clear to me why we would choose "log" over say "tracing" or any other option. I'm not yet sure if I would like to add that.

Generally, all changes should have a test.

Also, I'm a little uneasy with moving forward with this since it looks like the existing code has undefined behavior. Is that something you can maybe fix in a separate PR?

vcfxb:
> I'm a little uncertain about adding this, because it makes `log` become a public dependency of git2. Updating public dependencies can cause problems, and I generally would like to avoid them if at all possible. It's also not clear to me why we would choose "log" over say "tracing" or any other option. I'm not yet sure if I would like to add that.
> 
> Generally, all changes should have a test.
> 
> Also, I'm a little uneasy with moving forward with this since it looks like the existing code has undefined behavior. Is that something you can maybe fix in a separate PR?

I think you're right in your assessment of public dependencies and their issues. The best solution to this in my mind is to just make any function that interacts with a type from `log` private, so that we can still consume the crate's facade (much in the same way that `src/cred.rs` did before this PR) without publicly using `log` types. I'll make those changes and update this PR. 

In regards to the existing undefined behavior, I'll have to take a look, but I would be happy to open another PR if I can find what you're referring to.

vcfxb:
https://github.com/rust-lang/git2-rs/pull/1071 fixes the undefined behavior you mentioned.

src/tracing.rs
```diff
@@ -75,6 +98,27 @@ pub fn trace_set(level: TraceLevel, cb: TracingCb) -> bool {
     return true;
 }
 
+/// Passes [trace_set] a shim function to pass tracing info to the [log] crate.
+pub fn trace_shim_log_crate() {
```
weihanglo, line 102: While this shim is nice, I don't think every git2 user needs this. git2, from my observation [^1], has tried to be a thin layer upon libgit2. This kind of glue code is pretty application facing, and IMO should leave for users to decide. 

[^1]: I am not really a so-called long-term maintainer

src/tracing.rs
```diff
@@ -57,14 +58,36 @@ impl Binding for TraceLevel {
...
+    ///
+    /// This is done trivially with two exceptions:
+    /// - [TraceLevel::None] goes to [None]
+    /// - [TraceLevel::Fatal] goes to [log::Level::Error].
+    const fn as_log_level(self) -> Option<log::Level> {
+        use log::Level;
+
+        match self {
+            Self::None => None,
+            Self::Fatal | Self::Error => Some(Level::Error),
```
weihanglo, line 74: This has the same symptom. Different users may have different thoughts on mapping Fatal and Error to different levels. We try not to decide this kind of application logic for users.

weihanglo:
Just saw ehuss has left comments and noticed the public dependency issue I wasn't aware of. I totally agree with that being a way bigger problem than other issues I've pointed out. Given that, I don't think this is a thing we're going to merge. Thanks for being interested in upstreaming your changes anyway!

vcfxb:
Of course! I completely understand -- it may be worth upgrading `log` regardless -- I leave that to you.

PR #1069, opened 2024-07-18
Make `transport::Service` `Debug` and add some HTTP constants from libgit2.

vcfxb:
This PR adds a dependency on the `http` crate for the `http::Method` type, and adds some HTTP related constants defined in `libgit2` at https://github.com/libgit2/libgit2/blob/2ecc8586f7eec4063b5da1563d0a33f9e9f9fcf7/src/libgit2/transports/http.c#L68-L95.

### DIFF
--- a/src/transport.rs
+++ b/src/transport.rs
@@ -54,13 +54,58 @@ pub trait SmartSubtransport: Send + 'static {
 }
 
 /// Actions that a smart transport can ask a subtransport to perform
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, Debug)]
 #[allow(missing_docs)]
 pub enum Service {
     UploadPackLs,
     UploadPack,
     ReceivePackLs,
     ReceivePack,
+}
+
+/// HTTP implementation related info for various services.
+/// 
+/// This information was pulled from 
+/// <https://github.com/libgit2/libgit2/blob/2ecc8586f7eec4063b5da1563d0a33f9e9f9fcf7/src/libgit2/transports/http.c#L68-L95>. 
+impl Service {
+    /// The HTTP Method used by libgit2's implementation for http(s) transport.
+    pub const fn http_method(self) -> http::Method {
+        use http::Method;
+
+        match self {
+            Service::UploadPackLs | Service::ReceivePackLs => Method::GET,
+            Service::UploadPack | Service::ReceivePack => Method::POST,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


vcfxb:
This would be useful for users implementing their own git transport using one of rust's http(s) libraries (i.e. ureq, reqwest, etc). 

I think it's super fair to not want to add more public dependencies (due to the cost downstream to match the version and have an extra crate to compile). 

The solutions here are either to gate anything using/exposing the `http` crate behind a feature flag (thus making the dependency on the `http` crate optional, off by default), or replace the use of `http::Method` with a bespoke enum native to the git2 crate (eliminating the need to depend on `http` at all).

src/transport.rs
```diff
@@ -63,6 +63,51 @@ pub enum Service {
...
+        match self {
+            Service::UploadPackLs | Service::ReceivePackLs => Method::GET,
+            Service::UploadPack | Service::ReceivePack => Method::POST,
+        }
+    }
+
+    /// The value of the HTTP "Accept" header that is used by libgit2's implementation for http(s) transport.
+    pub const fn http_accept_header(self) -> &'static str {
+        match self {
+            Service::UploadPackLs => "application/x-git-upload-pack-advertisement",
```
weihanglo, line 84: Is this a public interface exposed by libgit2? If not I tend not to add this. We don't really want to get into troubles that there is no stability guarantee in libgit2 while git2-rs exposes it.

src/transport.rs
```diff
@@ -54,7 +54,7 @@ pub trait SmartSubtransport: Send + 'static {
 }
 
 /// Actions that a smart transport can ask a subtransport to perform
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, PartialEq, Debug)]
```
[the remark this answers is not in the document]
vcfxb: Sounds good -- split off into #1074

src/transport.rs
vcfxb: The entire `Service` type is native to `git2-rs` -- libgit2 doesn't have anything really matching it. The closest thing is https://github.com/libgit2/libgit2/blob/main/src/libgit2/transports/http.c but that is not publicly exposed. I can definitely understand not merging this because of that.

weihanglo: Really appreciate your understanding. Yeah we need to weight on that.

Anyway, as GitHub keeps diff in a pull request indefinitely, this can be used as a reference when people need such a glue code, which is great!

Per discussion, I am going to close this and thank you again!

vcfxb: Of course! You're welcome!

PR #1071, opened 2024-07-22
(Breaking Changes) Re-implement tracing wrapper in a safer way, replacing the use of `&str` with `&[u8]`.

vcfxb:
This changes the type signature of `TracingCb` to use a `&[u8]` over an `&str`, leaving the conversion to the consumer as planned. This is a breaking change and will require a new release. 

Internally, this replaces the use of a static `AtomicUsize` to store the tracing subscriber with a `AtomicPtr`. This is still unsafe, but lets us avoid the integer to pointer transmutation that was previously happening which is more explicitly a possible violation of Rust's memory model. Now all the casts and transmutations in `tracing.rs` are between pointer types, which is (relatively) safer.

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -1,8 +1,8 @@
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::{ffi::CStr, sync::atomic::{AtomicPtr, Ordering}};
 
-use libc::c_char;
+use libc::{c_char, c_int};
 
-use crate::{panic, raw, util::Binding};
+use crate::{panic, raw, util::Binding, Error};
 
 /// Available tracing levels.  When tracing is set to a particular level,
 /// callers will be provided tracing at the given level and all lower levels.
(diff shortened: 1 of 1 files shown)


src/tracing.rs
```diff
@@ -57,29 +60,82 @@ impl Binding for TraceLevel {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::TraceLevel;
+
+    // Test that using the above function to set a tracing callback doesn't panic.
+    #[test]
+    fn smoke() {
+        super::trace_set(TraceLevel::Trace, |level, msg| {
```
ehuss, line 136: Just noting that since this is a global, we can't really have multiple tests for this. And this is registering a trace function that triggers for all tests running after this. That might be ok, but something we'll need to be careful of.

ehuss (approved):
Thanks for the detailed comments!

PR #1072, opened 2024-07-22
(Breaking Changes) Remove error unwraps

vcfxb:
Built on top of #1071, this PR removes the obsolete `Option` wrapper around `Error::last_error` and removes all the unwraps that existed to handle it.


ehuss (approved):
Nice!

vcfxb:
Thanks!

PR #1073, opened 2024-07-25
README: document when libgit2 is statically linked

ilyagr:
I found the previous paragraph in the current version of the README quite confusing.

The text follows my understanding of https://github.com/rust-lang/git2-rs/blob/f1f09cee7b332d2b494d14d46fd2ec8e5203916a/libgit2-sys/build.rs#L25-L56. I hope it's OK to hide the bit about `zlib` under the case where an "appropriate version of libgit2 cannot be found". Feel free to edit this further.

Cc: commit 59a81cac

I looked into this because I was confused that, after upgrading the Homebrew version of libgit2, my builds of https://github.com/martinvonz/jj started linking libgit2 statically. I'm planning to try enabling `vendored_libgit2` in that project permanently and recommending that people use the environment variable `LIBGIT2_NO_VENDOR=1` if they want dynamic linking.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,17 @@ stable release as well.
 Currently this library requires libgit2 1.7.2 (or newer patch versions). The
 source for libgit2 is included in the libgit2-sys crate so there's no need to
 pre-install the libgit2 library, the libgit2-sys crate will figure that and/or
-build that for you.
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1075, opened 2024-08-11
Don't show console on Windows

sheremetyev:
When used inside GUI application, git2 causes console window to appear temporarily, due to execution of sh.

Command needs special flags to prevent that. More details: https://stackoverflow.com/a/60958956

(For context - similar change in gitoxide: https://github.com/Byron/gitoxide/pull/1506)

### DIFF
--- a/src/cred.rs
+++ b/src/cred.rs
@@ -372,6 +372,12 @@ impl CredentialHelper {
         // If that fails then it's up to the user to put `sh` in path and make
         // sure it works.
         let mut c = Command::new("sh");
+        #[cfg(windows)]
+        {
+            use std::os::windows::process::CommandExt;
+            const CREATE_NO_WINDOW: u32 = 0x08000000;
+            c.creation_flags(CREATE_NO_WINDOW);
+        }
         c.arg("-c")
             .arg(&format!("{} get", cmd))
             .stdin(Stdio::piped())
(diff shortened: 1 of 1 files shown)


ehuss:
@ChrisDenton does this make sense to you?  Essentially, when configuring a credential helper on Windows, the git2 library executes it (first trying with `sh`). It sounds like in a GUI application this is popping open a console window. Is this the right way to suppress that?

ChrisDenton:
Sure.

The only downside of `CREATE_NO_WINDOW` I can think of is that it may be slightly slower to execute in some circumstances because it needs to create a new console session (albeit not a shell or a window) even if one already exists. However, I doubt that will have a measurable impact. Using `DETACHED_PROCESS` would fix that except if the child process itself spawns child processes then you may get a console window again.

Another option would be to merely [hide any new window](https://doc.rust-lang.org/stable/std/os/windows/process/trait.CommandExt.html#tymethod.show_window). But the Rust API is nightly only and on older versions of Windows it may still show a flash of the console window before it's hidden.

**tl;dr `CREATE_NO_WINDOW` is probably the best option currently**. An alternative would be to give library users some measure of control over process spawning. But that's more complicated and perhaps unnecessary.

ehuss:
OK, thanks! I'll go ahead and merge, and if we have any issues, we can revisit.

ehuss (approved):
Thanks!

PR #1087, opened 2024-10-02
delete ineffective ssh_key_from_memory crate feature

### LINKED ISSUE #363: ssh_key_from_memory no longer respected
(Disclaimer: I'm very new at this)

It seems to me that one of the side effects of  https://github.com/alexcrichton/git2-rs/commit/76f4b74aef2bc2a54906ddcbf7fbe0018936a69d#diff-612b86ed90c06a3454760b47abd89aea was that the ssh_key_from_memory feature is no longer checked independently of the ssh feature, i.e. GIT_SSH_MEMORY_CREDENTIALS is always defined if GIT_SSH is. This broke my backport of cargo to an older release of Ubuntu where libssh2 is a bit old.

martinvonz:
The ssh_key_from_memory crate feature was added in commit 8ca0e92 (2018-07-26). Then it was accidentally made always-on in commit 76f4b74 (2018-09013). Since it seems like no one has missed the ability to turn it off since then (not enough to send a PR anyway), let's just delete the crate feature.

Closes #363

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,12 +33,11 @@ tempfile = "3.1.0"
 
 [features]
 unstable = []
-default = ["ssh", "https", "ssh_key_from_memory"]
+default = ["ssh", "https"]
 ssh = ["libgit2-sys/ssh"]
 https = ["libgit2-sys/https", "openssl-sys", "openssl-probe"]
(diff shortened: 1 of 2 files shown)


ehuss:
Thanks! Will merge with the next major release.

PR #1088, opened 2024-10-02
Added merge_base_octopus

Caleb-T-Owens:
In gitbutler we've found the need for the octopus-y kind of `merge_base_many`, so I've introduced the `merge_base_octopus` function from libgit2, and also documented the [potentially surprising behavior](https://git-scm.com/docs/git-merge-base#_discussion) of `merge_base_many` which caught me off guard.


ehuss (approved):
Thanks!

PR #1096, opened 2024-11-12
Maintain backwards compatibility for bitflags types

### LINKED ISSUE #1095: Missing `Ord` implementation for `bitflags` types
Between 0.17.0 and git2-0.18.0,  bitflags was updated from 1.x to 2.1, where bitflags removed some derived traits on generated flags, including `Ord` which implies that all bitflags types now no longer are `Ord`. Being a minor release I'd expect it to backwards compatible but it no longer is.

kjughx:
Between versions 0.17.x and git2-0.18.0, bitflags was updated from 1.x to 2.1 which reduced the amount of derived traits on bitflags types. This in turn broke backwards compatibility since now bitflags types no longer implement `PartialOrd`, `Ord` and `Hash`.

The changelog of bitflags 2.0.0 mentions that:
> If you need to maintain backwards compatibility, you can derive the following yourself:
> `#[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]`

This fixes #1095.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -419,7 +419,7 @@ pub enum FileFavor {
 
 bitflags! {
     /// Orderings that may be specified for Revwalk iteration.
-    #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+    #[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Hash)]
     pub struct Sort: u32 {
         /// Sort the repository contents in no particular ordering.
         ///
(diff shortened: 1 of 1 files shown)


ehuss:
Thanks! Can you say what your use case is for being able to order the flags?

kjughx:
Perhaps minor thing, I wrote a program which displays a git status over a workspace which includes many repositories, and wanted the changes to be grouped by their status flag, i.e. first all new files, then all modified or whatever, but group them in sections based on status. A niche use case sure but not a reason to break backwards compatibility imho.

ehuss (approved):
Sounds good, thanks!

PR #1098, opened 2024-11-19
Propagate errors on Tree::walk

edrevo:
`Tree::walk` was ignoring errors coming from `libgit2`. This PR fixes that and adds a unit test.

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -742,7 +742,7 @@ git_enum! {
 }
 
 pub type git_treewalk_cb =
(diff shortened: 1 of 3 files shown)


edrevo:
@weihanglo, would you be able to take a look at this PR?

ehuss (approved):
Thanks! I pushed an update to try to minimize some of the changes here.

I'm not really clear on why https://github.com/rust-lang/git2-rs/commit/d70b64599a080ea91f6052cb234a4a485f5dcdff made all function pointers `Option`. In this particular case, the pointer cannot be NULL, so it seems like it should be fine. 🤷

PR #1100, opened 2024-11-22
Add support for custom allocator

edrevo:
Add support for the `GIT_OPT_SET_ALLOCATOR` option.

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -33,6 +33,29 @@ where
     Ok(())
 }
 
+/// Set the memory allocator to a different memory allocator.
+/// 
+/// This allocator will then be used to make all memory allocations for
+/// libgit2 operations. If the given `allocator` is None, then the
+/// system default will be restored.
+pub unsafe fn set_allocator(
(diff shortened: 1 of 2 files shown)


src/opts.rs
```diff
@@ -33,6 +33,29 @@ where
     Ok(())
 }
 
+/// Set the memory allocator to a different memory allocator.
+/// 
+/// This allocator will then be used to make all memory allocations for
+/// libgit2 operations. If the given `allocator` is None, then the
+/// system default will be restored.
+pub unsafe fn set_allocator(
```
weihanglo, line 41: For an unsafe fn, we'll need to put a comment here for callers to know how to use it safely

weihanglo: Alternatively, we don't provide this unsafe binding. People can use libgit2-sys directly

edrevo: Yeah, that's a fair point. Would it be better to take an `std::alloc::Allocator` as an input?

PR #1102, opened 2024-11-27
docs(config): improve docs of `Config::get_str`/`Config::get_bytes`

RobWalt:
We noticed that the docs around config are not as helpful yet as they could be. Newer rust users tend to assume that using `get_str` wouldn't be as hard to use and that it instead would work the same way as the other getters. This commit:

- Adds a link to `snapshot` method documentation to provide additional guidance for newer Rust users
- Extends the error explanation with a vague reason to enhance understanding of the current behavior


ehuss (approved):
Thanks!

PR #1110, opened 2025-01-02
Expose git_packbuilder_write fn

Letiste:
Expose the libgit2 function `git_packbuilder_write` ([doc](https://libgit2.org/docs/reference/main/pack/git_packbuilder_write.html)) via a new method `write` on the `PackBuilder`.

I need this instead of `write_buf` not to write the whole packfile in a buffer. This simplifies writing a packfile and its index file as well, as instead of having an index and a packbuilder, we can only use the packbuilder.

### DIFF
--- a/src/packbuilder.rs
+++ b/src/packbuilder.rs
@@ -87,9 +87,8 @@ impl<'repo> PackBuilder<'repo> {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1118, opened 2025-01-19
Add binding for GIT_OPT_SET_CACHE_OBJECT_LIMIT

qaqland:
Add one new function following existing code. Document is copy from libgit2.

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -5,7 +5,7 @@ use std::ptr;
 
 use crate::string_array::StringArray;
 use crate::util::Binding;
-use crate::{raw, Buf, ConfigLevel, Error, IntoCString};
+use crate::{raw, Buf, ConfigLevel, Error, IntoCString, ObjectType};
 
 /// Set the search path for a level of config data. The search path applied to
 /// shared attributes and ignore files, too.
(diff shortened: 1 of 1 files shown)


src/opts.rs
```diff
@@ -89,6 +89,25 @@ pub fn enable_caching(enabled: bool) {
     debug_assert!(error >= 0);
 }
 
+/// Set the maximum data size for the given type of object to be considered
+/// eligible for caching in memory.  Setting to value to zero means that that
+/// type of object will not be cached.  Defaults to 0 for [`ObjectType::Blob`]
+/// (i.e. won't cache blobs) and 4k for [`ObjectType::Commit`],
+/// [`ObjectType::Tree`], and [`ObjectType::Tag`].
+pub fn set_cache_object_limit(kind: ObjectType, size: libc::size_t) {
```
ehuss, line 97: I think the function should be unsafe, since there are requirements of synchronizing thread-safe access to global mutable state.

src/opts.rs
```diff
@@ -89,6 +89,25 @@ pub fn enable_caching(enabled: bool) {
...
+    let error = unsafe {
+        raw::git_libgit2_opts(
+            raw::GIT_OPT_SET_CACHE_OBJECT_LIMIT as libc::c_int,
+            kind as libc::c_int,
+            size,
+        )
+    };
+    // This function cannot actually fail, but the function has an error return
+    // for other options that can.
+    debug_assert!(error >= 0);
```
ehuss, line 108: I don't think this is entirely correct, as I believe this can fail if it is given an invalid type. That shouldn't happen since our `ObjectType` should stay in sync with libgit2. Can this comment be updated to clarify that?

src/opts.rs
qaqland: I think you're right. But from this view, isn't `enable_caching` also unsafe?

- https://github.com/libgit2/libgit2/blob/346930e2808eaf9d74d895b6efc3abd085086a33/src/libgit2/cache.c#L21
- https://github.com/libgit2/libgit2/blob/346930e2808eaf9d74d895b6efc3abd085086a33/src/libgit2/cache.c#L25

ehuss (approved):
Thanks!

PR #1120, opened 2025-01-22
fix: prevent crash when repository url is empty

divine:
This is a fix for [#23453](https://github.com/zed-industries/zed/issues/23453)

When git remote URL is a blank it's panicking.


ehuss (approved):
Thanks!

I'm not sure I agree with libgit2's behavior here, so I opened https://github.com/libgit2/libgit2/issues/7028. Returning an empty slice here is probably the most reasonable thing it can do, but that means upstream callers still need to be able to handle that.

PR #1121, opened 2025-01-25
Remove panic wrapper in tracing callback

vcfxb:
This was there when I last modified the tracing code and I left it in without checking on it but looking again at it now, it seems like it should be removed in favor of rust's native `extern "C"` panic handling.

This is arguably a bugfix, so it does not require cutting a new release in my opinion, although you may want to anyway since there's a potential for user's code to start panicking/aborting at runtime where it wasn't before.

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -5,7 +5,7 @@ use std::{
 
 use libc::{c_char, c_int};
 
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

PR #1122, opened 2025-01-29
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2014 to 2025 in the LICENSE file


ehuss:
Thank you for the PR. I'm going to close as the date does not (and usually should not) be updated. It is commonly the "year of first publication" (although that notion is complicated for a continually developed product), and updating it doesn't have a practical effect for the duration of time we care about.

PR #1128, opened 2025-02-21
Update url to min `2.5.4`

ericswpark:
This also resolves the security warning that comes from idna 0.5.0 being vulnerable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ categories = ["api-bindings"]
 edition = "2018"
 
 [dependencies]
-url = "2.0"
+url = "2.5.4"
 bitflags = "2.1.0"
 libc = "0.2"
 log = "0.4.8"
--- a/git2-curl/Cargo.toml
+++ b/git2-curl/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 curl = "0.4.33"
-url = "2.0"
+url = "2.5.4"
 log = "0.4"
 git2 = { path = "..", version = "0.20", default-features = false }
 


weihanglo (approved):
While this looks legit, git2 doesn't pin the version of `url` crate. Packages depending on `git2` should be able to get the latest `url` without problems.

What is the actual security warning from?

ericswpark:
@weihanglo you're right. I initially received an update from Dependabot telling me one of my projects was vulnerable due to idna, and tracked down the dependency graph until I landed on git2. Dependabot couldn't create a patch automatically, so I assumed 2.0 on `cargo.toml` locked versions to `2.0.x`. After running `cargo update` I can verify `url` is updated to 2.5.4.

I can edit the patch to say 2.5 instead of 2.5.4 so it goes up to the version before 3.x, or this PR can be closed if it is unnecessary.

weihanglo:
> I can edit the patch to say `2.5` instead of `2.5.4` so it goes up to the version before `3.x`, or this PR can be closed if it is unnecessary.

Specifying `url = "2.5.4"` doesn't lock the dependency to `2.5.4` either. See <https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#default-requirements>.

weihanglo:
To clarify, `url = "2.0"` has no problem for downstream users, as they will always get the latest `2.y.z` version if they have no pre-existent Cargo.lock, or run `cargo update`.

Your original commit e31ba6a5ac70c104f4b9d081f014e278759bd897 is a correct fix that downstream users will no longer get any version older than 2.5.4, moving the ecosystem away from the vulnerability. Could you also help update `git2-curl` and squash into one commit?

https://github.com/rust-lang/git2-rs/blob/7b7be4398416622fabb7320e478cfc95bfca0833/git2-curl/Cargo.toml#L17

ericswpark:
@weihanglo sounds good, I've dropped the last commit and updated `git2-curl` as well.

weihanglo (approved):

PR #1131, opened 2025-02-28
expose libgit2 `git_branch_upstream_merge`

vlad-anger:
Add missing binding

### DIFF
--- a/libgit2-sys/lib.rs
+++ b/libgit2-sys/lib.rs
@@ -3018,6 +3018,11 @@ extern "C" {
         repo: *mut git_repository,
         refname: *const c_char,
     ) -> c_int;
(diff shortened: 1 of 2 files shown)


weihanglo (approved):
Seems reasonable!

extrawurst:
@weihanglo would be awesome to get a release cut with this

weihanglo:
I've never done any publish git2. Usually it was ehuss doing that. The release process seems [documented here](https://github.com/rust-lang/git2-rs/blob/master/CONTRIBUTING.md#release-process), so could try one when I am available in later this week or during the weekend.

That said, if we already merged some breaking changes, then I would lean towards waiting for the next upstream libgit2 version, and realease a major bump of git2/libgit2-sys with that.

Issue #1136, opened 2025-03-04
Credentials helpers such as `git config credential.helper "store --file /path/to/credentials"`  interpreted wrongly


aibaars:
The implementation of `CredentialsHelper::add_command` of `git2-rs` is different from that of the `git` CLI.

While `git2-rs` checks whether a command contains a `/` or `\` anywhere, the `git` CLI checks whether the command is an absolute path (roughly whether it starts with `/` or `\`)

Compare 
https://github.com/rust-lang/git2-rs/blob/d1ae3b6c2d1200e7d82468af447fa66259225ecf/src/cred.rs#L302-L303
with
https://github.com/git/git/blob/6a64ac7b014fa2cfa7a69af3c253bcd53a94b428/credential.c#L492-L493

This leads to rather surprise behaviour when using the `store` helper with a nontrivial `--file` argument. For example `git config credential.helper "store --file /path/to/credentials"` should be interpreted as `git credential-store --file /path/to/credentials`, but it is not .

PR #1137, opened 2025-03-05
Fix CredentialHelper::add_command

aibaars:
Previously the add_helper function would check whether the command contains a `\` or `/` character to determine whether it is absolute. It should have used starts_with instead of contains. In addition an absolute on Windows a path my start with a drive letter. The git cli implements this by checking if the second character is a ':'.

See also: https://github.com/git/git/blob/6a64ac7b014fa2cfa7a69af3c253bcd53a94b428/credential.c#L492-L493

Fixes https://github.com/rust-lang/git2-rs/issues/1136

### DIFF
--- a/src/cred.rs
+++ b/src/cred.rs
@@ -578,13 +578,13 @@ echo username=c
             return;
         } // shell scripts don't work on Windows
         let td = TempDir::new().unwrap();
-        let path = td.path().join("git-credential-script");
+        let path = td.path().join("git-credential-some-script");
         File::create(&path)
             .unwrap()
             .write(
                 br"\
 #!/bin/sh
-echo username=c
+echo username=$1
 ",
             )
             .unwrap();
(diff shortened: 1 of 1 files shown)


src/cred.rs
```diff
@@ -481,6 +481,12 @@ impl CredentialHelper {
     }
 }
 
+fn is_absolute_path(path: &str) -> bool {
+    path.starts_with('/')
+        || path.starts_with('\\')
+        || cfg!(windows) && path.chars().nth(1).is_some_and(|x| x == ':')
```
ehuss, line 487: Just a minor note, this is slightly different from https://github.com/git-for-windows/git/blob/cca1f38702730b35f52c29efd62864b85e85ddcc/compat/win32/path-utils.c#L9-L31, but probably not enough to matter here.

ehuss (approved):
Thanks!

ehuss:
Note that this was relaxed in https://github.com/rust-lang/git2-rs/pull/429 which previously looked for `starts_with`, but I agree it is probably better to match behavior with git itself, and the `:` check should accommodate the original issue.

src/cred.rs
aibaars: Yes, `git2-rs` uses the `str` type while the git C uses a `char*` and if I understand the git C code correctly, it just tries to handle the case where the drive letter is a UTF8 character. I'd expect the `str::nth` function to take care of that already. However, there might indeed be some behaviour differences in corner cases at I'm not aware of.

PR #1141, opened 2025-03-12
Fix lifetimes of outputs for Patch

### LINKED ISSUE #1135: DiffLine mangled memory
I've come across a strange issue where the use of a `DiffLine` in two different locations ends up with some mangled looking output. I've created a reproducible in this repository https://github.com/FintanH/git2-diff-memory. If you `cargo run` you should see the following output:
```
PRINT FROM INSIDE find_lines
- h2 {
+ .event {
    margin-bottom: 1rem;
  }
- .event {
+ h2 {
    margin-bottom: 1rem;
+   padding: 1px;
  }
  hr {
    margin: 1rem 0;
PRINT FROM OUTSIDE find_lines
�H��U  >�r��Sn-bottom: 1rem;
  }
- .event {
+ h2 {
    margin-bottom: 1rem;
+   padding: 1px;
  }
  hr {
    margin: 1rem 0;
```

Afaict, the outputs should be equal – the range of lines is not different so it should not be selecting other data. My best guess is that the raw pointer within the `DiffLine` is getting corrupted somehow by returning it within the `Vec`.

Let me know if I can provide anymore info to help 😊

ehuss:
This fixes the lifetimes for the functions that return data from a `Patch`. From what I can tell, these are returning data owned by the `Patch` object itself. Previously they had lifetimes tied to the values that *created* the `Patch`, but that allows these outputs to outlive the patch, allowing a use-after-free.

Note that there was a previous fix here in https://github.com/rust-lang/git2-rs/pull/523, but I think that was mistake to change these accessor functions. The `Patch` itself still has a lifetime tied to its inputs introduced in #523.

Fixes #1135

### DIFF
--- a/src/patch.rs
+++ b/src/patch.rs
@@ -125,7 +125,7 @@ impl<'buffers> Patch<'buffers> {
     }
 
     /// Get the DiffDelta associated with the Patch.
-    pub fn delta(&self) -> DiffDelta<'buffers> {
+    pub fn delta<'a>(&'a self) -> DiffDelta<'a> {
         unsafe { Binding::from_raw(raw::git_patch_get_delta(self.raw) as *mut _) }
     }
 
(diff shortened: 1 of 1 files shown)


ehuss:
@weihanglo Would you be able to review this? It's a bit difficult at times to understand the ownership requirements, but I think this is correct.

weihanglo (approved):
Checking with [this libgit2 code snippet](https://github.com/libgit2/libgit2/blob/21a351b0ed207d0871cb23e09c027d1ee42eae98/src/libgit2/patch_parse.c#L1169-L1220), it looks correct. At this it should be there when patch is still valid. Thanks!

PR #1149, opened 2025-03-25
docs(#1148): Add note regarding potentially confusing behavior of `git_checkout_head`.

### LINKED ISSUE #1148: checkout_head is missing disclaimer
[git_checkout_head](https://libgit2.org/docs/reference/main/checkout/git_checkout_head.html) has the following disclaimer:

> Note that this is not the correct mechanism used to switch branches; do not change your HEAD and then call this method, that would leave you with checkout conflicts since your working directory would then appear to be dirty. Instead, checkout the target of the branch and then update HEAD using [git_repository_set_head](https://libgit2.org/docs/reference/main/repository/git_repository_set_head.html) to point to the branch you checked out.

This should be translated to make sense with the rust api and included.

In my usecase `git_checkout_head` was doing literally nothing and it was very confusing.

vcfxb:
closes #1148

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2065,6 +2065,13 @@ impl Repository {
 
     /// Updates files in the index and the working tree to match the content of
     /// the commit pointed at by HEAD.
+    ///
+    /// Note that this is _not_ the correct mechanism used to switch branches;
+    /// do not change your `HEAD` and then call this method, that would leave
+    /// you with checkout conflicts since your working directory would then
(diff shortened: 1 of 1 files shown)


lolbinarycat:
maybe we should provide an example/doctest? I believe the correct sequence is `git_checkout_tree` followed by a `set_head_*`.

ehuss (approved):
Thanks! I instead inlined the upstream docs instead of pointing to a url.

PR #1151, opened 2025-03-29
add a missing `git2::Status::WT_UNREADABLE` field

seokju-na:
This PR addes missing `git2::Status::WT_UNREADABLE` field, which exists in `git_status_t` binding.


ehuss (approved):
Thanks!

PR #1153, opened 2025-04-03
Add missing code checks to Error::raw_code()

alyssais:
These codes are both already handled by `Error::code()`, but because they weren't listed in `Error::raw_code()`, it would return -1, and so the paths in `Error::code()` for them would never be hit.  This fixes the code of errors from `Repository::merge_commits()` with fail_on_conflict set for me.


ehuss (approved):
Thanks!

PR #1155, opened 2025-04-22
fix: add a backtick

reddevilmidzy:
**cleanup_state()** is the method name, but it is not highlighted.

### DIFF
--- a/src/repo.rs
+++ b/src/repo.rs
@@ -2137,7 +2137,7 @@ impl Repository {
     ///
     /// For compatibility with git, the repository is put into a merging state.
     /// Once the commit is done (or if the user wishes to abort), you should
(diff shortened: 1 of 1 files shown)


src/repo.rs
```diff
@@ -2137,7 +2137,7 @@ impl Repository {
     ///
     /// For compatibility with git, the repository is put into a merging state.
     /// Once the commit is done (or if the user wishes to abort), you should
-    /// clear this state by calling cleanup_state().
+    /// clear this state by calling `cleanup_state()`.
```
ehuss, line 2140: If possible, it's usually nice to make it an intra-doc link.

```suggestion
    /// clear this state by calling [`cleanup_state()`][Repository::cleanup_state].
```

PR #1165, opened 2025-06-12
ran `cargo clippy --fix -- -Wclippy::use_self`

adamnemecek:


src/apply.rs
weihanglo, line 1: What is the motivation of this PR?

adamnemecek: Are you taking about the default impl? Or the whole thing?

weihanglo: This description of this PR is missing, which I was looking for.

adamnemecek: I did not enable it, I just ran it. 

It is more idiomatic, explicit types require you to read the type to make sure it's the same.

adamnemecek: ok #1166

ehuss:
Similar to https://github.com/rust-lang/git2-rs/pull/1166#issuecomment-4577075610, I'm going to close this out. Thanks, though!

PR #1166, opened 2025-06-13
Use `Self` where possible

adamnemecek:


vcfxb (approved):

ehuss:
Thanks for the PR! However, I'm going to close as this is probably not worth the churn. Also, this is a matter of style that I think some people prefer the more explicit type name.

PR #1168, opened 2025-06-18
Make url optional via a new "cred" feature gate for credential helpers; disable default features

joshtriplett:
Add a new feature flag for credential helpers, which are the only uses of the `url` crate. This reduces a build of git2 from 54 dependencies to 17, for a configuration that many tools (e.g. anything that only cares about local repositories) can happily use.

Make the "https" and "ssh" features depend on "cred".

Disable all of these features by default, to avoid extra dependencies in the common case of tools that only care about local repositories.

Bump the crate version to 0.21.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,7 +357,7 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.20.2"
+version = "0.21.0"
 dependencies = [
  "bitflags 2.6.0",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ categories = ["api-bindings"]
 edition = "2018"
 
 [dependencies]
-url = "2.5.4"
+url = { version = "2.5.4", optional = true }
 bitflags = "2.1.0"
 libc = "0.2"
 log = "0.4.8"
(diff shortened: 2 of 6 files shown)


joshtriplett:
Rather than hoping the change will have no impact on users of 0.20, I've bumped the version to 0.21, and then disabled default features (since we can do that over a semver-major release).

joshtriplett:
> @joshtriplett I'm having some second thoughts about removing the default features. For example, this essentially breaks some of the examples that are in this repository. I don't really have any data to say whether the majority of users of this library need networking or not. Generally I believe that the difficulty of carefully including the correct set of features (`features = ["https", "ssh", "cred"]`) needs to significantly outweigh setting `default-features=false` which I think is a much easier thing to do. I worry that this is going to be a somewhat pointless annoyance for a large number of users, considering that if there are users that want to trim their dependencies, setting `default-features=false` is relatively much easier.

I think it's a question of what we expect the most common case to be. From what I can tell, the vast majority of git2 crate users are *not* touching the network; they're manipulating local repositories.

You don't need to enable `cred`, since both `https` and `ssh` depend on it; you just need to enable the feature you need for remote access.

I'll send in a PR to fix the examples, and expand the documentation, if that helps.

PR #1180, opened 2025-08-17
git2 feature: Custom Odb backends 

tecc:
This PR is largely an attempt at making a "safe" API for adding custom ODB backends.
Additions made for this purpose have been concentrated in `src/odb_backend.rs`, with minor modifications to `src/odb.rs` and `src/lib.rs` to make it accessible.

A disclaimer, however: custom ODB backends lack documentation in libgit2, so most of the code and documentation (of which there is plenty) is based upon what I think the code does; if someone else knows better, feel free to tell me and I'll fix it. 

I also haven't verified correctness nor safety, although I suspect the code does not violate all too many of Rust's memory rules.

The code has however been documented as thoroughly as I could be bothered to. It's not guaranteed to be correct, because libgit2 lacks documentation, but it should be largely correct as I cross-referenced with both libgit2 and libgit2sharp's source code and documentation to what extent I could.

# Usage

This is largely untested (as of writing), so I may have made mistakes here and there. 

1. Create a type that contains all the information you want from your backend (let's call it `YourBackend`).
2. Implement `git2::odb_backend::OdbBackend` trait for `YourBackend`. The only required method is `fn supported_operations(&self) -> SupportedOperations` - we'll come back to it.
3. Implement the methods you want your backend to support. All methods come with a great deal of implementation notes and recommendations.
4. In the `supported_operations` method, add a flag for each method your backend supports. The appropriate flags are described in each method's documentation.
5. To add `YourBackend` as a backend to an `odb: Odb`, use `odb.add_custom_backend(your_backend, priority)`.
6. If you want to access `YourBackend`, you can do so with the `CustomOdbBackend` handle that was returned in the previous call.

# TODO

There may also be some TODOs littered throughout the code.

- [x] Implement the `readstream` and `writestream` methods (203f81e)
- [x] Implement the `foreach` method (6a6b9ee)
- [ ] Add more implementation notes; the documentation is currently very lacking.
- [ ] Create an example backend using this system. Perhaps try to reimplement one of libgit2's builtin backends?
- [ ] Combine duplicated bindings (most notably `odb_backend::Indexer` to `indexer::Indexer` and `odb_backend::IndexerProgress` to `indexer::Progress`)
- [ ] Reconsider the naming of some of the types and methods before merging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,10 +235,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "fallible-iterator"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2acce4a10f12dc2fb14a218589d4f1f62ef011b2d0cc4b3cb1bba8e94da14649"
+
+[[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
+
+[[package]]
 name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
+
+[[package]]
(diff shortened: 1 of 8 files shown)


WoutDev:
I've been eying this PR for a while now. If I can help in any way, please let me know!

PR #1188, opened 2025-10-06
Add bindings for cache size options

rpelliard:
This PR adds two simple bindings to interact with the ODB cache:
- `GIT_OPT_SET_CACHE_MAX_SIZE` to update the cache max size
- `GIT_OPT_GET_CACHED_MEMORY` to retrieve current and max cache size

See relevant libgit2 docs [here](https://libgit2.org/docs/reference/main/common/git_libgit2_opts.html#:~:text=GIT_OPT_SET_CACHE_MAX_SIZE).

### DIFF
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -111,6 +111,41 @@ pub unsafe fn set_cache_object_limit(kind: ObjectType, size: libc::size_t) -> Re
     Ok(())
 }
 
+/// Set the maximum total data size that will be cached in memory across all
+/// repositories before libgit2 starts evicting objects from the cache. This
+/// is a soft limit, in that the library might briefly exceed it, but will start
+/// aggressively evicting objects from cache when that happens. The default
+/// cache size is 256MB.
+///
+/// # Safety
+/// This function is modifying a C global without synchronization, so it is not
+/// thread safe, and should only be called before any thread is spawned.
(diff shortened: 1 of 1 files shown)


src/opts.rs
```diff
@@ -111,6 +111,41 @@ pub unsafe fn set_cache_object_limit(kind: ObjectType, size: libc::size_t) -> Re
...
+/// Set the maximum total data size that will be cached in memory across all
+/// repositories before libgit2 starts evicting objects from the cache. This
+/// is a soft limit, in that the library might briefly exceed it, but will start
+/// aggressively evicting objects from cache when that happens. The default
+/// cache size is 256MB.
+///
+/// # Safety
+/// This function is modifying a C global without synchronization, so it is not
+/// thread safe, and should only be called before any thread is spawned.
+pub unsafe fn set_cache_max_size(size: libc::size_t) -> Result<(), Error> {
```
ehuss, line 123: Shouldn't all of these values be `ssize_t` (two s's)?

Not that it makes a huge difference, and it's not clear why it is signed, but probably best to match upstream, no?

ehuss (approved):
Thanks!

(And I realize the tests themselves aren't honoring the safety invariants, but I think are mostly ok for that purpose.)

PR #1197, opened 2025-12-06
Bump git2 to 0.20.3

ehuss:
This updates the git2 crate to 0.20.3, with only one change to increase the requirement of libgit2-sys to 0.18.3 to force an update to libgit2 1.9.2. This is only done as a caution to ensure that packages that are only checking direct-dependency versions will pick up the update.


weihanglo (approved):
Thanks!

PR #1205, opened 2026-01-11
refactor: zero OID handling

weihanglo:
* Refactor zero Oid initialization in preparation for SHA256 support (libgit2 add a new `type` field to git_oid struct)
* Reflect that upstream `git_oid_iszero` is deprecated. Use `git_oid_is_zero` instead.

(Spun off from rust-lang/git2-rs#1202)

### DIFF
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -95,9 +95,7 @@ impl<'repo> BlobWriter<'repo> {
     pub fn commit(mut self) -> Result<Oid, Error> {
         // After commit we already doesn't need cleanup on drop
         self.need_cleanup = false;
-        let mut raw = raw::git_oid {
-            id: [0; raw::GIT_OID_MAX_SIZE],
-        };
+        let mut raw = crate::util::zeroed_raw_oid();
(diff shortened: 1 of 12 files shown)


ehuss (approved):
Thanks!

I have some doubts if it makes sense to allow `kind` to be 0 in all situations, since that seems to be an invalid kind. But I don't immediately see any concerns.

weihanglo:
> Thanks!
> 
> I have some doubts if it makes sense to allow `kind` to be 0 in all situations, since that seems to be an invalid kind. But I don't immediately see any concerns.


Good call out!

For output param specifically, kind being 0 should be fine. libgit2 seems to always set out->type based on the type parameter passed to the function or the surrounding context.

For input param, user should not use `Oid::zero`. It guess somehow we should deprecate it in high level API during sha256 stabilization. Even with zero-initialized oid, libgit2 tends to call `git_oid_is_zero` and `git_oid_is_type_valid` to validate input param. I didn't do a full thorough check but seems likely fine.

PR #1201, opened 2025-12-22
feat(sys): experimental SHA256 OID support

weihanglo:
## What this is for

Integrate upstream libgit2 experimental SHA256 support in `libgit2-sys`, including

* A new unstable Cargo feature `unstable-sha256`, gating all new ABI/API changes. We don't offer any SemVer or stability for this feature, as it is subject to change in upstream.
* `libgit2-sys` system lib probing also support discovering `libgit2-experimental.so` at best effort. Since `-experimental` lib is ABI-incompatible with a normal build, if you enable `unstable-sha256`, it won't probe ordinary `libgit2.so`.
* Extend `systest` to check SHA256 bindings.

Note: there is no git2-rs high level API support in this PR

Part of #1090 

### Changes in constants 

* **NEW** `GIT_OID_SHA1_SIZE`
* **NEW** `GIT_OID_SHA1_HEXSIZE`
* **DEPRECATED** `GIT_OID_RAWSZ` (when `usntable-sha256` enabled)
* **DEPRECATED** `GIT_OID_HEXSZ` (when `usntable-sha256` enabled)
* **NEW** `GIT_OID_SHA256_SIZE`
* **NEW** `GIT_OID_SHA256_HEXSIZE`
* **NEW** `GIT_OID_MAX_SIZE` (size varies on `unstable-sha256` feature)
* **NEW** `GIT_OID_MAX_HEXSIZE` (size varies on `unstable-sha256` feature)
* **NEW** `GIT_INDEX_OPTIONS_INIT`
* **NEW** `GIT_INDEX_OPTIONS_VERSION`
* **NEW** `GIT_REPOSITORY_NEW_OPTIONS_INIT`
* **NEW** `GIT_REPOSITORY_NEW_OPTIONS_VERSION`
* **NEW** `GIT_ODB_OPTIONS_VERSION`
* **NEW** `GIT_ODB_BACKEND_PACK_OPTIONS_VERSION`
* **NEW** `GIT_ODB_BACKEND_LOOSE_OPTIONS_VERSION`
* **NEW** `GIT_DIFF_PARSE_OPTIONS_VERSION`

### Changes in structs 

* **NEW** `git_oid.type` field
* **CHANGED** `git_oid.id` max size is now `GIT_OID_MAX_SIZE`
* **NEW** `git_repository_init_options.oid_type` field
* **NEW** `git_index_options`
* **NEW** `git_repository_new_options`
* **NEW** `git_odb_options`
* **NEW** `git_odb_backend_pack_options`
[...]

PR #1206, opened 2026-01-12
feat: unstable SHA256 support

### LINKED ISSUE #1090: SHA-256
Consider the improvements regarding support for SHA-256 in Git, last mentioned in the release notes for version 2.45:

 - https://github.com/git/git/blob/master/Documentation/RelNotes/2.45.0.txt
 - https://github.blog/open-source/git/highlights-from-git-2-45/

libgit2 has experimental support for SHA-256, [guarded by `-DEXPERIMENTAL_SHA256`](https://github.com/libgit2/libgit2/discussions/5840#discussioncomment-10845971).

There is a reference to this in git2-rs:

https://github.com/rust-lang/git2-rs/blob/abbd64e80366b113b0a5056929ca3b58861e2570/libgit2-sys/lib.rs#L1174

It might be useful to allow users explore SHA-256 compatibility with git2-rs, probably behind a feature flag.

weihanglo:
This adds an `unstable-sha256` Cargo feature, as a follow-up of #1201.
Also adds smoke tests for affected operations/types.

Part of #1090

## Design

Introduce `ObjectFormat` accessors and `*_ext` API variants unconditionally,
The actual SHA256 wiring are gated behind the `unstable-sha256` Cargo feature
in order to keep public API additive.

The `*_ext` suffix follows libgit2's convention for extended function variants.
At next major version, the original methods may absorb the format parameter.
The actual change will leave for future.

## Insta-stable

- **NEW** `ObjectFormat` enum (`#[non_exhaustive]` when unstable-sha256 is not enabled)
- **NEW** `Oid::ZERO_SHA1` const
- **NEW** `Oid::object_format()` to access underlying OID type
- **NEW** `Oid::raw_bytes()` to return the full underlying byte buffer
- **NEW** `Repository::object_format()` to get hash algo on a repo
- **NEW** `Remote::object_format()` to get hash algo on a remote
- **NEW** `RepositoryInitOptions::object_format()` setter
- **NEW** `Oid::from_str_ext` / `Oid::hash_object_ext` / `Oid::hash_file_ext`
- **NEW** `Diff::from_buffer_ext`
- **NEW** `Index::new_ext` / `Index::open_ext`
- **NEW** `Indexer::new_ext`
- **NEW** `Odb::new_ext`
- **DEPRECATED** `Oid::zero()` in favor of `Oid::ZERO_SHA1`

## Behind `unstable-sha256`

- **NEW** `ObjectFormat::Sha256` enum variant
- **NEW** `Oid::ZERO_SHA256` const
- **CHANGED** All `_ext` methods dispatch to SHA256-aware FFI paths
- **CHANGED** `Oid::as_bytes` returns logical bytes (SHA1 -> 20; SHA256 -> 32)
- **CHANGED** `Oid::from_bytes` accepts 20 or 32 bytes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,9 +35,13 @@ jobs:
       shell: bash
     - run: cargo test --locked
     - run: cargo test --features https,ssh
+    # Single-threaded: libgit2's builtin SHA256 is not thread-safe.
+    # See https://github.com/rust-lang/git2-rs/issues/1255.
+    - run: cargo test --features unstable-sha256 -- --test-threads=1
+    - run: cargo test --features https,ssh,unstable-sha256
+    - run: cargo test -p git2-curl
     - run: cargo run -p systest
     - run: cargo run -p systest --features unstable-sha256
-    - run: cargo test -p git2-curl
 
   rustfmt:
     name: Rustfmt
--- a/examples/init.rs
+++ b/examples/init.rs
@@ -15,6 +15,7 @@
 #![deny(warnings)]
 
 use clap::Parser;
+use git2::ObjectFormat;
 use git2::{Error, Repository, RepositoryInitMode, RepositoryInitOptions};
 use std::path::{Path, PathBuf};
 
@@ -40,6 +41,9 @@ struct Args {
     #[structopt(name = "perms", long = "shared")]
     /// permissions to create the repository with
     flag_shared: Option<String>,
+    #[structopt(name = "object-format", long, value_parser = parse_object_format)]
+    /// object format to use (sha1 or sha256, requires unstable-sha256 feature to use the sha256 format)
+    flag_object_format: Option<ObjectFormat>,
 }
 
 fn run(args: &Args) -> Result<(), Error> {
@@ -48,6 +52,7 @@ fn run(args: &Args) -> Result<(), Error> {
         && args.flag_template.is_none()
         && args.flag_shared.is_none()
         && args.flag_separate_git_dir.is_none()
+        && args.flag_object_format.is_none()
     {
         Repository::init(&path)?
     } else {
@@ -68,6 +73,12 @@ fn run(args: &Args) -> Result<(), Error> {
         if let Some(ref s) = args.flag_shared {
             opts.mode(parse_shared(s)?);
         }
+
+        #[cfg(feature = "unstable-sha256")]
+        if let Some(format) = args.flag_object_format {
+            opts.object_format(format);
+        }
+
         Repository::init_opts(&path, &opts)?
     };
 
@@ -133,6 +144,15 @@ fn parse_shared(shared: &str) -> Result<RepositoryInitMode, Error> {
                 Err(Error::from_str("unknown value for --shared"))
             }
         }
+    }
+}
+
+fn parse_object_format(format: &str) -> Result<ObjectFormat, Error> {
+    match format {
+        "sha1" => Ok(ObjectFormat::Sha1),
+        #[cfg(feature = "unstable-sha256")]
+        "sha256" => Ok(ObjectFormat::Sha256),
+        _ => Err(Error::from_str("object format must be 'sha1' or 'sha256'")),
     }
 }
 
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -9,6 +9,7 @@ use std::ptr;
 use std::slice;
 
 use crate::util::{self, Binding};
+use crate::ObjectFormat;
 use crate::{panic, raw, Buf, Delta, DiffFormat, Error, FileMode, Oid, Repository};
 use crate::{DiffFlags, DiffStatsFormat, IntoCString};
 
@@ -310,16 +311,26 @@ impl Diff<'static> {
     /// two trees, however there may be subtle differences. For example,
     /// a patch file likely contains abbreviated object IDs, so the
     /// object IDs parsed by this function will also be abbreviated.
+    ///
+    /// This parses the diff assuming SHA1 object IDs. Use
+    /// [`Diff::from_buffer_ext`] to specify a different format.
     pub fn from_buffer(buffer: &[u8]) -> Result<Diff<'static>, Error> {
+        Self::from_buffer_ext(buffer, ObjectFormat::Sha1)
+    }
+
+    /// Reads the contents of a git patch file into a `git_diff` object,
+    /// with a specific object format.
+    ///
+    /// See [`Diff::from_buffer`] for more details.
+    pub fn from_buffer_ext(buffer: &[u8], format: ObjectFormat) -> Result<Diff<'static>, Error> {
         crate::init();
         let mut diff: *mut raw::git_diff = std::ptr::null_mut();
+        let data = buffer.as_ptr() as *const c_char;
+        let len = buffer.len();
         unsafe {
+            let _ = format;
             // NOTE: Doesn't depend on repo, so lifetime can be 'static
-            try_call!(raw::git_diff_from_buffer(
-                &mut diff,
-                buffer.as_ptr() as *const c_char,
-                buffer.len()
-            ));
+            try_call!(raw::git_diff_from_buffer(&mut diff, data, len));
             Ok(Diff::from_raw(diff))
         }
     }
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -327,10 +327,20 @@ impl Diff<'static> {
         let mut diff: *mut raw::git_diff = std::ptr::null_mut();
         let data = buffer.as_ptr() as *const c_char;
         let len = buffer.len();
+        // NOTE: Doesn't depend on repo, so lifetime can be 'static
         unsafe {
-            let _ = format;
-            // NOTE: Doesn't depend on repo, so lifetime can be 'static
-            try_call!(raw::git_diff_from_buffer(&mut diff, data, len));
+            #[cfg(not(feature = "unstable-sha256"))]
+            {
+                let _ = format;
+                try_call!(raw::git_diff_from_buffer(&mut diff, data, len));
+            }
+            #[cfg(feature = "unstable-sha256")]
+            {
+                let mut opts: raw::git_diff_parse_options = std::mem::zeroed();
+                opts.version = raw::GIT_DIFF_PARSE_OPTIONS_VERSION;
+                opts.oid_type = format.raw();
+                try_call!(raw::git_diff_from_buffer(&mut diff, data, len, &mut opts));
+            }
             Ok(Diff::from_raw(diff))
         }
     }
@@ -1563,6 +1573,8 @@ impl DiffPatchidOptions {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "unstable-sha256")]
+    use crate::Diff;
     use crate::{DiffLineType, DiffOptions, Oid, Signature, Time};
     use std::borrow::Borrow;
     use std::fs::File;
@@ -1868,5 +1880,38 @@ mod tests {
         );
 
         assert_eq!(result.unwrap_err().code(), crate::ErrorCode::User);
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn diff_sha256() {
+        let (_td, repo) = crate::test::repo_init_sha256();
+        let diff = repo.diff_tree_to_workdir(None, None).unwrap();
+        assert_eq!(diff.deltas().len(), 0);
+        let stats = diff.stats().unwrap();
+        assert_eq!(stats.insertions(), 0);
+        assert_eq!(stats.deletions(), 0);
+        assert_eq!(stats.files_changed(), 0);
+        let patchid = diff.patchid(None).unwrap();
+
+        // Verify SHA256 OID (32 bytes)
+        assert_eq!(patchid.as_bytes().len(), 32);
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn diff_from_buffer_sha256() {
+        // Minimal patch with SHA256 OID (64 chars)
+        let patch = b"diff --git a/file.txt b/file.txt
+index 0000000000000000000000000000000000000000000000000000000000000000..1111111111111111111111111111111111111111111111111111111111111111 100644
+--- a/file.txt
++++ b/file.txt
+@@ -1 +1 @@
+-old
++new
+";
+
+        let diff = Diff::from_buffer_ext(patch, crate::ObjectFormat::Sha256).unwrap();
+        assert_eq!(diff.deltas().len(), 1);
     }
 }
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -1591,7 +1591,7 @@ mod tests {
         assert_eq!(stats.deletions(), 0);
         assert_eq!(stats.files_changed(), 0);
         let patchid = diff.patchid(None).unwrap();
-        assert_ne!(patchid, Oid::zero());
+        assert_ne!(patchid, Oid::ZERO_SHA1);
     }
 
     #[test]
--- a/src/index.rs
+++ b/src/index.rs
@@ -869,7 +869,9 @@ mod tests {
     use std::path::Path;
     use tempfile::TempDir;
 
-    use crate::{ErrorCode, Index, IndexEntry, IndexTime, Oid, Repository, ResetType};
+    use crate::{
+        ErrorCode, Index, IndexEntry, IndexTime, ObjectFormat, Oid, Repository, ResetType,
+    };
 
     #[test]
     fn smoke() {
@@ -972,7 +974,7 @@ mod tests {
 
     #[test]
     fn add_then_read() {
-        let mut index = Index::new().unwrap();
+        let mut index = Index::new_ext(ObjectFormat::Sha1).unwrap();
         let mut e = entry();
         e.path = b"foobar".to_vec();
         index.add(&e).unwrap();
@@ -982,7 +984,7 @@ mod tests {
 
     #[test]
     fn add_then_find() {
-        let mut index = Index::new().unwrap();
+        let mut index = Index::new_ext(ObjectFormat::Sha1).unwrap();
         let mut e = entry();
         e.path = b"foo/bar".to_vec();
         index.add(&e).unwrap();
--- a/src/index.rs
+++ b/src/index.rs
@@ -9,6 +9,7 @@ use libc::{c_char, c_int, c_uint, c_void, size_t};
 
 use crate::util::{self, path_to_repo_path, Binding};
 use crate::IntoCString;
+use crate::ObjectFormat;
 use crate::{panic, raw, Error, IndexAddOption, IndexTime, Oid, Repository, Tree};
 
 /// A structure to represent a git [index][1]
@@ -90,10 +91,21 @@ impl Index {
     ///
     /// This index object cannot be read/written to the filesystem, but may be
     /// used to perform in-memory index operations.
+    ///
+    /// This always creates a SHA1 index.
+    /// Use [`Index::new_ext`] to create an index with a specific object format.
     pub fn new() -> Result<Index, Error> {
+        Self::new_ext(ObjectFormat::Sha1)
+    }
+
+    /// Creates a new in-memory index with a specific object format.
+    ///
+    /// See [`Index::new`] for more details.
+    pub fn new_ext(format: ObjectFormat) -> Result<Index, Error> {
         crate::init();
         let mut raw = ptr::null_mut();
         unsafe {
+            let _ = format;
             try_call!(raw::git_index_new(&mut raw));
             Ok(Binding::from_raw(raw))
         }
@@ -107,12 +119,23 @@ impl Index {
     ///
     /// If you need an index attached to a repository, use the `index()` method
     /// on `Repository`.
+    ///
+    /// This opens the index assuming SHA1 object format. Use
+    /// [`Index::open_ext`] to specify a different format.
     pub fn open(index_path: &Path) -> Result<Index, Error> {
+        Self::open_ext(index_path, ObjectFormat::Sha1)
+    }
+
+    /// Opens a Git index with a specific object format.
+    ///
+    /// See [`Index::open`] for more details.
+    pub fn open_ext(index_path: &Path, format: ObjectFormat) -> Result<Index, Error> {
         crate::init();
         let mut raw = ptr::null_mut();
         // Normal file path OK (does not need Windows conversion).
         let index_path = index_path.into_c_string()?;
         unsafe {
+            let _ = format;
             try_call!(raw::git_index_open(&mut raw, index_path));
             Ok(Binding::from_raw(raw))
         }
--- a/src/index.rs
+++ b/src/index.rs
@@ -1050,9 +1050,9 @@ mod tests {
             gid: 0,
             file_size: 0,
             #[cfg(not(feature = "unstable-sha256"))]
-            id: Oid::from_bytes(&[0; 20]).unwrap(),
+            id: Oid::ZERO_SHA1,
             #[cfg(feature = "unstable-sha256")]
-            id: Oid::from_bytes(&[0; 32]).unwrap(),
+            id: Oid::ZERO_SHA256,
             flags: 0,
             flags_extended: 0,
             path: Vec::new(),
--- a/src/index.rs
+++ b/src/index.rs
@@ -105,8 +105,18 @@ impl Index {
         crate::init();
         let mut raw = ptr::null_mut();
         unsafe {
-            let _ = format;
-            try_call!(raw::git_index_new(&mut raw));
+            #[cfg(not(feature = "unstable-sha256"))]
+            {
+                let _ = format;
+                try_call!(raw::git_index_new(&mut raw));
+            }
+            #[cfg(feature = "unstable-sha256")]
+            {
+                let mut opts: raw::git_index_options = std::mem::zeroed();
+                opts.version = raw::GIT_INDEX_OPTIONS_VERSION;
+                opts.oid_type = format.raw();
+                try_call!(raw::git_index_new(&mut raw, &opts));
+            }
             Ok(Binding::from_raw(raw))
         }
     }
@@ -135,8 +145,18 @@ impl Index {
         // Normal file path OK (does not need Windows conversion).
         let index_path = index_path.into_c_string()?;
         unsafe {
-            let _ = format;
-            try_call!(raw::git_index_open(&mut raw, index_path));
+            #[cfg(not(feature = "unstable-sha256"))]
+            {
+                let _ = format;
+                try_call!(raw::git_index_open(&mut raw, index_path));
+            }
+            #[cfg(feature = "unstable-sha256")]
+            {
+                let mut opts: raw::git_index_options = std::mem::zeroed();
+                opts.version = raw::GIT_INDEX_OPTIONS_VERSION;
+                opts.oid_type = format.raw();
+                try_call!(raw::git_index_open(&mut raw, index_path, &opts));
+            }
             Ok(Binding::from_raw(raw))
         }
     }
@@ -1029,10 +1049,38 @@ mod tests {
             uid: 0,
             gid: 0,
             file_size: 0,
+            #[cfg(not(feature = "unstable-sha256"))]
             id: Oid::from_bytes(&[0; 20]).unwrap(),
+            #[cfg(feature = "unstable-sha256")]
+            id: Oid::from_bytes(&[0; 32]).unwrap(),
             flags: 0,
             flags_extended: 0,
             path: Vec::new(),
         }
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn index_sha256() {
+        let (_td, repo) = crate::test::repo_init_sha256();
+        let mut index = repo.index().unwrap();
+
+        // Test opening with correct format
+        Index::open_ext(&repo.path().join("index"), ObjectFormat::Sha256).unwrap();
+
+        // Test basic operations with SHA256
+        index.clear().unwrap();
+        index.read(true).unwrap();
+        index.write().unwrap();
+        let tree_id = index.write_tree().unwrap();
+
+        // Verify OID is 32 bytes (SHA256)
+        assert_eq!(tree_id.as_bytes().len(), 32);
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn smoke_in_memory_index_sha256() {
+        let _index = Index::new_ext(ObjectFormat::Sha256).unwrap();
     }
 }
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -9,6 +9,7 @@ use libc::{c_char, c_int, c_uint, c_void, size_t};
 
 use crate::panic;
 use crate::util::Binding;
+use crate::ObjectFormat;
 use crate::{
     raw, Error, IndexerProgress, Mempack, Object, ObjectType, OdbLookupFlags, Oid, Progress,
 };
@@ -45,10 +46,22 @@ impl<'repo> Drop for Odb<'repo> {
 
 impl<'repo> Odb<'repo> {
     /// Creates an object database without any backends.
+    ///
+    /// This always creates a SHA1 object database.
+    /// Use [`Odb::new_ext`] to create one with a specific object format.
     pub fn new<'a>() -> Result<Odb<'a>, Error> {
+        Self::new_ext(ObjectFormat::Sha1)
+    }
+
+    /// Creates an object database without any backends,
+    /// with a specific object format.
+    ///
+    /// See [`Odb::new`] for more details.
+    pub fn new_ext<'a>(format: ObjectFormat) -> Result<Odb<'a>, Error> {
         crate::init();
         unsafe {
             let mut out = ptr::null_mut();
+            let _ = format;
             try_call!(raw::git_odb_new(&mut out));
             Ok(Odb::from_raw(out))
         }
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -648,7 +648,7 @@ mod tests {
         let db = repo.odb().unwrap();
         let id = db.write(ObjectType::Blob, &dat).unwrap();
         let id_prefix_str = &id.to_string()[0..10];
-        let id_prefix = Oid::from_str(id_prefix_str).unwrap();
+        let id_prefix = Oid::from_str_ext(id_prefix_str, repo.object_format()).unwrap();
         let found_oid = db.exists_prefix(id_prefix, 10).unwrap();
         assert_eq!(found_oid, id);
     }
--- a/src/odb.rs
+++ b/src/odb.rs
@@ -61,8 +61,18 @@ impl<'repo> Odb<'repo> {
         crate::init();
         unsafe {
             let mut out = ptr::null_mut();
-            let _ = format;
-            try_call!(raw::git_odb_new(&mut out));
+            #[cfg(not(feature = "unstable-sha256"))]
+            {
+                let _ = format;
+                try_call!(raw::git_odb_new(&mut out));
+            }
+            #[cfg(feature = "unstable-sha256")]
+            {
+                let mut opts: raw::git_odb_options = std::mem::zeroed();
+                opts.version = raw::GIT_ODB_OPTIONS_VERSION;
+                opts.oid_type = format.raw();
+                try_call!(raw::git_odb_new(&mut out, &opts));
+            }
             Ok(Odb::from_raw(out))
         }
     }
@@ -252,7 +262,10 @@ impl<'repo> Odb<'repo> {
     /// ```compile_fail
     /// use git2::Odb;
     /// let mempack = {
+    ///     #[cfg(not(feature = "unstable-sha256"))]
     ///     let odb = Odb::new().unwrap();
+    ///     #[cfg(feature = "unstable-sha256")]
+    ///     let odb = Odb::new_ext(git2::ObjectFormat::Sha1).unwrap();
     ///     odb.add_new_mempack_backend(1000).unwrap()
     /// };
     /// ```
--- a/src/oid.rs
+++ b/src/oid.rs
@@ -42,19 +42,35 @@ pub struct Oid {
 impl Oid {
     /// Parse a hex-formatted object id into an Oid structure.
     ///
+    /// This always parses as SHA1 (up to 40 hex characters). Use
+    /// [`Oid::from_str_ext`] to parse with a specific format.
+    ///
     /// # Errors
     ///
     /// Returns an error if the string is empty, is longer than 40 hex
     /// characters, or contains any non-hex characters.
     pub fn from_str(s: &str) -> Result<Oid, Error> {
+        Self::from_str_ext(s, ObjectFormat::Sha1)
+    }
+
+    /// Parses a hex-formatted object id with a specific object format.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the string is
+    ///
+    /// * is empty
+    /// * is longer than 40 hex with SHA1 object format
+    /// * is longer than 64 hex with SHA256 object format
+    /// * contains any non-hex characters
+    pub fn from_str_ext(s: &str, format: ObjectFormat) -> Result<Oid, Error> {
         crate::init();
         let mut raw = crate::util::zeroed_raw_oid();
+        let data = s.as_bytes().as_ptr() as *const libc::c_char;
+        let len = s.len() as libc::size_t;
         unsafe {
-            try_call!(raw::git_oid_fromstrn(
-                &mut raw,
-                s.as_bytes().as_ptr() as *const libc::c_char,
-                s.len() as libc::size_t
-            ));
+            let _ = format;
+            try_call!(raw::git_oid_fromstrn(&mut raw, data, len));
         }
         Ok(Oid { raw })
     }
@@ -86,17 +102,29 @@ impl Oid {
     /// Hashes the provided data as an object of the provided type, and returns
     /// an Oid corresponding to the result. This does not store the object
     /// inside any object database or repository.
+    ///
+    /// This always hashes using SHA1. Use [`Oid::hash_object_ext`]
+    /// to hash with a specific format.
     pub fn hash_object(kind: ObjectType, bytes: &[u8]) -> Result<Oid, Error> {
+        Self::hash_object_ext(kind, bytes, ObjectFormat::Sha1)
+    }
+
+    /// Hashes the provided data as an object of the provided type,
+    /// with a specific object format.
+    ///
+    /// See [`Oid::hash_object`] for more details.
+    pub fn hash_object_ext(
+        kind: ObjectType,
+        bytes: &[u8],
+        format: ObjectFormat,
+    ) -> Result<Oid, Error> {
         crate::init();
 
         let mut out = crate::util::zeroed_raw_oid();
+        let data = bytes.as_ptr() as *const libc::c_void;
         unsafe {
-            try_call!(raw::git_odb_hash(
-                &mut out,
-                bytes.as_ptr() as *const libc::c_void,
-                bytes.len(),
-                kind.raw()
-            ));
+            let _ = format;
+            try_call!(raw::git_odb_hash(&mut out, data, bytes.len(), kind.raw()));
         }
 
         Ok(Oid { raw: out })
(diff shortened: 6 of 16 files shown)


src/oid.rs
```diff
@@ -104,21 +154,36 @@ impl Oid {
...
     }
 
-    /// View this OID as a byte-slice 20 bytes in length.
+    /// View this OID as a byte-slice.
+    ///
+    /// * 20 bytes in length if the feature `unstable-sha256` is not enabled.
+    /// * 32 bytes in length if the feature `unstable-sha256` is enabled.
     pub fn as_bytes(&self) -> &[u8] {
         &self.raw.id
     }
```
jorendorff: I think this should return 20 bytes if `unstable-sha256` is enabled but the object format is SHA-1.

One reason for this is that crate features should be additive; turning on the feature shouldn't unnecessarily affect how the crate behaves with already-supported repositories.

weihanglo: Thanks for calling this out. That's a fair point, but I'm not sure "feature additivity" is the core issue here.

I think the real question is what `Oid::as_bytes()` is supposed to mean in this crate: the logical object ID bytes, or the raw backing storage from libgit2.

I'm leaning toward the logical interpretation. That seems like the less surprising API for callers, as you mentioned.

The reason I wasn’t fully convinced yet is that this type has historically been a fairly thin wrapper over libgit2, so I wanted to check whether anyone might rely on the raw representation or the backing buffer. Or maybe we should just mark this with a big ‼️ in the changelog, so we can transit to your proposal. Do you know any use of this functions in the wild?

lorenzleutgeb:
I believe to have found something that you missed in this PR. Note that `git2::util::zeroed_raw_oid` does not take any arguments:

https://github.com/rust-lang/git2-rs/blob/e6919b76ae08e2fc60291d414fab9af6a67feae3/src/util.rs#L271-L275

I believe that this means that the `kind: c_uchar` (where `c_uchar` is defined as `u8`) member of the returned `libgit2_sys::git_oid` will have the value `0u8`.

The function `git2::util::zeroed_raw_oid` is called by `git2::Oid::zero`, which also takes no arguments. If the function `git2::Oid::object_format` is called on the result of `git2::Oid::zero`, the call

https://github.com/rust-lang/git2-rs/blob/e6919b76ae08e2fc60291d414fab9af6a67feae3/src/oid.rs#L231

will hit the `panic!` at

https://github.com/rust-lang/git2-rs/blob/e6919b76ae08e2fc60291d414fab9af6a67feae3/src/oid.rs#L29

This is because `0u8 as u32` (which we got from `std::mem::zeroed`) is neither `GIT_OID_SHA1 = 1`, nor `GIT_OID_SHA256 = 2`.

I discovered this experimenting in code that is downstream of `git2`, building against this PR. In one of my tests I did essentially `git2::Oid::zero().object_format()`.

Then I started to think about how much `unsafe` really is necessary here. Maybe `zeroed_raw_oid` could look like

```rust
#[inline]
pub(crate) fn zeroed_raw_oid(format: ObjectFormat) -> raw::git_oid {
  raw::git_oid {
    kind: format.raw(),
    id: [0; raw::GIT_OID_MAX_SIZE],
  }
}
```

In fact one might even ask why not to have

```rust
impl Oid {
  pub const ZERO_SHA1: Self = Self {
    raw: raw::git_oid {
      kind: raw::GIT_OID_SHA1,
      id: [0; raw::GIT_OID_MAX_SIZE],
    }
  };
  
  pub const ZERO_SHA256: Self = Self {
    raw: raw::git_oid {
      kind: raw::GIT_OID_SHA256,
      id: [0; raw::GIT_OID_MAX_SIZE],
    }
  };
  
  // NOTE: Maybe call this `zero_with_object_format` to not break `zero`.
  pub fn zero(format: ObjectFormat) -> Self {
    // TODO: Match on `format` and return `Oid::ZERO_SHA1` or `Oid::ZERO_SHA256`.
  }
}
```

This way, users that know at compile time which kind of zero they want can just use the constant, while those that decide at run time would call `zero`. And there is no `unsafe` involved.

If you do not want `git2::Oid::zero` to take an argument for backwards compatibility, then you could have it return a SHA-1 object identifier always, and mark it deprecated in favor of these constants and a new `zero_with_object_format` (if I have not missed a reason why the constants cannot be used).

For `oid.rs` I also had the gut feeling that some `unsafe` could be removed, but I do not have any constructive comments here, sorry.

lorenzleutgeb:
Another comment regarding `FromStr`: I understand you opted to remove it to "avoid misuse", but I would argue that having `FromStr` actually is quite convenient. Instead of removing it, you could consider implementing `FromStr` in a way that it checks the length of the argument and then calls into the more specialized function that takes a string and an object format. The caller can then still inspect the object format of the result. You leave all options open for your users, and keep backwards compatibility.

src/index.rs
```diff
@@ -90,6 +90,23 @@ impl Index {
...
+    ///
+    /// In future releases, this will be removed entirely to avoid misuse.
+    ///
+    /// Consider these alternatives:
+    ///
+    /// * [`Index::with_object_format`] if an in-memory index is needed
+    /// * [`Repository::index`] if you have repository context
+    ///
+    /// </div>
+    #[cfg(not(feature = "unstable-sha256"))]
```
DanielEScherzer: why is this function removed when sha256 support is enabled? If the idea is that we should migrate to with_object_format() anyway, maybe add that function in a separate PR that doesn't need to wait for SHA-256 to be ready?

src/odb.rs
```diff
@@ -54,6 +65,25 @@ impl<'repo> Odb<'repo> {
         }
     }
 
+    /// Creates an object database without any backends, with the specified object format.
+    ///
+    /// The `format` parameter determines the OID type (SHA1 or SHA256) used by this object database.
+    ///
+    /// Note: Most users should obtain an ODB from [`Repository::odb`](crate::Repository::odb),
+    /// which automatically inherits the repository's object format.
+    #[cfg(feature = "unstable-sha256")]
```
DanielEScherzer: this can match Index and introduce the new method without the feature gate, so that code can start to use it without requiring the unstable feature

src/oid.rs
```diff
@@ -45,34 +52,68 @@ impl Oid {
     /// # Errors
     ///
     /// Returns an error if the string is empty, is longer than 40 hex
-    /// characters, or contains any non-hex characters.
-    pub fn from_str(s: &str) -> Result<Oid, Error> {
+    /// characters (or 64 for SHA256), or contains any non-hex characters.
+    pub fn from_str(
+        s: &str,
+        #[cfg(feature = "unstable-sha256")] format: crate::ObjectFormat,
```
DanielEScherzer: this parameter could also be ungated so that callers can start passing it

src/oid.rs
```diff
@@ -45,34 +52,68 @@ impl Oid {
...
         crate::init();
         let mut raw = crate::util::zeroed_raw_oid();
-        if bytes.len() != raw::GIT_OID_MAX_SIZE {
-            Err(Error::from_str("raw byte array must be 20 bytes"))
-        } else {
+
+        #[cfg(not(feature = "unstable-sha256"))]
+        {
+            if bytes.len() != raw::GIT_OID_SHA1_SIZE {
+                return Err(Error::from_str("raw byte array must be 20 bytes"));
```
DanielEScherzer: maybe a clearer error for trying to accidentally use 32 bytes when the feature isn't enabled?

Also, why have different ways of creating the SHA1 oid depending on if the SHA256 feature is available?

```rust
let oid_type = match bytes.len() {
    raw::GIT_OID_SHA1_SIZE => raw::GIT_OID_SHA1,
    #[cfg(feature = "unstable-sha256")]
    raw::GIT_OID_SHA256_SIZE => raw::GIT_OID_SHA256,
    #[cfg(not(feature = "unstable-sha256"))]
    raw::GIT_OID_SHA256_SIZE => {
        return Err(Error::from_str(
            "raw byte array of 32 bytes requires unstable-sha256 feature to be enabled",
        ))
    }
    _ => {
        return Err(Error::from_str(
            "raw byte array must be 20 bytes (SHA1) or 32 bytes (SHA256, if enabled)",
        ))
    }
};
unsafe {
    try_call!(raw::git_oid_from_raw(&mut raw, bytes.as_ptr(), oid_type));
}
Ok(Oid { raw })
```

src/oid.rs
```diff
@@ -85,16 +126,25 @@ impl Oid {
     /// Hashes the provided data as an object of the provided type, and returns
     /// an Oid corresponding to the result. This does not store the object
     /// inside any object database or repository.
-    pub fn hash_object(kind: ObjectType, bytes: &[u8]) -> Result<Oid, Error> {
+    pub fn hash_object(
+        kind: ObjectType,
+        bytes: &[u8],
+        #[cfg(feature = "unstable-sha256")] format: crate::ObjectFormat,
```
DanielEScherzer: another parameter that can already be added immediately - not going to call them out each time though

src/remote.rs
```diff
@@ -178,6 +187,27 @@ impl<'repo> Remote<'repo> {
...
+    /// remote repository. The format is available as soon as the connection to
+    /// the remote is initiated and it remains available after disconnecting.
+    pub fn object_format(&self) -> Result<ObjectFormat, Error> {
+        let mut oid_type = raw::GIT_OID_SHA1;
+        unsafe {
+            try_call!(raw::git_remote_oid_type(&mut oid_type, self.raw));
+        }
+        #[cfg(not(feature = "unstable-sha256"))]
+        {
+            let _ = oid_type;
```
DanielEScherzer: hmm, even if the git2-rs bindings don't have SHA256 enabled, if the oid_type corresponds to SHA256 we should say that and probably return an error, rather than just claiming that the repository is formatted with SHA1?

DanielEScherzer (commented):
just some thoughts, not going to pretend that I understand this fully

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/9a1b398881ef288866fd79fb0b37780eef20c6c2..e6919b76ae08e2fc60291d414fab9af6a67feae3)*

src/remote.rs
weihanglo: IIUC, without `unstable-sha256`, the underlying libgit2-sys can't return a SHA256 oid or even open a SHA256 repo.

src/oid.rs
weihanglo: Good suggestion, though I am not sure if here we should talk about unstable-sha256 feature, downstream consumer of git2-rs may just propagate error message to end-users, and that might be weird to see that.

weihanglo:
> Another comment regarding `FromStr`: I understand you opted to remove it to "avoid misuse", but I would argue that having `FromStr` actually is quite convenient. Instead of removing it, you could consider implementing `FromStr` in a way that it checks the length of the argument and then calls into the more specialized function that takes a string and an object format. The caller can then still inspect the object format of the result. You leave all options open for your users, and keep backwards compatibility.

This is still ambiguous for those not 41-64 char hex strings. I'd still consider deprecating it and eventually remove.
Alternatively, we just keep it there and have a compatibility note.

src/index.rs
weihanglo: Kept. See the latest revision

src/oid.rs
weihanglo: Kept. See the latest revision

src/oid.rs
weihanglo: Kept. See the latest revision

weihanglo (commented):
New revision is up.

The major difference is that we stop follow `libgit2@main` but 1.9.3. The other is API design, which is stated in the PR description:

> Introduce `ObjectFormat` accessors and `*_ext` API variants unconditionally,
> The actual SHA256 wiring are gated behind the `unstable-sha256` Cargo feature
> in order to keep public API additive.
> 
> The `*_ext` suffix follows libgit2's convention for extended function variants.
> At next major version, the original methods may absorb the format parameter.
> The actual change will leave for future.
> 



*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/a70dbf42bf029dafb8ed003a0bd5ce20544a3472..57782382e8aab16e561c469beeed3b728da8fb1d)*

weihanglo:
cc @ehuss

src/oid.rs
```diff
@@ -10,9 +10,13 @@ use crate::util::{c_cmp_to_ordering, Binding};
 
 /// Object ID format (hash algorithm).
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum ObjectFormat {
     /// SHA1 object format (20-byte object IDs)
     Sha1,
+    /// SHA256 object format (32-byte object IDs)
+    #[cfg(feature = "unstable-sha256")]
+    Sha256,
```
lorenzleutgeb, line 19: Consider using an [explicit discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#r-items.enum.discriminant.explicit) that matches with libgit2. Like

```suggestion
    Sha256 = raw::GIT_OID_SHA256,
```
or
```suggestion
    Sha256 = 2,
```

src/oid.rs
```diff
@@ -10,9 +10,13 @@ use crate::util::{c_cmp_to_ordering, Binding};
 
 /// Object ID format (hash algorithm).
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
```
lorenzleutgeb: I should have been more explicit when I suggested this in https://github.com/rust-lang/git2-rs/issues/1090#issuecomment-4305963865, sorry.

Adding `#[non_exhaustive]` breaks downstream crates, since their `match`es will fail to compile. Thus I think this is a breaking change. This might be completely fine, as this crate is 0.x anyway, but I think it's worth noting. That being said...

Adding `#[non_exhaustive]` even when `unstable-sha256` is enabled, might be "too much". This way you are preparing your users for whatever comes after SHA-256. Which might be your intention, and then that's fine. But you might as well consider that it might take another decade or two until that happens, and you could spare your dependents the `_ => unimplemented!("...")` arm. This would mean that adding the next object format is a breaking change. And that might be fine as well.

So one approach might be to have
```suggestion
#[cfg_attr(not(feature = "unstable-sha256"), non_exhaustive)]
```
now, and whenever SHA-256 becomes stable, to remove `non_exhaustive`.

This way, all dependents *not* enabling `unstable-sha256` will be forced to think about further object formats (which might make sense, since SHA-256 is already out in the wild), and everyone that opts in to `unstable-sha256` can enjoy that this `enum` is exhaustive. The migration (as SHA-256 stabilizes) also works out nicely: The `_ => unimplemented!("...")` arm nicely takes the spot of the `Sha256` variant, but since now the `non_exhaustive` attribute would be *removed* no third arm is required once the dependent decides to implement support for SHA-256.

src/oid.rs
```diff
@@ -10,9 +10,13 @@ use crate::util::{c_cmp_to_ordering, Binding};
...
 pub enum ObjectFormat {
     /// SHA1 object format (20-byte object IDs)
     Sha1,
```
lorenzleutgeb, line 16:
```suggestion
    Sha1 = raw::GIT_OID_SHA1,
```
or
```suggestion
    Sha1 = 1,
```
(See https://github.com/rust-lang/git2-rs/pull/1206/changes#r3207277482.)

src/oid.rs
```diff
@@ -65,36 +106,78 @@ impl Oid {
     pub fn from_bytes(bytes: &[u8]) -> Result<Oid, Error> {
         crate::init();
         let mut raw = crate::util::zeroed_raw_oid();
```
lorenzleutgeb, line 131: If `unstable-sha256` is enabled, `raw.kind` will be zero and thus not a valid object format!

lorenzleutgeb (changes requested):
I will double down on what I explained in https://github.com/rust-lang/git2-rs/pull/1206#issuecomment-4320277625: If the feature `unstable-sha256` is enabled, the implementation of the unsafe function `crate::util::zeroed_raw_oid` is incorrect. It allows to create OIDs in memory, which are invalid. Their `kind` of value zero does not correspond to a known `git2::ObjectFormat` and also neither to libgit2's `GIT_OID_SHA1` nor `GIT_OID_SHA256`. That is because the function uses `std::mem::zeroed` but the object formats [are non-zero bytes](
https://github.com/libgit2/libgit2/blob/1affb8b19346c4f90e163a9a0364959ff1410f64/include/git2/oid.h#L26-L27). This is a huge footgun. Please consider these options:
 1. Remove `crate::util::zeroed_raw_oid`.
 2. Remove the `unsafe` keyword from `crate::util::zeroed_raw_oid` and reimplement it without `unsafe` blocks. This will force you to add an argument `format: git2::ObjectFormat`, otherwise, if `unstable-sha256` is enabled, you will not be able to construct a legal zero OID.
 3. Use `#[cfg(not(feature = "unstable-sha256"))] crate::util::zeroed_raw_oid`.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/a70dbf42bf029dafb8ed003a0bd5ce20544a3472..57782382e8aab16e561c469beeed3b728da8fb1d)*

src/oid.rs
weihanglo: Then this is probably not worthy. We will anyway have another major version bump when stabilizing sha256 feature.

weihanglo: Think about it more, yeah this sounds good.

src/oid.rs
weihanglo: Thanks for the suggestion. Some of the pre-existing translation enums in git2-rs follow the plain enum + `Binding` + explicit match pattern. [`libgit2_sys::git_oid_t`](https://docs.rs/libgit2-sys/latest/libgit2_sys/type.git_oid_t.html) is a numeric type alias, and `ObjectFormat` is the safe abstraction over it. Explicit discriminants would definitly simplify `ObjectFormat::raw()`, though `from_raw` still needs the match for safety against invalid C values. I'd keep `ObjectFormat` as-is for now.

src/oid.rs
weihanglo: @lorenzleutgeb commented in <https://github.com/rust-lang/git2-rs/pull/1206#pullrequestreview-4250628098>

(inline comment is better for discussion)

> I will double down on what I explained in https://github.com/rust-lang/git2-rs/pull/1206#issuecomment-4320277625: If the feature `unstable-sha256` is enabled, the implementation of the unsafe function `crate::util::zeroed_raw_oid` is incorrect. It allows to create OIDs in memory, which are invalid. Their `kind` of value zero does not correspond to a known `git2::ObjectFormat` and also neither to libgit2's `GIT_OID_SHA1` nor `GIT_OID_SHA256`. That is because the function uses `std::mem::zeroed` but the object formats [are non-zero bytes](
> https://github.com/libgit2/libgit2/blob/1affb8b19346c4f90e163a9a0364959ff1410f64/include/git2/oid.h#L26-L27). This is a huge footgun. Please consider these options:
>  1. Remove `crate::util::zeroed_raw_oid`.
>  2. Remove the `unsafe` keyword from `crate::util::zeroed_raw_oid` and reimplement it without `unsafe` blocks. This will force you to add an argument `format: git2::ObjectFormat`, otherwise, if `unstable-sha256` is enabled, you will not be able to construct a legal zero OID.
>  3. Use `#[cfg(not(feature = "unstable-sha256"))] crate::util::zeroed_raw_oid`.

weihanglo: Thanks for the detailed follow-up. Will do a cleanup then, though let me clarify the current state:

- `Oid::zero()` is now deprecated and returns `Oid::ZERO_SHA1` always. The reasoning being high-level binding should not have this kind of zeroed struct as output param. If there is any probably an API design flaw. If people want oid placeholder, they probalby already know the object format so can choice either const. Tthanks for the const suggestion!
- `zeroed_raw_oid()` is `pub(crate)`. Every internal call site uses it as an output parameter that libgit2 overwrites before any read. Agreed it may become an internal footgun if misused.
- FWIW, libgit2 itself uses uninitialized `git_oid` as output buffers in lots of places. Like [`maybe_sha_or_abbrev`](https://github.com/libgit2/libgit2/blob/1affb8b19346c4f90e163a9a0364959ff1410f64/src/libgit2/revparse.c#L24-L29): `git_oid oid;` is declared uninitialized, then filled by `git_oid__fromstrn(&oid, ...)` before any read.




Though thanks again, removing one more unsafe is a good thing! Fixed in https://github.com/rust-lang/git2-rs/pull/1206/changes/8c17e75db79413f5f32583506d064ca89ea059b6

The only one concern is that if libgit2 somehow start checking `kind` for output parameter cases, it might become buggy. Though that will be considered an upstream libgit2 bug.

src/oid.rs
weihanglo: Fixed and squashed into <https://github.com/rust-lang/git2-rs/pull/1206/changes/17d0f510292624a45b26a52223191a0f5ef3f4c4>. Thanks again for the forward-looking suggestion!

src/oid.rs
```diff
@@ -10,9 +10,13 @@ use crate::util::{c_cmp_to_ordering, Binding};
 
 /// Object ID format (hash algorithm).
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(not(feature = "unstable-sha256"), non_exhaustive)]
 pub enum ObjectFormat {
```
lorenzleutgeb, line 14: Downstream I just attempted to print an `ObjectFormat` and realized that it does not `impl Display`. Perhaps that might be a useful addition?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/remote.rs
DanielEScherzer: What if the underlying library was built with sha256 enabled, but the git2 crate wasn't?

weihanglo: AFAIK none of the mainstream packagers distribute `libgit2.so` with SHA256 support. The reason being it was broken before libgit2 1.9.3 (released a week ago). Also when sha256 is enabled, libgit2 built a shared libray with `-experimental` suffix. Unless packager intentionally patch that build script to remove the suffix, it is pretty distinguishable. However, it is unlikely someone does that, since the two are ABI incompatible and doing that breaks other programs dynamically linking to it.

I also raised this concern in <https://github.com/rust-lang/cargo/pull/16939#discussion_r3209423129> but from the other angle – Cargo may need to build from source always as nobody ships `libgit2-experimental.so`.

examples/init.rs
```diff
@@ -40,6 +41,9 @@ struct Args {
     #[structopt(name = "perms", long = "shared")]
     /// permissions to create the repository with
     flag_shared: Option<String>,
+    #[structopt(name = "object-format", long, value_parser = parse_object_format)]
+    /// object format to use (sha1 or sha256, requires unstable-sha256 feature)
```
DanielEScherzer:
```suggestion
    /// object format to use (sha1 or sha256, requires unstable-sha256 feature to use the sha256 format)
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/diff.rs
```diff
@@ -310,16 +310,38 @@ impl Diff<'static> {
     /// two trees, however there may be subtle differences. For example,
     /// a patch file likely contains abbreviated object IDs, so the
     /// object IDs parsed by this function will also be abbreviated.
+    ///
+    /// This parses the diff assuming SHA1 object IDs. Use
+    /// [`Diff::from_buffer_ext`] to specify a different format.
     pub fn from_buffer(buffer: &[u8]) -> Result<Diff<'static>, Error> {
+        Self::from_buffer_ext(buffer, crate::ObjectFormat::Sha1)
```
DanielEScherzer: I can't make suggestions on the unmodified line, but I suggest adding an import for `crate::ObjectFormat` rather than needing to fully qualify it

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/diff.rs
```diff
@@ -310,16 +310,38 @@ impl Diff<'static> {
...
+    pub fn from_buffer_ext(
+        buffer: &[u8],
+        format: crate::ObjectFormat,
+    ) -> Result<Diff<'static>, Error> {
         crate::init();
         let mut diff: *mut raw::git_diff = std::ptr::null_mut();
+        let data = buffer.as_ptr() as *const c_char;
+        let len = buffer.len();
         unsafe {
-            // NOTE: Doesn't depend on repo, so lifetime can be 'static
```
DanielEScherzer, line 317: this explanation of why things can be 'static is lost

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/index.rs
```diff
@@ -90,11 +90,32 @@ impl Index {
     ///
     /// This index object cannot be read/written to the filesystem, but may be
     /// used to perform in-memory index operations.
+    ///
+    /// This always creates a SHA1 index.
+    /// Use [`Index::new_ext`] to create an index with a specific object format.
     pub fn new() -> Result<Index, Error> {
+        Self::new_ext(crate::ObjectFormat::Sha1)
```
DanielEScherzer: same suggestion to add a `use` statement for the ObjectFormat, also applies in some other files, not going to keep repeating the suggestion

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
```diff
@@ -40,21 +48,54 @@ pub struct Oid {
...
     /// Returns an error if the string is empty, is longer than 40 hex
     /// characters, or contains any non-hex characters.
     pub fn from_str(s: &str) -> Result<Oid, Error> {
+        Self::from_str_ext(s, ObjectFormat::Sha1)
+    }
+
+    /// Parses a hex-formatted object id
+    /// with a specific object format.
+    ///
+    /// See [`Oid::from_str`] for more details.
```
DanielEScherzer: except `::from_str` documents that it is an error if the string is longer than 40 hex characters, and presumably for this new method the maximum length is format-dependent, so this method should have some extra documentation

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
DanielEScherzer: If the error message is propagated to end users, presumably the end users had some kind of say in the input value

src/oid.rs
```diff
@@ -104,29 +187,90 @@ impl Oid {
...
+                rpath,
+                kind.raw(),
+                format.raw()
+            ));
         }
 
         Ok(Oid { raw: out })
     }
 
     /// View this OID as a byte-slice 20 bytes in length.
```
DanielEScherzer: can we document that this is going to truncate sha256 ids?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
```diff
@@ -104,29 +187,90 @@ impl Oid {
...
+                raw::GIT_OID_SHA256 => raw::GIT_OID_SHA256_SIZE,
+                _ => panic!("Unknown git oid type"),
+            };
+            &self.raw.id[..size]
+        }
+    }
+
+    /// View the full underlying byte buffer of this OID.
+    ///
+    /// Currently equivalent to [`Oid::as_bytes`].
```
DanielEScherzer: This is a bin confusing - the "Currently equivalent" assumes that the feature is not enabled, but then the next line discusses that it could be enabled. Maybe

```suggestion
    /// For SHA1 hashes, this is equivalent to [`Oid::as_bytes`].
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
```diff
@@ -212,12 +361,15 @@ mod tests {
...
 
+    use libgit2_sys as raw;
+
     use super::Error;
     use super::Oid;
     use crate::ObjectType;
     use tempfile::TempDir;
 
     #[test]
+    #[cfg(not(feature = "unstable-sha256"))]
```
DanielEScherzer: so I noticed this on a few other files too - why are various tests disabled when using unstable-sha256? If they would be failing, then that is indicative of a BC break, and if they would be passing, then we should presumably keep the tests enabled?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
```diff
@@ -251,23 +480,188 @@ mod tests {
...
     }
 
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn comparisons_object_format() -> Result<(), Error> {
+        use crate::ObjectFormat;
+
+        // SHA1 OID comparisons with explicit format
+        assert_eq!(
+            Oid::from_str_ext("decbf2b", ObjectFormat::Sha1)?,
```
DanielEScherzer: similarly, for the *_ext methods when using Sha1, those parts can also be tested even when sha256 support is not enabled

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

.github/workflows/main.yml
```diff
@@ -33,10 +33,11 @@ jobs:
     - name: Install Rust (rustup)
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
-    - run: cargo test --locked
-    - run: cargo test --features https,ssh
     - run: cargo run -p systest
     - run: cargo run -p systest --features unstable-sha256
+    - run: cargo test --locked
```
ehuss, line 36: Moving these breaks the `--locked` check since the earlier `cargo run` will update the lockfile. Can you either move them back, or add `--locked` to the other commands?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

.github/workflows/main.yml
```diff
@@ -33,10 +33,11 @@ jobs:
...
+    - run: cargo test --features https,ssh
+    - run: cargo test --features unstable-sha256
```
ehuss: Can we also make sure this covers networking (and examples)? Maybe something like:


```suggestion
    - run: cargo test --features https,ssh,unstable-sha256
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

ehuss:
Note: Git 3 will switch to reftable by default ([source](https://git-scm.com/docs/BreakingChanges)). Reftable support was only just merged in https://github.com/libgit2/libgit2/pull/7117 which presumably will be in libgit2 2.0. SHA256 support is a major step towards being compatible with git3, but reftable will be another large piece that will be required. Fortunately I don't think that requires any API changes, though we should probably plan to do some testing ahead of the libgit2 2.0 release.

src/index.rs
```diff
@@ -1004,10 +1048,38 @@ mod tests {
             uid: 0,
             gid: 0,
             file_size: 0,
+            #[cfg(not(feature = "unstable-sha256"))]
             id: Oid::from_bytes(&[0; 20]).unwrap(),
+            #[cfg(feature = "unstable-sha256")]
+            id: Oid::from_bytes(&[0; 32]).unwrap(),
```
ehuss: Just a suggestion, take it or leave it:


```suggestion
            #[cfg(not(feature = "unstable-sha256"))]
            id: Oid::ZERO_SHA1,
            #[cfg(feature = "unstable-sha256")]
            id: Oid::ZERO_SHA256,
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/index.rs
```diff
@@ -1004,10 +1048,38 @@ mod tests {
...
+        index.write().unwrap();
+        let tree_id = index.write_tree().unwrap();
+
+        // Verify OID is 32 bytes (SHA256)
+        assert_eq!(tree_id.as_bytes().len(), 32);
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-sha256")]
+    fn smooke_in_memory_index_sha256() {
```
ehuss:
```suggestion
    fn smoke_in_memory_index_sha256() {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/oid.rs
ehuss: I'm not sure I understand, why would this truncate?

I would agree that the doc comment should probably be updated, though, to mention it can be 32 bytes for sha256.

src/oid.rs
```diff
@@ -104,29 +187,90 @@ impl Oid {
...
+            };
+            &self.raw.id[..size]
+        }
+    }
+
+    /// View the full underlying byte buffer of this OID.
+    ///
+    /// Currently equivalent to [`Oid::as_bytes`].
+    /// With the `unstable-sha256` feature, these two methods will differ.
+    pub fn raw_bytes(&self) -> &[u8] {
```
ehuss: Why was this method added? I don't see it used anywhere.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/test.rs
```diff
@@ -37,6 +37,28 @@ pub fn repo_init() -> (TempDir, Repository) {
     (td, repo)
 }
 
+#[cfg(feature = "unstable-sha256")]
+pub fn repo_init_sha256() -> (TempDir, Repository) {
```
ehuss, line 46: Instead of duplicating the code here, can we merge them somehow?

I might suggest having a `repo_init_ext` that takes an ObjectFormat as a parameter, and rewrite `repo_init` to just call `repo_init_ext(ObjectFormat::Sha1)`. 

Or any other approach, just try to remove the duplicate code.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/8aeb3efda95cad45ee000c473029582288422e16..79d5e2bb8cf1249316d11a43185a72d1786e65fd)*

src/index.rs
lorenzleutgeb: I suggested these constants, so that will surprise no one, but I second that suggestion. More readable and less `unwrap`ing!

src/oid.rs
lorenzleutgeb: https://github.com/rust-lang/git2-rs/pull/1206#discussion_r2910396322

src/oid.rs
weihanglo: Doc comment need an update. We don't need to truncate anything here.

weihanglo:
I believe every review comment is addressed. Let me know there is something needing updates.

src/oid.rs
DanielEScherzer: bump - still applies

src/index.rs
```diff
@@ -846,11 +889,13 @@ mod tests {
...
     use tempfile::TempDir;
 
-    use crate::{ErrorCode, Index, IndexEntry, IndexTime, Oid, Repository, ResetType};
+    use crate::{
+        ErrorCode, Index, IndexEntry, IndexTime, ObjectFormat, Oid, Repository, ResetType,
+    };
 
     #[test]
     fn smoke() {
-        let mut index = Index::new().unwrap();
```
DanielEScherzer, line 853: can you keep a test case that ::new() uses Sha1 ?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

src/oid.rs
```diff
@@ -40,21 +67,60 @@ pub struct Oid {
...
+
+    /// Parses a hex-formatted object id with a specific object format.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the string is
+    ///
+    /// * is empty
+    /// * is longer than 40 hex with SHA1 object format
+    /// * is longer than 64 hex with SHA1 object format
```
DanielEScherzer:
```suggestion
    /// * is longer than 64 hex with SHA256 object format
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

src/oid.rs
```diff
@@ -65,36 +131,84 @@ impl Oid {
...
         crate::init();
         let mut raw = crate::util::zeroed_raw_oid();
-        if bytes.len() != raw::GIT_OID_MAX_SIZE {
-            Err(Error::from_str("raw byte array must be 20 bytes"))
-        } else {
+
+        #[cfg(not(feature = "unstable-sha256"))]
+        {
+            if bytes.len() != raw::GIT_OID_SHA1_SIZE {
+                return Err(Error::new(
```
DanielEScherzer: I think this can be simplified with `Error::from_str`, same for the error created below

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

src/oid.rs
```diff
@@ -104,29 +218,91 @@ impl Oid {
...
+            &self.raw.id
+        }
+        #[cfg(feature = "unstable-sha256")]
+        {
+            let size = match self.raw.kind as raw::git_oid_t {
+                raw::GIT_OID_SHA1 => raw::GIT_OID_SHA1_SIZE,
+                raw::GIT_OID_SHA256 => raw::GIT_OID_SHA256_SIZE,
+                _ => panic!("Unknown git oid type"),
+            };
+            &self.raw.id[..size]
```
DanielEScherzer, line 271: so I was going to suggest adding a comment that when unstable-sha256 is enabled the id field is based on GIT_OID_MAX_SIZE and can be bigger than needed, but it is *always* based on GIT_OID_MAX_SIZE. The documentation currently states that it will be a specific size - can we add some assertions to validate the documentation?

```rust
#[cfg(not(feature = "unstable-sha256"))]
const _: () = assert_eq!(20, raw::GIT_OID_MAX_SIZE);
#[cfg(feature = "unstable-sha256")]
const _: () = {
    assert_eq!(20, raw::GIT_OID_SHA1_SIZE);
    assert_eq!(32, raw::GIT_OID_SHA256_SIZE);
    assert_eq!(32, raw::GIT_OID_MAX_SIZE);
}
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

src/oid.rs
```diff
@@ -104,29 +218,91 @@ impl Oid {
...
+        }
+    }
+
+    /// View the full underlying byte buffer of this OID.
+    ///
+    /// * 20 bytes in length if the feature `unstable-sha256` is not enabled.
+    /// * 32 bytes in length if the feature `unstable-sha256` is enabled.
```
DanielEScherzer:
```suggestion
    /// * 32 bytes in length if the feature `unstable-sha256` is enabled, even if the OID is SHA1.
```

also we should add assertions about these sizes are correct in terms of how big `raw::GIT_OID_MAX_SIZE` is

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

src/util.rs
```diff
@@ -268,10 +268,18 @@ fn fixup_windows_path(path: CString) -> Result<CString, Error> {
...
 }
 
-/// Creates a zeroed git_oid structure.
+/// Creates a zeroed `git_oid` to be used as an output buffer
+/// that libgit2 overwrites before any read.
+///
+/// Defaults `kind` to `GIT_OID_SHA1` so the value is a valid
+/// `git_oid_t` even transiently.
 #[inline]
 pub(crate) fn zeroed_raw_oid() -> raw::git_oid {
```
DanielEScherzer, line 277: this is now the same as just using `Oid::ZERO_SHA1` - maybe it should just return that constant, and in a follow-up this function be removed (since it is only `pub(crate)`)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/0d1fde9993a8a544b8ac66414576d032dacd32e3..cf9b64b1713624c1d5bd50090fd8e24ee69e22b2)*

DanielEScherzer:
I don't seem to be able to mark my previous reports as resolved, GitHub reports "Failed to resolve thread" - it could be because I don't have write access, or because the bot edited my comments, not sure. Anyway, other than the one report that I added a new comment to, the rest are all indeed resolved.

src/util.rs
weihanglo: `Binding::raw()` returns a `*const raw::git_oid`, and a bit annoying to deal with. I may leave this as a follow-up as it doesn't really affect any public interface I believe.

weihanglo:
@DanielEScherzer most of your review comments are addressed. Thanks!

src/oid.rs
DanielEScherzer: was done, thanks (don't have the rights to resolve this)

src/oid.rs
DanielEScherzer: looks like this was done, thanks (don't have the rights to resolve this)

src/index.rs
DanielEScherzer: was done, thanks (don't have the rights to resolve this)

src/oid.rs
DanielEScherzer: was done, thanks (don't have the rights to resolve this)

src/oid.rs
DanielEScherzer: was done, thanks (don't have the rights to resolve this)

src/oid.rs
DanielEScherzer: was done, thanks (don't have the rights to resolve this)

src/oid.rs
```diff
@@ -207,11 +387,25 @@ impl AsRef<[u8]> for Oid {
     }
 }
 
+// Sanity-check libgit2's OID size constants to keep the docs and the
```
DanielEScherzer: can we use a bit more inclusive language here?
```suggestion
// Double-check libgit2's OID size constants to keep the docs and the
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/f63d8f54e56f5d6bb5382de9fbd169d38fc44504..371036b15aa3f17e8a72d7ca929eb19a7d94fa30)*

src/oid.rs
```diff
@@ -225,6 +419,76 @@ mod tests {
         assert!(Oid::from_bytes(b"00000000000000000000").is_ok());
     }
 
+    #[test]
+    fn conversions_ext_sha1() {
+        use crate::ObjectFormat;
```
DanielEScherzer: instead of adding these `use`s to some tests, but using `crate::ObjectFormat::*` in others (object_format_always_sha1), maybe just add it to the overall `mod tests`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/f63d8f54e56f5d6bb5382de9fbd169d38fc44504..371036b15aa3f17e8a72d7ca929eb19a7d94fa30)*

src/oid.rs
weihanglo: Removed as requested. We can always add it later if there is a use case.

ehuss (approved):
Thanks!!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1206/e9854c6f1acdabb93be2c131d92c91931700c727..7bd145ee23884f19d1d79ce0e46c1ab105784040)*

PR #1210, opened 2026-01-25
Add function: merge_file

chalharu:
Some overlap with https://github.com/rust-lang/git2-rs/pull/635 but that PR has been open for years.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,9 +185,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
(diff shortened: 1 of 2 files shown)


src/merge.rs
```diff
@@ -1,3 +1,4 @@
+use libc::{c_char, size_t};
```
ehuss: Can you combine this with the other libc import?

chalharu:
Thanks for the review — the changes are now updated.

PR #1213, opened 2026-01-31
fix: check ptr nullity before calling from_raw_parts

### LINKED ISSUE #1211: Potential null pointer dereference with `Deref` implementation on `Buf`
We consider that there should be a null check before dereference on `Buf`.

https://github.com/rust-lang/git2-rs/blob/03ec58122907ba45fc9f9ba7721a7e62be1f719e/src/buf.rs#L23-L49
In this code, if we dereference the buffer right after calling `new()`, it pass null pointer to the unsafe function `slice::from_raw_parts`. Based on the documentation of function,

> data must be non-null and aligned even for zero-length slices or slices of ZSTs.

Even though it is zero-length slice, we should ensure that the pointer is not null. 

How do you think about it?

weihanglo:
Fixes #1211

### DIFF
--- a/src/buf.rs
+++ b/src/buf.rs
@@ -44,12 +44,18 @@ impl Buf {
 impl Deref for Buf {
     type Target = [u8];
     fn deref(&self) -> &[u8] {
+        if self.raw.ptr.is_null() {
+            return &[];
+        }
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks, I pushed up a small test, too.

PR #1230, opened 2026-04-13
Add `Revspec::into_objects()` and tests

### LINKED ISSUE #1147: Add Revspec::into_objects
It is currently not possible to extract the `Object`s from a `Revspec` without cloning, which is wasteful, since these are heap objects.

If this was a simple struct, you could just move out of the fields, but because it needs to be accessed through getter methods, this is not possible, and thus a separate method is needed.

```rust
impl<'repo> Revspec<'repo> {
  // RevparseMode is `Copy`, so it doesn't need to be returned in destructuring,
  // although maybe it should be for symmetry.
  fn into_objects(self) -> (Option<Object<'repo>>,  Option<Object<'repo>>);
}
```

This would be useful for when you want to collect all the commits into a `Vec`.

DanielEScherzer:
Closes #1147

### DIFF
--- a/src/revspec.rs
+++ b/src/revspec.rs
@@ -20,7 +20,7 @@ impl<'repo> Revspec<'repo> {
     /// Convert a Revspec into the underlying components
     ///
     /// If only references to `from` or `to` are needed, use the corresponding
-    /// [Self::from()] and [Self::to()] methods.
+    /// [`Revspec::from()`] and [`Revspec::to()`] methods.
     pub fn into_objects(self) -> (Option<Object<'repo>>, Option<Object<'repo>>, RevparseMode) {
         (self.from, self.to, self.mode)
     }
--- a/src/revspec.rs
+++ b/src/revspec.rs
@@ -17,6 +17,14 @@ impl<'repo> Revspec<'repo> {
         Revspec { from, to, mode }
     }
 
+    /// Convert a Revspec into the underlying components
+    ///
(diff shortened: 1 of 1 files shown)


src/revspec.rs
```diff
@@ -17,6 +17,14 @@ impl<'repo> Revspec<'repo> {
         Revspec { from, to, mode }
     }
 
+    /// Convert a Revspec into the underlying components
+    ///
+    /// If only references to `from` or `to` are needed, use the corresponding
+    /// [Self::from()] and [Self::to()] methods.
```
weihanglo: For myself, better to spell it out then `Self`.

```suggestion
    /// [`Revspec::from()`] and [`Revspec::to()`] methods.
```

weihanglo (approved):
Just one small nit. Thank you for the work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1230/ce4dc01328e837db9ac63e68cc084337851a92be..370c0c35f711269c488a15e988c8d1b5eec85d5e)*

src/revspec.rs
DanielEScherzer: done

DanielEScherzer:
Hmm, I guess there was a race condition in the test from #1237 ?

Will send a patch in a minute to just remove that test, it isn't really important

DanielEScherzer:
rebased on top of the removal of the flaky test

weihanglo:
Thanks!

PR #1232, opened 2026-04-13
Clarify `CheckoutBuilder::update_index()` documentation

### LINKED ISSUE #967: Documentation for update_index should be tweaked
Current documentation says this:

```rust
/// Prevents checkout from writing the updated files' information to the
/// index.
///
/// Defaults to true.
pub fn update_index(&mut self, update: bool) -> &mut CheckoutBuilder<'cb>;
```

It's not completely clear when reading this whether setting this flag to true enables writing to the index or inhibits it (though the function and param names point towards the correct interpretation).

A better doc comment might be

> **When set to false**, prevents checkout from writing the updated files' information to the index.

DanielEScherzer:
Closes #967

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -418,8 +418,8 @@ impl<'cb> CheckoutBuilder<'cb> {
         self.flag(raw::GIT_CHECKOUT_UPDATE_ONLY, update)
     }
 
-    /// Prevents checkout from writing the updated files' information to the
-    /// index.
+    /// When set to false, prevents checkout from writing the updated files'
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1232/ce4dc01328e837db9ac63e68cc084337851a92be..7bea3b5788f2097cbcdcaa3849eea38d13f5b0fd)*

Issue #851, opened 2022-06-30
Should `Reference` implement `PartialEq`/`Eq`?


nasso:
Currently, `Reference` has an implementation for `Eq` (and `PartialEq`) that uses `git_reference_cmp` under the hood to determine if two references are "equal".

However, `git_reference_cmp` may return `0` ("yes they're equal") when comparing two symbolic references that have different names but the same target. See libgit2/libgit2#6337 for further details, and [`refs.c:1069`](https://github.com/libgit2/libgit2/blob/3847522e86e9c65be674f1372cefefdbfbe9ba2b/src/libgit2/refs.c#L1069) for the current implementation of `git_reference_cmp`, effectively comparing the *targets* of symbolic references instead of the references themselves.

This leads to the following behaviour:

```sh
# create the refs (they can also be created using git2, but using the git CLI here for simplicity)
git symbolic-ref refs/foo refs/heads/master
git symbolic-ref refs/bar refs/heads/master
```
```rs
let foo = repo.find_reference("refs/foo")?;
let bar = repo.find_reference("refs/bar")?;

assert_ne!(foo.name(), bar.name()); // ok, "refs/foo" != "refs/bar" (disregarding the `Option` here)
assert_ne!(foo, bar); // assertion fails! despite being *different* references, they share the same target
```

While it can make sense in some cases to have `git_reference_cmp` behave the way it does, I'm not sure if this is expected for an implementation of `PartialEq`, and especially so for an implementation of `Eq`. It's easy to overlook, especially since I couldn't find any mention of this in the documentation (of both `git2-rs` or `libgit2`).

Maybe `git2-rs` shouldn't translate `git_reference_cmp` to a `PartialEq`/`Eq` implementation? Maybe `libgit2` should change the behaviour of `git_reference_cmp`? Maybe nothing should change and the documentation should just make this more clear? I'm not too sure what's best here, but this has definitely confused me more than once.

nasso:
For the record, libgit2 2.0 is right around the corner, and it "fixes" `git_reference_cmp` to compare the references themselves instead of just their type + target!

See: https://github.com/libgit2/libgit2/pull/6346

PR #1234, opened 2026-04-14
Allow comparing `Branch` objects

### LINKED ISSUE #422: Implement Eq and Ord for Branch?
Since `Reference` is `Eq` and `Ord`, should `Branch` be also?

I saw the discussion and conclusion in #153, which might apply here too, but figured `Branch` might be a little different since it's only a wrapper around a `Reference`. I volunteer to write a PR if this seems like a good idea.

DanielEScherzer:
Add implementations for `PartialOrd`, `Ord`, `PartialEq`, and `Eq`. The actual is delegated to the underlying `Reference` object.

Closes #422

### DIFF
--- a/src/branch.rs
+++ b/src/branch.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use std::ffi::CString;
 use std::marker;
 use std::ptr;
@@ -124,6 +125,26 @@ impl<'repo> Branch<'repo> {
     }
 }
 
+impl<'repo> PartialOrd for Branch<'repo> {
+    fn partial_cmp(&self, other: &Branch<'repo>) -> Option<cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl<'repo> Ord for Branch<'repo> {
+    fn cmp(&self, other: &Branch<'repo>) -> cmp::Ordering {
+        self.inner.cmp(&other.inner)
+    }
+}
+
+impl<'repo> PartialEq for Branch<'repo> {
+    fn eq(&self, other: &Branch<'repo>) -> bool {
+        self.inner == other.inner
+    }
+}
+
+impl<'repo> Eq for Branch<'repo> {}
+
 impl<'repo> Branches<'repo> {
     /// Creates a new iterator from the raw pointer given.
     ///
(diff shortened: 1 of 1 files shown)


src/branch.rs
weihanglo, line 1: Haven't yet checked by myself. Does libgit2 have or plan to have an compare functions for branch or references?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1234/f497fc4602dee9e7204bb08933e5815df42a6114..4a68e3299e0f56f673ed6e99287821091f99e714)*

DanielEScherzer: It allows comparing references with `git_reference_cmp()` which is what `impl<'repo> Ord for Reference<'repo>` uses

src/branch.rs
```diff
@@ -173,7 +194,9 @@ mod tests {
 
         let mut b1 = repo.branch("foo", &commit, false).unwrap();
         assert!(!b1.is_head());
-        repo.branch("foo2", &commit, false).unwrap();
+        let b2 = repo.branch("foo2", &commit, false).unwrap();
```
weihanglo, line 197: Have we considered the equality here be based on branch name instead of `Reference` equality? It is a bit weird and too implicit when two different branch names differ comparsion says there are the same.

Perhaps we should be more explicit on the comparsion, and don't implement the eq operator?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1234/f497fc4602dee9e7204bb08933e5815df42a6114..4a68e3299e0f56f673ed6e99287821091f99e714)*

DanielEScherzer: What if you had two different references (lowercase) to the same branch, but the underlying `Reference`s are not the same - I would be surprised if those two branches are considered equal

weihanglo: Agree with that one, too.

Semantically a reference is composed of target and name and oid. However upstream libgit2 has a bug that didn't consider them when comparing. They fixed that in <https://github.com/libgit2/libgit2/pull/6346> but won't release it until libgit2 v2.

I think until then we might not want to tie to one about-to-change semantic. What do you think?

(<https://github.com/rust-lang/git2-rs/issues/851> has done a good research btw)

DanielEScherzer: No objections to waiting, but do we know when libgit2 v2 will come out?

weihanglo: I have no idea to be honest. See
 
* https://github.com/libgit2/libgit2/discussions/7149
* https://github.com/libgit2/libgit2/discussions/6988

weihanglo:
@rustbot label +S-blocked +upstream

wait for upstream change in libgit2 v2 probably

PR #1238, opened 2026-04-22
[Clippy] Address `clippy::unnecessary_cast` in production code

DanielEScherzer:
In a lot of places, the lint is triggered when casting a constant from the libgit2 bindings that is created (in libgit2-sys) via `git_enum!`. That macro will use i32 when `#[cfg(target_env = "msvc")]`, and u32 otherwise, meaning that a number of the casts that clippy flags as "unnecessary" are indeed necessary. In those cases, use `#[allow]` with a reason.

Similarly, clippy also lints against casting of `c_int` and `c_uint` to `i32` and `u32` - while those are usually the same types, the libc documentation makes it clear that they are not always the same. In those cases too, use `#[allow]` with a reason.

The only libc-provided type with a consistent rust type is that `size_t` is always the same as `usize` - fix the places that clippy identified where that conversion is unneeded. Other fixes include casts when the underlying type is explicitly set in the libgit2 bindings to be a specific rust type, or when converting mutable pointers to the same type they already are.

### DIFF
--- a/src/apply.rs
+++ b/src/apply.rs
@@ -102,7 +102,10 @@ impl<'cb> ApplyOptions<'cb> {
     }
 
     fn flag(&mut self, opt: raw::git_apply_flags_t, val: bool) -> &mut Self {
-        #[allow(clippy::unnecessary_cast, reason = "u32 unless compiling for msvc target env")]
+        #[allow(
+            clippy::unnecessary_cast,
+            reason = "u32 unless compiling for msvc target env"
+        )]
         let opt = opt as u32;
         if val {
             self.raw.flags |= opt;
--- a/src/apply.rs
+++ b/src/apply.rs
@@ -102,6 +102,7 @@ impl<'cb> ApplyOptions<'cb> {
     }
 
     fn flag(&mut self, opt: raw::git_apply_flags_t, val: bool) -> &mut Self {
+        #[allow(clippy::unnecessary_cast, reason = "u32 unless compiling for msvc target env")]
         let opt = opt as u32;
         if val {
             self.raw.flags |= opt;
--- a/src/blame.rs
+++ b/src/blame.rs
@@ -209,7 +209,7 @@ impl<'blame> BlameHunk<'blame> {
 
     /// Returns number of lines in this hunk.
     pub fn lines_in_hunk(&self) -> usize {
-        unsafe { (*self.raw).lines_in_hunk as usize }
+        unsafe { (*self.raw).lines_in_hunk }
     }
 
     /// Get the short "summary" of the git commit message for the hunk.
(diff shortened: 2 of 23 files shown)


DanielEScherzer:
This is the first of a few PRs to get clippy passing, and then I'll send a PR to add it to CI

src/apply.rs
weihanglo, line 1: A couple of thoughts:

* What is our MSRV? Edition 2021 is 1.56 I remembered, and lint reason was introduced along with `#[expect]` which is 1.81.
* This adds lots of visual noise. Not sure if there are other good way to improve this, like, should we just suppress for all?
* We don't have clippy in CI yet. Should probably add one in a separate PR.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1238/382f1e8a1ec07c1ec98c418e0d62db9b1f8e8177..66ebc7d5ab19bb241249fda3ef738a9e859e5486)*

DanielEScherzer: > What is our MSRV? Edition 2021 is 1.56 I remembered, and lint reason was introduced along with #[expect] which is 1.81.

I guess I had assumed that CI tested against the MSRV

> This adds lots of visual noise. Not sure if there are other good way to improve this, like, should we just suppress for all?

Some of the places that it caught things were indeed unneeded casts that I removed, so I think it is useful to have

> We don't have clippy in CI yet. Should probably add one in a separate PR.

I wanted to get it passing first, see my comment above that

> This is the first of a few PRs to get clippy passing, and then I'll send a PR to add it to CI

weihanglo: > Some of the places that it caught things were indeed unneeded casts that I removed, so I think it is useful to have

Indeed, though I would still argue that we probably don't want `allow` everywhere to keep the code slimmer. Alternatively, we might want to run this under Windows so we know where we need it where we dont.

weihanglo, line 1: https://github.com/rust-lang/git2-rs/pull/1238#issuecomment-4293451785

> This is the first of a few PRs to get clippy passing, and then I'll send a PR to add it to CI

The other way to do this. We integrate CI and allow the most noisy ones first, and then fix up and remove the allow as we go one by one. Though I think we should figure out MSRV first as we have yet set any.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1238/382f1e8a1ec07c1ec98c418e0d62db9b1f8e8177..66ebc7d5ab19bb241249fda3ef738a9e859e5486)*

DanielEScherzer: CI was added separately for clippy and does run things under Windows as well

PR #1239, opened 2026-04-22
`ReferenceNames`: Support non-UTF8 branch names

### LINKED ISSUE #1175: ReferenceNames panics if there is a non-UTF-8 reference
`git2::ReferenceNames` is an iterator that lists the names of references in the repository.  The documentation observes that `ReferenceNames` is more efficient compared to `References`, since `ReferenceNames` outputs a `&str` rather than a `Reference`.

The problem is that there might be references whose names aren't valid UTF-8 and therefore can't be represented as `str`.  If such a reference is found, `next` will panic (`str::from_utf8(bytes).unwrap()`, at `reference.rs` line 498.)

For example, if I create a repository like this:
```
git init foo
cd foo
git branch -m $'\300'
git commit --allow-empty -m whoops
```

Then the following program will crash when invoked in `foo`:
```
fn main() {
    let repo = git2::Repository::open_from_env().expect("open_from_env");
    let mut refs = repo.references().expect("references");
    for name in refs.names() {
        println!("name: {name:?}");
    }
}
```

Instead, the iterator probably ought to produce an error value in this case.  I don't know what that error should be, since it's not a libgit2 error.

DanielEScherzer:
Rather than using `Result::unwrap()` on the result of `str::from_utf8()`, handle failures by converting the `Utf8Error` into a `git2::Error`; the iterator for reference names already returns `Result`s that could be failures, so the signature is unchanged.

Add some end-to-end tests to confirm the results.

Fixes #1175

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,7 +3,7 @@ use std::env::JoinPathsError;
 use std::error;
 use std::ffi::{CStr, CString, NulError};
 use std::fmt;
-use std::str;
+use std::str::{self, Utf8Error};
 
 use crate::{raw, ErrorClass, ErrorCode};
 
@@ -389,6 +389,12 @@ impl From<NulError> for Error {
             "data contained a nul byte that could not be \
              represented as a string",
         )
+    }
+}
+
+impl From<Utf8Error> for Error {
+    fn from(e: Utf8Error) -> Error {
+        Error::from_str(&e.to_string())
     }
 }
 
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -509,7 +509,10 @@ impl<'repo, 'references> Iterator for ReferenceNames<'repo, 'references> {
         unsafe {
             try_call_iter!(raw::git_reference_next_name(&mut out, self.inner.raw));
             let bytes = crate::opt_bytes(self, out).unwrap();
-            let s = str::from_utf8(bytes).unwrap();
+            let s = match str::from_utf8(bytes) {
+                Ok(s) => s,
+                Err(e) => return Some(Err(e.into())),
+            };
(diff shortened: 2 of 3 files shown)


DanielEScherzer:
Looks like git won't let you create such a branch on windows or macos (though on Windows I could run `git branch -m $'\300'`, the resulting branch name had unicode replacement characters)

Anyway, if this looks good I'll rebase and squash it, we don't need all of these commits in the history

weihanglo (commented):
This is reasonable. Could you clean up the commit history? No need to strictly follow but this commit structure style is useful: <https://epage.github.io/dev/pr-style/#c-atomic>

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1239/382f1e8a1ec07c1ec98c418e0d62db9b1f8e8177..09c77b4a3da42a6327b2f276cabfed4d9ba25ed4)*

DanielEScherzer:
> This is reasonable. Could you clean up the commit history? No need to strictly follow but this commit structure style is useful: https://epage.github.io/dev/pr-style/#c-atomic
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1239/382f1e8a1ec07c1ec98c418e0d62db9b1f8e8177..09c77b4a3da42a6327b2f276cabfed4d9ba25ed4)_

Squashed it

weihanglo (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1239/9a1b398881ef288866fd79fb0b37780eef20c6c2..2027edb9e712ccc980c148b7e7fe3d0551aebac5)*

PR #1241, opened 2026-04-27
Return `Result` values rather than converting errors to `None`

### LINKED ISSUE #857: Return `Result<&str, Utf8Error>` instead of `Option<&str>`
I don't understand why many methods return `Option<&str>` instead of `Result<&str, Utf8Error>` when decoding a string.

It hides information that could be used by the caller to form a more meaningful error message, and doesn't prevent anyone from just calling `unwrap`, `expect` or `ok` when they don't care about the error.

Furthermore, it would disambiguate the return value of some methods, such as `Reference::symbolic_target`, for which [`None` may be returned if the reference is either not symbolic or not a valid utf-8 string](https://docs.rs/git2/0.14.4/git2/struct.Reference.html#method.symbolic_target). If that method returned `Result<Option<&str>, Utf8Error>` (or `Option<Result<&str, Utf8Error>>`, though one can switch between the two with `(Option|Result)::transpose`), it would make it easier to disambiguate with a single function call.

I couldn't find any past discussion regarding this API design choice, so I'm assuming nobody ever made that suggestion because nobody encountered this issue before. Despite being a breaking change, I doubt this would significantly hurt anyone's codebase.

DanielEScherzer:
Fixes #857

### DIFF
--- a/examples/cat-file.rs
+++ b/examples/cat-file.rs
@@ -100,7 +100,7 @@ fn show_tag(tag: &Tag) {
     println!("tag {}", tag.name().unwrap());
     show_sig("tagger", tag.tagger());
 
-    if let Some(msg) = tag.message() {
+    if let Ok(Some(msg)) = tag.message() {
         println!("\n{}", msg);
     }
 }
--- a/examples/cat-file.rs
+++ b/examples/cat-file.rs
@@ -89,7 +89,7 @@ fn show_commit(commit: &Commit) {
     }
     show_sig("author", Some(commit.author()));
     show_sig("committer", Some(commit.committer()));
-    if let Some(msg) = commit.message() {
+    if let Ok(msg) = commit.message() {
         println!("\n{}", msg);
     }
 }
--- a/examples/log.rs
+++ b/examples/log.rs
@@ -176,7 +176,7 @@ fn run(args: &Args) -> Result<(), Error> {
             if !sig_matches(&commit.committer(), &args.flag_committer) {
                 return None;
             }
-            if !log_message_matches(commit.message(), &args.flag_grep) {
+            if !log_message_matches(commit.message().ok(), &args.flag_grep) {
                 return None;
             }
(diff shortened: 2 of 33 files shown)


DanielEScherzer:
Now that #1239 has been merged, we should probably return `git2::Error` on failure rather than directly return Utf8Error

examples/cat-file.rs
weihanglo, line 1: This is nice, though also non-utf8 path is rare. I guess we might want to bundle this with libgit2 v2 change, as that would be a larger release having more API changes. What do you think?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1241/9257016899c9e4b35e064012d264dbb424fa4d0f..ba8cb9058449666960822c8c33802e5b6bf8ad3e)*

DanielEScherzer: I think that there is a benefit to doing this now, so that callers can know what the error was instead of just getting back an opaque None, but ultimately it isn't my call. I just want to say that without a timeline for libgit2 v2 I wouldn't want to delay this indefinitely

weihanglo (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1241/9257016899c9e4b35e064012d264dbb424fa4d0f..ba8cb9058449666960822c8c33802e5b6bf8ad3e)*

examples/cat-file.rs
weihanglo: Yeah reasonable. Fine making some breaking changes now than later. Though still want to point how non-utf8 is not that common.

weihanglo:
This may need a rebase onto main branch

DanielEScherzer:
Did a rebase with `cargo test` after each commit, rebase applied cleanly but needed to update the new tests that I added in #1254, #1244, and #1251

weihanglo (approved):

PR #1246, opened 2026-05-07
Add end-to-end test showing stash count

### LINKED ISSUE #564: Introduce stash_count(&self) -> Result<usize, Error>
Would it be possible to introduce something like this? I'm not sure if the goals of this project are to keep it mapping to the libgit2 directly or not. I'm also not sure if allowing not requiring a mutable reference here is sound.

```rust
/// Count number of git stashes
pub fn stash_count(&self) -> Result<usize, Error> {
    unsafe {
        let mut count = 0;
        let mut data = StashCbData {
            callback: &mut |_, _, _| {
                count += 1;
                true
            },
        };
        let cb: raw::git_stash_cb = Some(stash_cb);
        try_call!(raw::git_stash_foreach(
            self.raw(),
            cb,
            &mut data as *mut _ as *mut _
        ));
        Ok(count)
    }
}
```

DanielEScherzer:
Closes #564

### DIFF
--- a/tests/end_to_end.rs
+++ b/tests/end_to_end.rs
@@ -1,8 +1,9 @@
 //! Tests for some end-to-end logic about certain operations
-use git2::{Error, Repository, RepositoryInitOptions};
+use git2::{Error, Repository, RepositoryInitOptions, StashFlags};
 
 use libgit2_sys as raw;
 use std::ffi::{CString, OsString};
+use std::fs;
 use std::ptr;
 
 use tempfile::TempDir;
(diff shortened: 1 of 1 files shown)


weihanglo (approved):
Nice. Thanks you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1246/8efdb2597cd1affff712bc0905974d18424a89a7..a78125a6634582bfcc391e0afb2b0d16204b1e54)*

PR #1250, opened 2026-05-12
Fix `Remote::list()` with an empty list

DanielEScherzer:
Fixes #1217

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -922,6 +922,27 @@ mod tests {
     }
 
     #[test]
+    fn empty_remote_list() {
+        // Regression tests for issue #1217
+        let td = TempDir::new().unwrap();
+        let repo = Repository::init(td.path()).unwrap();
+
+        let remote_dir = TempDir::new().unwrap();
+        let _remote_repo = Repository::init_bare(remote_dir.path()).unwrap();
+        let remote_url = if cfg!(unix) {
+            format!("file://{}", remote_dir.path().display())
+        } else {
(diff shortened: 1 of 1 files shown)


src/remote.rs
```diff
@@ -921,6 +933,27 @@ mod tests {
         drop(origin.clone());
     }
 
+    #[test]
```
weihanglo, line 936: Is it possible to follow the style here?

https://epage.github.io/dev/pr-style/#c-test

* First commits documenting the buggy behavior (test passes and commit is bisectable)
* second commit showing the behavior change in test diff

weihanglo (approved):
Thanks!

Though I think we have a bunch of them 😮‍💨

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1250/8aeb3efda95cad45ee000c473029582288422e16..241e62928ac45c8593b4b385f09e78e9aeba703d)*

src/remote.rs
DanielEScherzer: in the initial commit, in debug builds the test panics and in release builds it passes. Thus, in release builds there is no behavior change in the diff. I could add `#[cfg_attr(debug_assertions, should_panic)]` but I'm not sure that that would be really helpful (and also, not fully sure that debug_assertions is the right condition)

weihanglo: Yeah that does not matter much. Merging

PR #1251, opened 2026-05-12
Add end-to-end tests showing use of `Repository::statuses()`

### LINKED ISSUE #1145: `repo.state()` returns `Clean` when actually dirty
`git status --porcelain` shows diff but running `repo.state()` returns `RepositoryState::Clean`

DanielEScherzer:
Specifically how it analyses files and should be used to determine if a repo is "clean".

Closes #1145

### DIFF
--- a/tests/end_to_end.rs
+++ b/tests/end_to_end.rs
@@ -1,9 +1,12 @@
 //! Tests for some end-to-end logic about certain operations
-use git2::{Error, ReferenceType, Repository, RepositoryInitOptions, StashFlags};
+use git2::{
+    Error, ReferenceType, Repository, RepositoryInitOptions, StashFlags, Status, StatusOptions,
+};
 
 use libgit2_sys as raw;
 use std::ffi::{CString, OsString};
 use std::fs;
+use std::path::Path;
 use std::ptr;
 
 use tempfile::TempDir;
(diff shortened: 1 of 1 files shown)


tests/end_to_end.rs
```diff
@@ -238,3 +241,228 @@ fn branch_name_on_init() {
...
+    fs::write(path.join("IndexNew"), "IndexNew").unwrap();
+    fs::remove_file(path.join("IndexDeleted")).unwrap();
+    fs::remove_file(path.join("IndexTypechange")).unwrap();
+
+    fn create_symlink(to: &Path, from: &Path) {
+        #[cfg(unix)]
+        std::os::unix::fs::symlink(to, from).unwrap();
+
+        #[cfg(windows)]
+        std::os::windows::fs::symlink_file(to, from).unwrap();
```
weihanglo, line 303: On windows this requires developer mode privileges btw. If we have already done this, fine, if not, better make this skippable locally and enforced in CI.

weihanglo (commented):
While I don't think this really resolve the confusion of #1145, we at least have a file can point at when confusion arises. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1251/8aeb3efda95cad45ee000c473029582288422e16..47dd883b20ff974a9fa5616895d3ff439fe152e0)*

tests/end_to_end.rs
DanielEScherzer: We don't appear to have done this already - not sure how to make it skippable locally and enforced in CI, even though I use a windows machine I do all of my development in docker containers that run linux.

weihanglo: Check out Cargo: https://github.com/rust-lang/cargo/blob/9a624d6147d11b3f23507c02366071e31daea4b8/crates/cargo-test-support/src/lib.rs#L1577-L1601

though that is fine. We can leave it for future :)

weihanglo (approved):

PR #1257, opened 2026-05-17
Change connected() to take &self instead of &mut self

cuishuang:
The connected method only reads the connection status without mutating state, so &self is sufficient and avoids unnecessary mutable borrow.

### DIFF
--- a/src/remote.rs
+++ b/src/remote.rs
@@ -225,7 +225,7 @@ impl<'repo> Remote<'repo> {
     }
 
     /// Check whether the remote is connected
-    pub fn connected(&mut self) -> bool {
+    pub fn connected(&self) -> bool {
         unsafe { raw::git_remote_connected(self.raw) == 1 }
     }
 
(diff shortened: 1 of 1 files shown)


DanielEScherzer (approved):
Looking at the various git_remote_connected callbacks in libgit2 there are none that mutate the remote, so looks good to me (once CI passes)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1257/e9854c6f1acdabb93be2c131d92c91931700c727..868d99fdd64ac0a884cb44bfed6ee2d771fff53c)*

cuishuang:
> Looking at the various git_remote_connected callbacks in libgit2 there are none that mutate the remote, so looks good to me (once CI passes)
> 
> _[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1257/e9854c6f1acdabb93be2c131d92c91931700c727..868d99fdd64ac0a884cb44bfed6ee2d771fff53c)_

Thanks.

I resubmitted the changes and modified the test cases to pass the CI

ehuss:
Is this safe even with a custom transport? The transport vtable does not seem to enforce this as const or have any comments or contracts around that.

cuishuang:
> Is this safe even with a custom transport? The transport vtable does not seem to enforce this as const or have any comments or contracts around that.

Oh, I missed the custom transport case. Even though `git_remote_connected` takes a `const git_remote *`, the transport vtable's `is_connected` callback takes a mutable `git_transport *`, and there does not seem to be a contract requiring it to be logically const.

PR #1265, opened 2026-05-25
[Result migration] Update documentation of `StringArray::iter()`

DanielEScherzer:
Missed in the initial updates, the iterated type is now `Result<Option<&str>, Error>` rather than `Option<&str>`.


DanielEScherzer:
Going a different direction

PR #1266, opened 2026-05-26
[Result migration] Don't return options for `StringArray` iterators

### LINKED ISSUE #1263: New StringArray Iterator API should not include Option
In https://github.com/rust-lang/git2-rs/pull/1241, the `StringArray` iterator was changed from returning `Option<&str>` to `Result<Option<&str>>`. I think it should instead have been changed to `Result<&str>`. The inner `Option` is unnecessary because it should only ever be None if there is a bug in the iterator implementation. This results in an unnecessary unwrap for the caller. I think the iterator should be:

```rust
    type Item = Result<&'a str, Error>;
    fn next(&mut self) -> Option<Result<&'a str, Error>> {
        self.range.next().map(|i| self.arr.get(i).transpose().unwrap())
    }
```

cc @weihanglo @DanielEScherzer

DanielEScherzer:
Instead, treat a `None` as indicative of a bug in the code, and panic with a clear message that something went wrong.

Also update the documentation of `StringArray::iter()` to reflect the correct iterated type.

Fixes #1263

### DIFF
--- a/examples/tag.rs
+++ b/examples/tag.rs
@@ -65,7 +65,7 @@ fn run(args: &Args) -> Result<(), Error> {
     } else if args.flag_list {
         let pattern = args.arg_pattern.as_ref().map(|s| &s[..]).unwrap_or("*");
         for name in repo.tag_names(Some(pattern))?.iter() {
-            let name = name.expect("Not invalid utf8").expect("Not None");
+            let name = name.expect("Not invalid utf8");
             let obj = repo.revparse_single(name)?;
 
             if let Some(tag) = obj.as_tag() {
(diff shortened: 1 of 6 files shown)


ehuss (approved):
Thanks! I appreciate it!

I was going to wait on semver breaking changes, but since https://github.com/rust-lang/git2-rs/pull/1274 was merged I suppose we can go ahead and cycle another version bump.


*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1266/d6acaa56ac1d0c86991883387f9e1c03f28b72d5..599df867a184aa1ce7b62ad8cad381174de90312)*

PR #1267, opened 2026-05-29
Don't crash if the matched pathspec is NULL

ehuss:
In the index add_all callback, if the matched pathspec is NULL, this was crashing due to the panic in the C callback.

The temporary solution here is to handle the panic inside the panic wrapper. A future version will change the signature of the callback to handle this situation.

cc https://github.com/rust-lang/git2-rs/issues/585


ehuss:
cc https://github.com/rust-lang/git2-rs/pull/1268 which has the real fix.

weihanglo (approved):

PR #1268, opened 2026-05-29
Change IndexMatchedPath callback to take an Option 

### LINKED ISSUE #585: Segfault in `Index::add_all()` with empty `pathspecs` slice and `Some(cb)`

ehuss:
The matched pathspec can sometimes be None (if there was no match). This was crashing because the code was panicking in the C callback without being within the panic wrapper. The fix here is to properly handle the case where the matched_pathspec is NULL.

Fixes https://github.com/rust-lang/git2-rs/issues/585

Opening as a draft since this is a semver-breaking change. We'll need to decide when to slot this in.

### DIFF
--- a/examples/add.rs
+++ b/examples/add.rs
@@ -38,7 +38,7 @@ fn run(args: &Args) -> Result<(), git2::Error> {
     let repo = Repository::open(&Path::new("."))?;
     let mut index = repo.index()?;
 
-    let cb = &mut |path: &Path, _matched_spec: &[u8]| -> i32 {
+    let cb = &mut |path: &Path, _matched_spec: Option<&[u8]>| -> i32 {
         let status = repo.status_file(path).unwrap();
 
         let ret = if status.contains(git2::Status::WT_MODIFIED)
--- a/src/index.rs
+++ b/src/index.rs
@@ -48,7 +48,7 @@ pub struct IndexConflict {
 /// Used by `Index::{add_all,remove_all,update_all}`.  The first argument is the
 /// path, and the second is the pathspec that matched it.  Return 0 to confirm
 /// the operation on the item, > 0 to skip the item, and < 0 to abort the scan.
-pub type IndexMatchedPath<'a> = dyn FnMut(&Path, &[u8]) -> i32 + 'a;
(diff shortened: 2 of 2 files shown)


DanielEScherzer:
> The matched pathspec can sometimes be None (if there was no match).

Is there a reason to not just use an empty slice for the missing pathspec, instead of converting to an option?

ehuss:
Hm, that's an interesting question. libgit2 treats an empty set of pathspecs (`[]`) the same as an empty pathspec (`[""]` or any element that is an empty string) (which means to match everything).

git has forbidden empty pathspecs since 2.16 (from 2018), but it looks like libgit2 hasn't followed that change.

One could argue that getting an empty pathspec isn't quite the same as an empty set of pathspecs, but since they are treated the same (and you can't distinguish between them), it seems reasonable to me.

The main risk is if libgit2 ever decides that the empty string is no longer allowed, but an empty set is. I'm not sure if that would ever happen.

ehuss:
Thanks, closing with https://github.com/rust-lang/git2-rs/pull/1273 as a replacement.

PR #1270, opened 2026-06-01
make raw public

wiiznokes:


DanielEScherzer:
You can already depend on the underlying sys crate manually, why is a re-export needed?

wiiznokes:
to no depend on two different version accidentally

weihanglo:
> to no depend on two different version accidentally

It is unlikely to happen. Cargo disallows duplicate `link` in the same crate graph, and libgit2-sys has `link = "git2"`.

https://github.com/rust-lang/git2-rs/blob/d689d20a12daeea3b93249d7c6c059e8dce85765/libgit2-sys/Cargo.toml#L5

wiiznokes:
I didn't know that, but i still think it would be more convenient to export it

PR #1273, opened 2026-06-08
Fix IndexMatchedPath callback to not panic on empty pathspec

ehuss:
When given an empty set of pathspecs or a set with an empty string is used, the matched pathspec in the IndexMatchedPath ends up being null. Instead of panicking, use an empty slice to mean the "all" pathspec.

git has forbidden empty pathspecs since 2.16 (from 2018), but it looks like libgit2 hasn't followed that change.

One could argue that getting an empty pathspec isn't quite the same as an empty set of pathspecs, but since they are treated the same (and you can't distinguish between them), it seems reasonable to me to just use an empty slice to mean the same.

The main risk is if libgit2 ever decides that the empty string is no longer allowed, but an empty set is. I'm not sure if that would ever happen.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -48,6 +48,10 @@ pub struct IndexConflict {
 /// Used by `Index::{add_all,remove_all,update_all}`.  The first argument is the
 /// path, and the second is the pathspec that matched it.  Return 0 to confirm
 /// the operation on the item, > 0 to skip the item, and < 0 to abort the scan.
+///
+/// An empty pathspec means that either no pathspecs were specified or one of
+/// them was the empty string, both of which mean "match all paths" in
+/// libgit2.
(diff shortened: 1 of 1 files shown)


DanielEScherzer (approved):
makes sense to me

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1273/d689d20a12daeea3b93249d7c6c059e8dce85765..4f2e641316e31a00373f29f053ec60817e53aa34)*

weihanglo (approved):
Looks reasonable 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1273/d689d20a12daeea3b93249d7c6c059e8dce85765..4f2e641316e31a00373f29f053ec60817e53aa34)*

PR #1274, opened 2026-06-10
Convert UTF-8 errors to errors rather than None

DanielEScherzer:


src/repo.rs
```diff
@@ -1114,9 +1114,11 @@ impl Repository {
         name: &str,
         flags: AttrCheckFlags,
     ) -> Result<Option<&str>, Error> {
-        Ok(self
-            .get_attr_bytes(path, name, flags)?
-            .and_then(|a| str::from_utf8(a).ok()))
+        // Use match instead of .map() so that ? can short circuit properly
```
weihanglo: I don't think the comment is necessary as it is straightforward

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1274/7415113423d3b852e3527a62a97a8505ad9a7fe5..3ba6911b1641db10fedb66cee50f5b74cd7c8fad)*

weihanglo (approved):

src/repo.rs
DanielEScherzer: removed

PR #1277, opened 2026-06-11
Declare MSRV as 1.87, add tests for it

### LINKED ISSUE #1276: Identify/decide minimum supported version of Rust
C.f. #1238 

If I understand correctly:

https://github.com/rust-lang/git2-rs/blob/7415113423d3b852e3527a62a97a8505ad9a7fe5/Cargo.toml#L16 means that MSRV >= 1.56.0

https://github.com/rust-lang/git2-rs/blob/7415113423d3b852e3527a62a97a8505ad9a7fe5/Cargo.toml#L20 means MSRV >= 1.56.0 (https://docs.rs/crate/bitflags/2.1.0/source/Cargo.toml#55)

https://github.com/rust-lang/git2-rs/blob/7415113423d3b852e3527a62a97a8505ad9a7fe5/Cargo.lock#L474-L476 means MSRV >= 1.63.0 (https://docs.rs/crate/libc/0.2.174/source/Cargo.toml#52)

https://github.com/rust-lang/git2-rs/blob/7415113423d3b852e3527a62a97a8505ad9a7fe5/src/cred.rs#L301 means MSRV >= 1.70.0 (https://blog.rust-lang.org/2023/06/01/Rust-1.70.0/) (unless we want to say that there is a different MSRV depending on which features are enabled?)

https://github.com/rust-lang/git2-rs/blob/7415113423d3b852e3527a62a97a8505ad9a7fe5/src/opts.rs#L527 means *tests* are MSRV >= 1.70.0 for the same reason

DanielEScherzer:
Closes #1276

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,15 +37,7 @@ jobs:
     - name: Install Rust (rustup)
       run: rustup update ${{ matrix.rust }} --no-self-update && rustup default ${{ matrix.rust }}
       shell: bash
-    - name: Cargo test (locked)
-      if: ${{ matrix.build != 'MSRV' }}
-      run: cargo test --locked
-    - name: Remove lock file
-      if: ${{ matrix.build == 'MSRV' }}
-      run: rm Cargo.lock
-    - name: Cargo test (unlocked)
-      if: ${{ matrix.build == 'MSRV' }}
-      run: cargo test
+    - run: cargo test --locked
     - run: cargo test --features https,ssh
     # Single-threaded: libgit2's builtin SHA256 is not thread-safe.
     # See https://github.com/rust-lang/git2-rs/issues/1255.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [stable, beta, nightly, macos, windows]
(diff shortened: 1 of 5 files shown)


DanielEScherzer:
```
error: rustc 1.87.0 is not supported by the following packages:
  time@0.3.47 requires rustc 1.88.0
  time-core@0.1.8 requires rustc 1.88.0
```

but lets just not use the locked version

.github/workflows/main.yml
```diff
@@ -33,7 +37,15 @@ jobs:
...
-    - run: cargo test --locked
+    - name: Cargo test (locked)
+      if: ${{ matrix.build != 'MSRV' }}
+      run: cargo test --locked
+    - name: Remove lock file
+      if: ${{ matrix.build == 'MSRV' }}
+      run: rm Cargo.lock
+    - name: Cargo test (unlocked)
+      if: ${{ matrix.build == 'MSRV' }}
+      run: cargo test
```
weihanglo, line 48: Can we generate the lockfile for 1.87, so that is likely be compatible with other versions of toolchains?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1277/9a08ba3c2a59104a5a11e8cbab76f4640828b633..436249ff3e864122e917b1aab24ff795b1325e10)*

weihanglo: Oh you addressed this in a later commit. Never mind.

weihanglo (approved):

PR #1280, opened 2026-06-15
[Clippy] Add clippy to CI, allow current violations at a module level

DanielEScherzer:
Add clippy to CI so that future changes do not increase the number of violations. The existing violations are allowlisted on a module level (i.e. at the top of files with inner attributes, or above the start of nested inline test modules with outer attributes).

CI runs clippy on both Window and Linux because some lints may fire on one but not the other; see, e.g. #1238.

Existing violations will be addressed in subsequent pull requests.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -1,5 +1,8 @@
 //! Builder-pattern objects for configuration various git operations.
 
+#![allow(clippy::manual_non_exhaustive)]
+#![allow(clippy::missing_safety_doc)]
+
 use libc::{c_char, c_int, c_uint, c_void, size_t};
 use std::ffi::{CStr, CString};
 use std::mem;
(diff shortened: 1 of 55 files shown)


src/build.rs
```diff
@@ -767,6 +770,8 @@ impl TreeUpdateBuilder {
 }
 
 #[cfg(test)]
+#[allow(clippy::needless_borrow)]
```
weihanglo, line 773: Is it possible to add these allow to `[lints.clippy]` table?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1280/63b680ba068fa064087358ea5ae319f224822b26..134e953e426b10b580daafad815c9bdba0a969cc)*

DanielEScherzer: I wanted to make it clear where the existing violations were so that it would be clear what needs fixing, and so that we don't introduce more warnings in other files. If this is merged (or if you expect to merge this soon) I can start sending the PRs to fix the violations and remove the allowing

weihanglo (approved):
> Existing violations will be addressed in subsequent pull requests.

I am sold on this. Thanks! Hope we can remove most of them.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1280/63b680ba068fa064087358ea5ae319f224822b26..5f690e4bcef945f8bba50fd600cdc354c774a76c)*

PR #1285, opened 2026-06-27
feat: expose `DiffDelta::similarity`

WillLillis:
The `TODO` comment here appears to be stale. The `similarity` score works correctly for me in the `delta_similarity` test as in a project depending on git2 with this patch applied.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -524,10 +524,11 @@ impl<'a> DiffDelta<'a> {
         result
     }
 
-    // TODO: expose when diffs are more exposed
-    // pub fn similarity(&self) -> u16 {
-    //     unsafe { (*self.raw).similarity }
-    // }
+    /// Returns the similarity score (0-100) for a `Renamed` or `Copied`
+    /// delta, or 0 if rename detection has not run.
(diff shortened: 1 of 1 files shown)


DanielEScherzer (approved):
LGTM, other methods already dereference this pointer so we can assume it is valid, similarity is indeed stored as a u16 based on libgit2, https://github.com/libgit2/libgit2/blob/32b564e63f9639eaf5ee90fb7a95b3a650156cbd/include/git2/diff.h#L327

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1285/ec20d5075863f639abcae4bb79805de2a745517c..7cf2aadadb428277a54316fde4db88e1b5a2e2d4)*

ehuss (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1285/ec20d5075863f639abcae4bb79805de2a745517c..7cf2aadadb428277a54316fde4db88e1b5a2e2d4)*

PR #1287, opened 2026-07-08
Fix `Remote::get_refspec()` lifetime bounds and add regression tests

### LINKED ISSUE #1286: Unsound lifetime bound in Remote::get_refspec
`Remote::get_refspec` returns a `RefSpec` with lifetime `'repo`, not the same lifetime as `&self`. This allows `Refspec` to outlive `Remote` (self).

https://github.com/rust-lang/git2-rs/blob/a00922bcdf0f78419aa5dd0d3643e73ce01672a8/src/remote.rs#L301-L306

## Reproduction

```rust
#[forbid(unsafe_code)]

use git2::Repository;

fn main() {
    let repo = Repository::init("temp_dir").unwrap();
    repo.remote("origin", "https://aaa.com/bbb.git").unwrap();
    
    let refspec = {
        let remote = repo.find_remote("origin").unwrap();
        remote.get_refspec(0)
    };
    // remote goes out of scope
    // but refspec's lifetime is not bounded to remote    
    
    // removing temp_dir, not necessary to trigger UAF
    let _ = std::fs::remove_dir_all("temp_dir");
    
    // triggers UAF
    let _ = refspec.unwrap().str();
}
```

You can get ASan UAF report by running following command. git2 latest release version (0.21.0) and latest commit in main branch (a00922bcdf0f78419aa5dd0d3643e73ce01672a8) print the same ASan report.
```
CFLAGS=-fsanitize=address RUSTFLAGS="-Zsanitizer=address" cargo +nightly run --target x86_64-unknown-linux-gnu
```

I reported this issue to security team first, and they said it's ok to make this issue publicly.

DanielEScherzer:
Set up testing with the `trybuild` crate so that we can confirm that the problematic code no longer compiles. I tried to add the regression test in a separate commit, but the problematic code compiled earlier (so `compile_fail()` could not be used) but did not run successfully (exit code 139 due to the use after free, so `pass()` could not be used either).

Fixes #1286

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ openssl-probe = { version = "0.1", optional = true }
 clap = { version = "4.4.13", features = ["derive"] }
 time = { version = "0.3.45", features = ["formatting"] }
 tempfile = "3.1.0"
+trybuild = "1.0.117"
 url = "2.5.4"
 
 [features]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,6 +225,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
+
+[[package]]
 name = "errno"
 version = "0.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -322,6 +328,7 @@ dependencies = [
  "openssl-sys",
  "tempfile",
  "time",
+ "trybuild",
  "url",
 ]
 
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -31,6 +31,7 @@ openssl-probe = { version = "0.1", optional = true }
 clap = { version = "4.4.13", features = ["derive"] }
 time = { version = "0.3.45", features = ["formatting"] }
 tempfile = "3.1.0"
+trybuild = "1.0.117"
```
weihanglo, line 34: Why 1.0.117 btw? The latest seems to be 1.0.120?

teor2345 (approved):
That's a lot of extra packages, but I guess it's only for tests.

Otherwise the fix looks good!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1287/a00922bcdf0f78419aa5dd0d3643e73ce01672a8..09269c6b4c292253054719f7bf968c1a100ca7be)*

weihanglo (approved):
Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1287/a00922bcdf0f78419aa5dd0d3643e73ce01672a8..09269c6b4c292253054719f7bf968c1a100ca7be)*

Cargo.toml
DanielEScherzer: 1.0.118 (and later) were released after I sent this PR

PR #1290, opened 2026-07-14
`Tag::is_valid_name()`: handle errors in CString conversion

DanielEScherzer:
Instead of panicking, treat failure to create a `CString` as indicating that the name is not valid.


DanielEScherzer:
C.f. #1229 where we did the same thing for reference names

weihanglo (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1290/1af7203bebfe78f59a272e301e13ce99d571d4aa..ac83bd42fa9f6d5bba90ff0337528e150cb11211)*

PR #1291, opened 2026-07-15
[Clippy] Address most clippy issues in production code

DanielEScherzer:
Commits are split by file and lint for ease of review. In most cases, the lints are addressed by fixing the reported issue as suggested.

# API changes

Notably, the following public API changes were made:

## New `Default` implementations

The following types now implement the `Default` trait:

* `ApplyOptions`
* `CherrypickOptions`
* `DiffPatchidOptions`
* `RepositoryInitOptions`
* `RevertOptions`
* `WorktreeAddOptions`
* `WorktreePruneOptions`

## New `is_empty()` methods

The following types have a new `is_empty()` method to go along with the existing `len()` method:

* `MessageTrailersStrs`
* `MessageTrailersBytes`
* `OdbObject`
* `Rebase`

## Other

* the `CloneLocal` enum is marked as `#[non_exhaustive]`, and the `__Nonexhaustive` variant was removed
* the implementations of the `Into` trait for converting `TreeWalkResult` to `i32` and to `raw::git_treewalk_mode` were replaced with implementations of the `From` trait

# Lints expected

In a few places, addressing the clippy issues would have required breaking changes to the existing API. Accordingly, instead of addressing the issues, the module-level `#[allow]` attributes were replaced with method-level `#[expect]` attributes, ensuring that new functions do not have the same issues while maintaining backwards compatibility. The follwing methods are expected to trigger clippy lints:

## `clippy::should_implement_trait`

* `ConfigEntries::next()`
* `Cred::default()`
* `Error::from_str()`
* `ObjectType::from_str()`
* `Oid::from_str()`

## `clippy::too_many_arguments`

* `Email::from_diff()`
* `Repository::diff_blobs()`

# Remaining `#[allow]` attributes

This PR does *not* address code for tests (code in `src/test.rs`, code in `mod tests` inline modules, or code in the `tests` directory). It also does not address code for examples (the `examples`) directory. When a file-wide `#[allow]` turned out to also be needed for the inline test modules, it was added to the tests module.

The library-wide `clippy::unnecessary_cast` allowing specified in `src/lib.rs` is also not addressed. Given that some of the casts are needed on unix but not windows, and others are needed on windows but not unix, that lint requires further attention. A dedicated PR already exists at #1238 reduce the unnecessary casts.

### DIFF
--- a/src/apply.rs
+++ b/src/apply.rs
@@ -1,8 +1,6 @@
 //! git_apply support
 //! see original: <https://github.com/libgit2/libgit2/blob/master/include/git2/apply.h>
 
-#![allow(clippy::new_without_default)]
-
 use crate::{panic, raw, util::Binding, DiffDelta, DiffHunk};
 use libc::c_int;
 use std::{ffi::c_void, mem};
@@ -152,6 +150,13 @@ impl<'cb> ApplyOptions<'cb> {
     /// must not outlive the [`ApplyOptions`] instance.
     pub unsafe fn raw(&mut self) -> *const raw::git_apply_options {
         &self.raw as *const _
+    }
+}
+
+impl<'cb> Default for ApplyOptions<'cb> {
+    /// Creates a new set of empty options (zeroed).
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/src/apply.rs
+++ b/src/apply.rs
@@ -1,7 +1,6 @@
 //! git_apply support
 //! see original: <https://github.com/libgit2/libgit2/blob/master/include/git2/apply.h>
 
-#![allow(clippy::missing_safety_doc)]
 #![allow(clippy::new_without_default)]
 
 use crate::{panic, raw, util::Binding, DiffDelta, DiffHunk};
@@ -146,6 +145,11 @@ impl<'cb> ApplyOptions<'cb> {
     }
 
     /// Pointer to a raw git_stash_apply_options
+    ///
+    /// # Safety
+    ///
+    /// The provided pointer must not be used to manipulate the options, and
+    /// must not outlive the [`ApplyOptions`] instance.
     pub unsafe fn raw(&mut self) -> *const raw::git_apply_options {
         &self.raw as *const _
     }
--- a/src/buf.rs
+++ b/src/buf.rs
@@ -1,5 +1,3 @@
-#![allow(clippy::explicit_auto_deref)]
-
 use std::ops::{Deref, DerefMut};
 use std::ptr;
 use std::slice;
@@ -38,7 +36,7 @@ impl Buf {
 
     /// Attempt to view this buffer as a string slice.
     pub fn as_str(&self) -> Result<&str, Error> {
-        str::from_utf8(&**self).map_err(|e| e.into())
+        str::from_utf8(self).map_err(|e| e.into())
     }
 }
 
@@ -78,6 +76,7 @@ impl Drop for Buf {
 }
 
 #[test]
+#[allow(clippy::explicit_auto_deref)]
 fn empty_buf() {
     let mut buf = Buf::new();
     let x: &[u8] = &*buf;
--- a/src/build.rs
+++ b/src/build.rs
@@ -1,7 +1,5 @@
 //! Builder-pattern objects for configuration various git operations.
 
-#![allow(clippy::missing_safety_doc)]
-
 use libc::{c_char, c_int, c_uint, c_void, size_t};
 use std::ffi::{CStr, CString};
 use std::mem;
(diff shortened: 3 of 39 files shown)


src/apply.rs
```diff
@@ -155,6 +153,13 @@ impl<'cb> ApplyOptions<'cb> {
     }
 }
 
+impl<'cb> Default for ApplyOptions<'cb> {
```
weihanglo, line 156: I am actually not fond of this. This may not be useful if we don't really need it. And that was why rust-lang/cargo only turn on a subset of lint group <https://github.com/rust-lang/cargo/blob/371a35405026fc16dd9968589f13cd746b96e024/Cargo.toml#L139-L145>.

Some of them are too pedantic.

src/buf.rs
```diff
@@ -78,6 +76,7 @@ impl Drop for Buf {
 }
 
 #[test]
+#[allow(clippy::explicit_auto_deref)]
```
weihanglo, line 79: I forgot in the last PR. we should use `#[expect]` instead for all `#[allow]`.

src/build.rs
```diff
@@ -590,8 +588,14 @@ impl<'cb> CheckoutBuilder<'cb> {
 
     /// Configure a raw checkout options based on this configuration.
     ///
+    /// # Safety
+    ///
     /// This method is unsafe as there is no guarantee that this structure will
-    /// outlive the provided checkout options.
+    /// outlive the provided checkout options. The caller must ensure that
+    /// `opts` does not outlive the `CheckoutBuilder`, and also that for any
```
weihanglo: Intra doc link would be nice:

```
[`CheckoutBuilder`]
```

src/cred.rs
```diff
@@ -503,6 +502,7 @@ impl CredentialHelper {
 
 #[cfg(test)]
 #[cfg(feature = "cred")]
+#[allow(clippy::needless_borrows_for_generic_args)]
```
weihanglo, line 504: `#[expect]`

src/oid.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::should_implement_trait)]
```
weihanglo: `#[expect]`

src/odb.rs
```diff
@@ -530,9 +528,8 @@ impl<'repo> Drop for OdbPackwriter<'repo> {
     fn drop(&mut self) {
         unsafe {
             let writepack = &*self.raw;
-            match writepack.free {
-                Some(free) => free(self.raw),
-                None => (),
+            if let Some(free) = writepack.free {
```
weihanglo, line 531: Like, this one is pretty annoying. It generally doesn't really matter.

DanielEScherzer:
For the claim about from implementations for TreeWalkResult: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=14aafe81106484b31590080fdd2386ae

src/oid.rs
weihanglo: `#[expect]`

weihanglo (commented):
Thanks. This a huge work!

BTW, is it possible to convert all `#[allow]` to `#[expect]`?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1291/00f31853745e635d3bef45edcad9955efe60f761..e45e67c8aaf15b727cba8944f12860584b06138b)*

src/oid.rs
DanielEScherzer: this gets removed later in the same PR

DanielEScherzer:
> Thanks. This a huge work!
> 
> BTW, is it possible to convert all `#[allow]` to `#[expect]`?

I didn't use expect because as I found out both here and on #1238 some lints might fire for one OS but not another
Plus the new allows that are added here are either
* moved from the file to the tests, and I will deal with soon
* temporarily added to `src/oid.rs` and I deal with it already in this PR

Should I remove the `Default` implementations? Disable the lint entirely?

weihanglo:
> Should I remove the Default implementations? Disable the lint entirely?

I am fine keeping them anyway. Just share with you my preference :)

> moved from the file to the tests, and I will deal with soon

This might be better, if not too many churns.

src/build.rs
DanielEScherzer: done here and for the others

weihanglo (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1291/3993d04410a38f5c24eddba37dafdb0b472a4ac2..3aedab1998db7a3e717c1a33002d6edd6b73f036)*

PR #1295, opened 2026-07-16
`StashSaveOptions::pathspec()`: return a `Result`, handle invalid values

DanielEScherzer:
When the provided path spec cannot be converted to a `CString`, return an `Err` variant rather than panicking.

See also

* <https://github.com/rust-lang/git2-rs/issues/1297>

### DIFF
--- a/src/stash.rs
+++ b/src/stash.rs
@@ -1,6 +1,6 @@
 use crate::build::CheckoutBuilder;
 use crate::util::{self, Binding};
-use crate::{panic, raw, IntoCString, Oid, Signature, StashApplyProgress, StashFlags};
+use crate::{panic, raw, Error, IntoCString, Oid, Signature, StashApplyProgress, StashFlags};
 use libc::{c_char, c_int, c_void, size_t};
 use std::ffi::{c_uint, CStr, CString};
 use std::mem;
@@ -45,11 +45,11 @@ impl<'a> StashSaveOptions<'a> {
     }
 
     /// Add to the array of paths patterns to build the stash.
-    pub fn pathspec<T: IntoCString>(&mut self, pathspec: T) -> &mut Self {
-        let s = util::cstring_to_repo_path(pathspec).unwrap();
+    pub fn pathspec<T: IntoCString>(&mut self, pathspec: T) -> Result<&mut Self, Error> {
+        let s = util::cstring_to_repo_path(pathspec)?;
         self.pathspec_ptrs.push(s.as_ptr());
         self.pathspec.push(s);
-        self
+        Ok(self)
     }
 
     /// Acquire a pointer to the underlying raw options.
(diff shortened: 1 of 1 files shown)


src/stash.rs
```diff
@@ -45,11 +45,11 @@ impl<'a> StashSaveOptions<'a> {
     }
 
     /// Add to the array of paths patterns to build the stash.
-    pub fn pathspec<T: IntoCString>(&mut self, pathspec: T) -> &mut Self {
-        let s = util::cstring_to_repo_path(pathspec).unwrap();
+    pub fn pathspec<T: IntoCString>(&mut self, pathspec: T) -> Result<&mut Self, Error> {
```
weihanglo, line 48: I know we kinda work towards this though it is still a bit annoying to users. Most of the people won't hit this issue and they need to take care of `Result`. I am a bit unsure.

Should we just make it an `expect` and document it would panic?





*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1295/0a6ff05f568109374bae1f48e850fb50d4170714..41ab557421dedff9e888514ff995d1c60902a094)*

DanielEScherzer: We could, but I think it would be nicer to return a result - if the library is able to avoid a panic without being left in a problematic state, why not do so?

weihanglo: Yes, and panic should be use mostly as a program invariant. This is kinda like a bad input. However, we can also see this as an invariant of these functions.

I am not sure though. I think technically it is more correct, though sacrifice ergonomic. And it may also be confusing as API caller now don't know how they should handle the error or not unless looking at the function body and realize it only checks bad cstring.

DanielEScherzer: We could also switch it to requiring that the provided value be a CString rather than just something that can be converted to one?

weihanglo: That also loses ergonomic in another way 😞.

DanielEScherzer: i figured a result was more ergonomic than requiring a CString, because I think panicking is worse than either

weihanglo: Can you point to a realistic workflow where an invalid pathspec reaches this API at runtime and the application needs to recover from it? Like in particular, is this intended for arbitrary user-provided data, or is it primarily addressing the fact that `CString` conversion is technically fallible?

Note that most OS and filesystem disallowed `\0` in the file system

DanielEScherzer: So I just started working on writing an extension for PHP in Rust that integrates with the git2 rust library for the logic - and that is what I was thinking of when I wanted to reduce the use of panics

Basically, returning a result (from Rust) means that we can throw an exception (in PHP) without needing to do the validation manually

weihanglo: > Basically, returning a result (from Rust) means that we can throw an exception (in PHP) without needing to do the validation manually

I knew the mechanism. Ideally we want to recover from more errors, but in reality what case on application side would hit this? If PHP is going to throw an exception what the difference between that and Rust panic?

BTW, PHP seems to have <https://github.com/libgit2/php-git>.

DanielEScherzer: PHP exceptions can be caught by the userland code

https://github.com/libgit2/php-git is very much outdated (and I'm doing this to be able to present a workshop on writing extensions in Rust, not just for fun)

PR #1284, opened 2026-06-25
CheckoutBuilder: bind `baseline{,_index}`

jonathantanmy2:
Add support for the `baseline` and `baseline_index` fields in CheckoutBuilder (to `struct git_checkout_options`).

Regarding lifetime of arguments passed to CheckoutBuilder, unlike the other fields in which owned versions are created upon function invocation, CheckoutBuilder only takes references to the input tree (`baseline()`) and index (`baseline_index()`) respectively. The Rust borrow checker enforces that the references live at least as long as CheckoutBuilder does.

Regarding safety of conversion of these new fields from CheckoutBuilder to `struct git_checkout_options` (in `configure()`), it is as unsafe as it currently is: the resulting `struct git_checkout_options` is valid as long as the source CheckoutBuilder lives.

PR #1303, opened 2026-07-22
CheckoutBuilder: bind `baseline{,_index}`

jonathantanmy2:
Add support for the `baseline` and `baseline_index` fields in CheckoutBuilder (to `struct git_checkout_options`).

Regarding lifetime of arguments passed to CheckoutBuilder, unlike the other fields in which owned versions are created upon function invocation, CheckoutBuilder only takes references to the input tree (`baseline()`) and index (`baseline_index()`) respectively. The Rust borrow checker enforces that the references live at least as long as CheckoutBuilder does.

Regarding safety of conversion of these new fields from CheckoutBuilder to `struct git_checkout_options` (in `configure()`), it is as unsafe as it currently is: the resulting `struct git_checkout_options` is valid as long as the source CheckoutBuilder lives.

### DIFF
--- a/src/build.rs
+++ b/src/build.rs
@@ -7,7 +7,7 @@ use std::path::Path;
 use std::ptr;
 
 use crate::util::{self, Binding};
-use crate::{panic, raw, Error, FetchOptions, IntoCString, Oid, Repository, Tree};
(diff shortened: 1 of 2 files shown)


jonathantanmy2:
Sorry, duplicate to #1284.

PR #1271, opened 2026-06-01
Reduce the amount of code in `unsafe` blocks

DanielEScherzer:
For statements that do not need to be marked as unsafe, pull them out of the blocks

PR #1306, opened 2026-07-26
Reduce the amount of code in `unsafe` blocks (2)

DanielEScherzer:
For statements that do not need to be marked as unsafe, pull them out of the blocks


DanielEScherzer:
Commits saying "Further reduce" are follow-ups to initial reduction in #1271

PR #1307, opened 2026-07-27
Remove redundant casts

FranciscoTGouveia:
I believe these casts can be safely removed, as the source type is already the target type.

For reference, these changes were found by an optimizer developed as a research project.

CC @nunoplopes


DanielEScherzer:
See #1238
Some of the casts are only redundant on linux machines and are needed on windows

FranciscoTGouveia:
Yes, after testing on another machine, I've just observed the issue.
That said, I do not think it's worthwhile to keep this PR open, so I will close it.
Apologies.

Issue #1297, opened 2026-07-16
Stop panicking when CString conversion fails


DanielEScherzer:
I started this at #1295 and #1296 and then realized that there are a lot more places that should be adjusted, so filing to discuss first.

```
CheckoutBuilder::path(), ancestor_label(), our_label(), their_label()
TreeUpdateBuilder::remove(), upsert()
DescribeFormatOptions::dirty_suffix()
DescribeOptions::pattern()
DiffOptions::old_prefix(), new_prefix(), pathspec()
EmailCreateOptions::subject_prefix()
ObjectType::from_str()
MergeFileOptions::ancestor_label(), our_label(), their_label()
MergeFileInput::path()
ProxyOptions::url()
RebaseOptions::rewrite_notes_ref()
Refspec::dst_matches(), src_matches(), transform(), rtransform()
FetchOptions::custom_headers()
PushOptions::custom_headers(), remote_push_options()
RepositoryInitOptions::workdir_path(), description(), template_path(), initial_head(), origin_url()
StatusOptions::pathspec()
Transaction::lock_ref(), set_target(), set_symbolic_target(), set_reflog(), remove()
Tree::get_name_bytes()
```

DanielEScherzer:
Should I be adding regression tests in one commit and switching to stop panicking in a separate commit, or would those tests not really be useful? I added tests in 1295 and 1296 but if there are this many places to update I realized that the tests might not really be useful

DanielEScherzer:
Also crosslinking #1294, #1290, #1229

weihanglo:
> Should I be adding regression tests in one commit and switching to stop panicking in a separate commit, or would those tests not really be useful? I added tests in 1295 and 1296 but if there are this many places to update I realized that the tests might not really be useful

Slightly prefer having tests, though fine with either way.

weihanglo:
See <https://github.com/rust-lang/git2-rs/pull/1295#discussion_r3595712680>.

weihanglo:
### Arguments

From <https://github.com/rust-lang/git2-rs/pull/1295#discussion_r3640615882>

> So I just started working on writing an extension for PHP in Rust that integrates with the git2 rust library for the logic - and that is what I was thinking of when I wanted to reduce the use of panics
> 
> Basically, returning a result (from Rust) means that we can throw an exception (in PHP) without needing to do the validation manually

From <https://github.com/rust-lang/git2-rs/pull/1295#discussion_r3640771112>

[...]

PR #1308, opened 2026-07-30
[CString panic reduction] Reduce panics without changing return types

DanielEScherzer:
For functions where the failure can be effectively community via the existing return type, instead of panicking when a CString cannot be created, return `false` or `None` or an error variant.

Part of #1297

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -955,7 +955,8 @@ impl ObjectType {
     /// Convert a string object type representation to its object type.
     #[expect(clippy::should_implement_trait)]
     pub fn from_str(s: &str) -> Option<ObjectType> {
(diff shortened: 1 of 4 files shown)


teor2345 (approved):
These all seem sensible to me, it's a routine and somewhat mechanical change.

(I don't have merge permissions here, but sometimes reviewers that do appreciate knowing others have checked the PR.)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/git2-rs/1308/da6c126f7733329d82748e710266af1c61591bdd..75bf90b2dd9ca86e02a2ec6da184fdf7ca3f5f45)*

